PR #71, opened 2022-05-27
[nt] Fix rerender table bug

nathaniel-mage:
# Summary

- Fix rerender bug causing routing to break
# Tests

- Went to localhost:3000/datatsets/{id} and viewed console for any rerenders/warnings/errors.
cc: @johnson-mage


nathaniel-mage:
going to close this as it's solved in another PR for the metrics table.

PR #73, opened 2022-05-27
[nt]  metrics table sort mapping

nathaniel-mage:
# Summary

- Sort metrics shown based on what's in Figma design
- Change naming to be more professional by adding a mapping.
- Removed count from metrics table.

# Tests

Localhost viewing the table on the page.
image

cc:

@johnson-mage @dy46


mage_ai/frontend/pages/datasets/[slug].tsx
```diff
@@ -36,14 +36,23 @@ function Data() {
...
+  const percentageKeys = ["completeness", "validity"];
+
+  // Map text and display priotities to backend keys.
+  const humanReadableMapping = {
+    "avg_invalid_value_count":[3, "Invalid values"],
+    "avg_null_value_count":[2, "Missing values"],
+    "completeness":[1, "Completeness"],
+    "duplicate_row_count":[4, "Duplicate values"],
+    "validity":[0, "Validity"],   
+  } || {};
```
Comment (dy46) [inline mage_ai/frontend/pages/datasets/[slug].tsx:55]: don't add `|| {}` when it's never going to null

Comment (dy46) [reply mage_ai/frontend/pages/datasets/[slug].tsx:55]: also i really don't like this. it's super confusing what you're trying to do here. it's just sorted alphabetically, can we just do a sort somewhere?

mage_ai/frontend/pages/datasets/[slug].tsx
```diff
@@ -66,22 +75,29 @@ function Data() {
...
-      if (metricsKeys.includes(val)) {
-        metricRows.push(values);
+    const metricRows = Array(metricsKeys.length).fill(0);
+    stats.forEach(function (key) {
+      if (metricsKeys.includes(key)) {
+        let value;
+        const order = humanReadableMapping[key];
+        if (percentageKeys.includes(key)){
+          value = statistics[key] * 100;
+          value = value + '%';
```
Comment (dy46) [inline mage_ai/frontend/pages/datasets/[slug].tsx:85]: can we use the `${...}%` syntax here?

Comment (nathaniel-mage) [reply mage_ai/frontend/pages/datasets/[slug].tsx:85]: for the concat string?

mage_ai/frontend/pages/datasets/[slug].tsx
Comment (nathaniel-mage) [reply mage_ai/frontend/pages/datasets/[slug].tsx:55]: It's actually not supposed to be sorted alphabetically. I had to change to change it so the display is sorted by what's on Figma. What's a better way to do this?

mage_ai/frontend/pages/datasets/[slug].tsx
Comment (dy46) [reply mage_ai/frontend/pages/datasets/[slug].tsx:85]: yes, sorry i updated my comment

mage_ai/frontend/pages/datasets/[slug].tsx
Comment (dy46) [reply mage_ai/frontend/pages/datasets/[slug].tsx:55]: oh oops. can you just create another array with the values in order and then create the metrics based on that array? it should keep the order if you use an array

johnson-mage (approved):

PR #75, opened 2022-05-27
[sp] column view backend integration

shrey-mage:
# Summary
- see above
- update styling to be closer to Figma

# Tests
image

This PR will not work until Nathaniel's bugfix (#73) is merged as well, since that affects the dataset view used to get to this page.

In addition, the bug where NextJS's router doesn't allow direct paths affects this page too, so accessing /datasets/[id]/features directly does not work and causes an error. If there's a fix please let me know

cc: @johnson-mage @nathaniel-mage

### DIFF
--- a/mage_ai/frontend/pages/datasets/[slug]/features/index.tsx
+++ b/mage_ai/frontend/pages/datasets/[slug]/features/index.tsx
@@ -1,5 +1,6 @@
-import Router from 'next/router'
+import Router, { useRouter } from 'next/router'
 import type { NextPage } from 'next'
+import { useEffect, useMemo, useState } from 'react'
 
 import Button from '@oracle/elements/Button'
 import FlexContainer from '@oracle/components/FlexContainer'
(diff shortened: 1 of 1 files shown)


nathaniel-mage:
I saw that you mentioned #73, how is the metrics table blocking the PR? Not sure if I'm missing something.

shrey-mage:
> I saw that you mentioned #73, how is the metrics table blocking the PR? Not sure if I'm missing something.

It's not blocking, we just can't test the column view independently without your bugfix to the infinite rerender issue on the dataset view, because you can't directly access the /features page because of the NextJS routing bug. I tested it locally by making the change you sent, but I didn't want to recommit that, so I removed that before pushing this PR.

nathaniel-mage (commented):
Is it possible to remove updating the out files from this commit? I have another one pending that removes all out files and I think this could cause issues.

shrey-mage:
Oh yeah I didn't mean to add out files to this commit, don't know why it's in there. I'll try to remove that

nathaniel-mage (approved):
Looks good.

PR #86, opened 2022-05-28
[sp] add suggestions view + integration

shrey-mage:
# Summary
- see above

# Tests
Styling may not be exact, but if needed for MVP, it can be updated to match the Figma more closely.

cc: @johnson-mage @nathaniel-mage @wangxiaoyou1993


mage_ai/frontend/pages/datasets/[slug].tsx
```diff
@@ -190,14 +244,119 @@ function Data() {
...
+          padding="0px"
+          transparent
+        >
+          <Close muted />
+        </Button>
+      </FlexContainer>
+    </RowCard>
+  );
+
+  const actionsEl = (
```
Comment (johnson-mage) [inline mage_ai/frontend/pages/datasets/[slug].tsx:303]: Prob wudve been better to create a new component in the `components` directory called `SuggestionRow` or something similar instead of creating the component directly in the pages file, but for sake of time, we'll leave it here and reorganize the components later.

mage_ai/frontend/pages/datasets/[slug].tsx
```diff
@@ -190,14 +244,119 @@ function Data() {
...
+          name={title}
+          numFeatures={numFeatures}
+          onClose={() => removeAction(idx)}
+          showIdx
+        />
+      );
+    })
+  );
+
+  const suggestionsEl = (
```
Comment (johnson-mage) [inline mage_ai/frontend/pages/datasets/[slug].tsx:328]: Same thing goes for this. Seems better to create a `SuggestionsList` component separately and import it into this file.

mage_ai/frontend/pages/datasets/[slug].tsx
```diff
@@ -190,14 +244,119 @@ function Data() {
...
+                key={`${idx}-${suggestion.title}`}
+                link={() => addAction(idx)}
+                name={suggestion.title}
+                numFeatures={numFeatures}
+                onClose={() => removeSuggestion(idx)}
+              />
+            )
+          })
+          :
+          <>{/* TODO: what do we render when no suggestions exist? */}</>
```
Comment (johnson-mage) [inline mage_ai/frontend/pages/datasets/[slug].tsx:354]: Would be similar to the empty datasets table. Would just say something like "No suggestions available."
image

mage_ai/frontend/pages/datasets/[slug].tsx
```diff
@@ -37,10 +43,24 @@ function Data() {
...
+  ), [
+    datasetResponse?.suggestions,
+  ]);
+
   const [columnHeaderSample, setColumnHeaderSample] = useState([{}]);
   const [rowGroupDataSample, setRowGroupDataSample] = useState({});
   const [metricSample, setMetricSample] = useState({});
   const [statSample, setStatSample] = useState({});
+
+  const [suggestions, setSuggestions] = useState([]);
```
Comment (johnson-mage) [inline mage_ai/frontend/pages/datasets/[slug].tsx:57]: We'll need to add in a `SuggestionType` interface.

mage_ai/frontend/pages/datasets/[slug].tsx
```diff
@@ -132,6 +152,40 @@ function Data() {
...
+
+  const addAction = i => {
+    setActions(actions.concat({ i, suggestions: suggestions[i] }));
+  };
+
+  const removeAction = i => {
+    setActions(actions.filter((x, idx) => i !== idx));
+  }
+
+  const removeSuggestion = i => {
```
Comment (johnson-mage) [inline mage_ai/frontend/pages/datasets/[slug].tsx:175]: I believe we keep track of skipped/removed suggestions with the suggestion's `status`, which needs to be persisted, rather than just using the React state.

mage_ai/frontend/pages/datasets/[slug].tsx
```diff
@@ -190,14 +244,119 @@ function Data() {
...
-      >
+      <Button onClick={viewColumns}>
         <Text bold> Column view </Text>
       </Button>
     </FlexContainer>
   );
 
+  type ActionProps = {
+    idx: number;
+    link?: () => void;
```
Comment (johnson-mage) [inline mage_ai/frontend/pages/datasets/[slug].tsx:255]: The property name `link` seems a bit misleading. Initially thought it was an object versus an action.

mage_ai/frontend/pages/datasets/[slug].tsx
```diff
@@ -37,10 +43,24 @@ function Data() {
...
+
   const [columnHeaderSample, setColumnHeaderSample] = useState([{}]);
   const [rowGroupDataSample, setRowGroupDataSample] = useState({});
   const [metricSample, setMetricSample] = useState({});
   const [statSample, setStatSample] = useState({});
+
+  const [suggestions, setSuggestions] = useState([]);
+
+  // structured as [{ idx, action_data }]
+  const [actions, setActions] = useState([]);
```
Comment (johnson-mage) [inline mage_ai/frontend/pages/datasets/[slug].tsx:60]: A type or interface would've been helpful here. Hard to follow the type of data being stored here.

johnson-mage (approved):
Approving so we can get this merged into master in case conflicts arise. But we'll need to refactor some stuff and update how we handle skipped or applied suggestions.

mage_ai/frontend/pages/datasets/[slug].tsx
Comment (shrey-mage) [reply mage_ai/frontend/pages/datasets/[slug].tsx:60]: Agreed. For the sake of time I didn't end up making interfaces, but we should definitely do that for all types of payloads (eg. for feature sets payload, a single feature set payload, and individual actions/suggestions payload).

mage_ai/frontend/pages/datasets/[slug].tsx
Comment (shrey-mage) [reply mage_ai/frontend/pages/datasets/[slug].tsx:303]: Completely agree. Like we talked about earlier we'll have to make a pass through the frontend and componentize elements into separate files

mage_ai/frontend/pages/datasets/[slug].tsx
Comment (shrey-mage) [reply mage_ai/frontend/pages/datasets/[slug].tsx:175]: Yes, this was my biggest concern, if the functionality of suggestions/actions lined up with how the data cleaning tool should actually work. I think we should check in with Xiaoyou/David ASAP if everything works the way the backend intended

mage_ai/frontend/pages/datasets/[slug].tsx
Comment (shrey-mage) [reply mage_ai/frontend/pages/datasets/[slug].tsx:255]: I'll make notes of all of these comments as P2 updates and make double-checking skipped/applied suggestions as P1.

PR #108, opened 2022-06-01
[sp] update suggestions list

shrey-mage:
# Summary
- fix suggestions so they persist state throughout frontend
- add suggestions/actions list to column view page
- encapsulate logic in `SuggestionRow` and `SuggestionsList` components

# Tests
Modifications to the suggestions list on the column view page seem to get overwritten by the old state on the dataset detail page. Since all suggestions/actions logic is encapsulated in `SuggestionsList`, I'm not sure why this would happen. If someone could double check, that would be great.

Marking this as a draft PR since Tommy's addition of the action form diverged from my encapsulation changes, so I wanted to check before rebasing to see if another method of merging/cherry-picking would be easier, and also to get a check for code structure before going for a full rebase.

cc: @tommydangerous @johnson-mage


johnson-mage:
You can organize `SuggestionRow` and `SuggestionList` under `mage_ai/frontend/components/suggestions` since using an `actions` directory might get confused with the transformer actions, which seems like a separate thing.

mage_ai/frontend/components/actions/SuggestionsList.tsx
```diff
@@ -0,0 +1,137 @@
+import api from '@api';
```
johnson-mage, line 1: alphabetize imports

johnson-mage: caps, lowercase, bracket caps, bracket lowercase

mage_ai/frontend/components/actions/SuggestionsList.tsx
```diff
@@ -0,0 +1,137 @@
...
+import Accordion, { AccordionPanel } from '@oracle/components/Accordion';
+import Spacing from '@oracle/elements/Spacing';
+import { useMemo, useState, useEffect } from 'react';
+import SuggestionRow from './SuggestionRow';
+
+export type SuggestionsListProps = {
+  featureSet: any;
+  featureSetId: any;
```
johnson-mage, line 9: We will need to add a `featureSet` interface later, but `featureSetId` type should be defined, like as `number`.

mage_ai/frontend/components/actions/SuggestionsList.tsx
```diff
@@ -0,0 +1,137 @@
...
+          name={title}
+          numFeatures={numFeatures}
+          onClose={() => removeAction(idx)}
+          showIdx
+        />
+      );
+    })
+  );
+
+  const suggestionsEl = (
```
johnson-mage, line 96: You can just put this in the return statement instead of separating it out like this, espec since we've already abstracted this component from the page. We sometimes initialize components outside of the return statement when it gets too verbose, but the return statement in this component is only 3 lines.

johnson-mage:
done reviewing @shrey-mage

PR #125, opened 2022-06-02
[dy] Add cd job

dy46:
# Summary

Add CD job to build the frontend and push it to master on an hourly basis. I decided to do it this way because I thought it might be too much if it pushed to master every time a change was made in the repo. I can also change it to a daily schedule when we don't need to update the frontend as much.

# Tests

tested the logic with the github circle ci job, but not sure how to test the hourly trigger logic.

cc: @wangxiaoyou1993


wangxiaoyou1993:
Another way is to add a manual approval step to export frontend build.

dy46:
@wangxiaoyou1993 would that be on every frontend build? would it be ok if we skipped that step sometimes?

wangxiaoyou1993:
Yes, you can skip that step by just not clicking the approval button.

dy46:
Ok that's a good idea, I'll do that

dy46:
@wangxiaoyou1993 ok I updated it

.circleci/config.yml
```diff
@@ -23,3 +63,12 @@ workflows:
   build_and_test:
     jobs:
       - test
+      - build_frontend
+      - export_frontend:
```
wangxiaoyou1993, line 67: Could you follow this example to add the approval step?
https://github.com/mage-ai/backend/blob/master/.circleci/config.yml#L499-L515
https://github.com/mage-ai/backend/blob/master/.circleci/config.yml#L523-L531

dy46:
@wangxiaoyou1993 ok i updated it, but i didn't include the job to post to slack since it feels a bit unnecessary

PR #155, opened 2022-06-03
[sk] Fixed infinite loop bug

skunichetty:
# Summary
Fixed bug where pipeline loops forever if auto transform is disabled. Switched back to using a deque to allow for fast front element access and removal so that pipeline will properly empty.

cc:
@wangxiaoyou1993


skunichetty:
Please take a look

PR #158, opened 2022-06-03
[sk] Fixed Flask resource loading bugs

skunichetty:
# Summary
Fixed a couple of bugs in Flask endpoints:
- When a non-integer id is used for a pipeline, originally it would try to be converted to an integer and throw a hard to find error. Now the error is raised directly, letting the user know that someone an invalid id had been created.
     image
    Usually the error is raised when the user manually adds in their own feature set or pipeline incorrectly to the files folder.
- When a dataset with some id xx does not exist, accessing the Flask GET endpoint (feature_sets/xx or pipelines/xx) with creates an empty pipeline and feature set in files (a ghost storage model). Now when this is attempted an internal server error gets raised. The PUT endpoint is not modified.
    image
   This is the error now generated when accessing http://localhost:3000/datasets/329423 when the feature set and pipeline with id 329423 don't exist.

# Tests
Errors were tested locally through accessing the endpoints to check for intended behavior.

cc:
@wangxiaoyou1993, @dy46

### DIFF
--- a/mage_ai/server/app.py
+++ b/mage_ai/server/app.py
@@ -8,12 +8,10 @@ from mage_ai.server.data.models import FeatureSet, Pipeline
 from numpyencoder import NumpyEncoder
 import logging
 import json
-import os
 import simplejson
 import sys
 import threading
 
-RESOURCE_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), 'data/files'))
 
 log = logging.getLogger('werkzeug')
 log.setLevel(logging.ERROR)
@@ -92,13 +90,13 @@ def feature_sets():
     ]
     """
     feature_sets = list(map(lambda fs: fs.to_dict(detailed=False), FeatureSet.objects()))
-    valid_feture_sets = [
+    valid_feature_sets = [
         f
         for f in feature_sets
         if set(['column_types', 'statistics']).issubset(f['metadata'].keys())
     ]
     response = app.response_class(
-        response=simplejson.dumps(valid_feture_sets, ignore_nan=True),
+        response=simplejson.dumps(valid_feature_sets, ignore_nan=True),
         status=200,
         mimetype='application/json',
     )
--- a/mage_ai/server/app.py
+++ b/mage_ai/server/app.py
@@ -115,13 +115,8 @@ def feature_set(id):
         }
     ]
     """
-    path = RESOURCE_PATH + '/FeatureSet'
-    try:
-        opts = os.listdir(path)
-        if id not in opts:
-            raise RuntimeError(f'Unknown feature set id: {id}')
-    except OSError:
-        pass
+    if not FeatureSet.is_valid_id(id):
+        raise RuntimeError(f'Unknown feature set id: {id}')
     feature_set = FeatureSet(id=id)
     query_column = request.args.get('column')
     response = app.response_class(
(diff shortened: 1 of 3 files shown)


mage_ai/server/app.py
```diff
@@ -110,12 +115,19 @@ def feature_set(id):
         }
     ]
     """
+    path = RESOURCE_PATH + '/FeatureSet'
```
wangxiaoyou1993, line 118: Let's add these logic into this class: https://github.com/mage-ai/mage-ai/blob/master/mage_ai/server/data/base.py

mage_ai/server/data/base.py
```diff
@@ -65,11 +66,22 @@ def path_name(cls):
...
         return arr
+
+    @staticmethod
+    def gen_integer_dir_list(pathname):
+        dirs = []
+        for dirname in os.listdir(pathname):
+            try:
+                dirs.append(int(dirname))
+            except ValueError:
+                raise RuntimeError(
```
wangxiaoyou1993, line 83: Just ignore the non-numerical directory. No need to raise exception.

mage_ai/server/app.py
skunichetty: Added to base class

mage_ai/server/data/base.py
skunichetty: Dataset view now ignores nonnumerical directories

mage_ai/server/app.py
```diff
@@ -90,12 +92,15 @@ def feature_sets():
     ]
     """
     feature_sets = list(map(lambda fs: fs.to_dict(detailed=False), FeatureSet.objects()))
-    valid_feture_sets = [f for f in feature_sets
-                         if set(['column_types', 'statistics']).issubset(f['metadata'].keys())]
+    valid_feture_sets = [
```
dy46, line 95: nit: can we fix the typo in this variable

mage_ai/server/app.py
```diff
@@ -8,17 +8,19 @@
 from numpyencoder import NumpyEncoder
 import logging
 import json
+import os
 import simplejson
 import sys
 import threading
 
+RESOURCE_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), 'data/files'))
```
wangxiaoyou1993, line 16: where is it used?

mage_ai/server/data/base.py
```diff
@@ -1,3 +1,4 @@
+from multiprocessing.sharedctypes import Value
```
wangxiaoyou1993, line 1: where is it used?

mage_ai/server/data/base.py
```diff
@@ -46,30 +47,43 @@ def read_parquet_file(self, file_name):
...
         pass
-    
+
     @classmethod
     def folder_name(cls):
         return cls.__name__
 
+    @classmethod
+    def is_valid_id(cls, id):
+        opts = os.listdir(cls.path_name())
```
wangxiaoyou1993, line 64: not sure whether there will be type difference between id and the dir names. let's convert them to int for comparison to avoid the issue.

skunichetty: Converted both id and directory names to integers before comparison

mage_ai/server/data/base.py
skunichetty: Bad import, removed.

mage_ai/server/app.py
skunichetty: Bad import, removed.

mage_ai/server/app.py
skunichetty: Fixed

wangxiaoyou1993 (approved):

mage_ai/server/data/base.py
```diff
@@ -46,30 +46,43 @@ def read_parquet_file(self, file_name):
...
     def path_name(cls):
         return os.path.join(DATA_PATH, cls.folder_name())
 
     @classmethod
     def objects(cls):
         arr = []
-        dirs = sorted([int(name) for name in os.listdir(cls.path_name()) if name.isnumeric()],
-                      reverse=True)
-        for id in dirs:
+        for id in Model.gen_integer_dir_list(cls.path_name()):
```
wangxiaoyou1993, line 73: @skunichetty could you add the sorted function back?

skunichetty: Oops, not sure how that got deleted. I'll add that back in right now.

PR #161, opened 2022-06-03
[sk] Adding cleaning action dependencies

skunichetty:
# Summary
Added logic to enforce that some conditions are met before cleaning actions are taken. Currently, the main dependency is that any filter action requires column types to be cleaned beforehand (no whitespace otherwise row filtering breaks). If dependencies are not met then an error is raised.

There are some reformatting edits as well.

# Tests
Prior unit tests were re-evaluated alongside newly added unit tests to confirm that an error is raised when dependencies are not satisfied but otherwise functionality remains the same.

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_ai/data_cleaner/transformer_actions/base.py
+++ b/mage_ai/data_cleaner/transformer_actions/base.py
@@ -1,8 +1,8 @@
 from mage_ai.data_cleaner.transformer_actions import column, row
 from mage_ai.data_cleaner.transformer_actions.constants import ActionType, Axis, VariableType
 from mage_ai.data_cleaner.transformer_actions.dependency_resolution import (
-    resolve_filter_action,
     default_resolution,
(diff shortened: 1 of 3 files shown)


mage_ai/data_cleaner/transformer_actions/base.py
```diff
@@ -1,15 +1,20 @@
 from mage_ai.data_cleaner.transformer_actions import column, row
 from mage_ai.data_cleaner.transformer_actions.constants import ActionType, Axis, VariableType
+from mage_ai.data_cleaner.transformer_actions.dependency_resolution import (
+    resolve_filter_action,
```
wangxiaoyou1993, line 4: nit: alphabetical order

wangxiaoyou1993 (approved):

PR #172, opened 2022-06-04
[nt]  dataset metrics formating

nathaniel-mage:
# Summary

### Updates Dataset details > Reports tab
- Changes the wording of the message: values -> rows, features -> columns
- Updates the formatting to not display e+...
- Refactors code to use the utils functions.
- Update payload keys to use totals instead of average.

# Tests

### Reports Tables (minus feature profile)
image

cc:

@johnson-mage @dy46

### DIFF
--- a/mage_ai/frontend/components/datasets/overview/utils.ts
+++ b/mage_ai/frontend/components/datasets/overview/utils.ts
@@ -1,3 +1,4 @@
+import { getPercentage } from '@utils/number';
 import {
   CATEGORICAL_TYPES,
   DATE_TYPES,
@@ -23,7 +24,7 @@ export function createMetricsSample(statistics) {
         value = value.toPrecision(2);
         value *= 100;
         bar = [true, value];
-        value = `${value}%`;
+        value = getPercentage(value);
       }
       metricRows[index] = {
         columnValues: [order, value, bar],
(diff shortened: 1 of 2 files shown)


nathaniel-mage:
No need to review yet, still need to add "Display percentage of total"

wangxiaoyou1993:
"Missing rows" => "Missing cells". "Invalid rows" => "Invalid cells".
"total_invalid_value_count" is already added to backend.

nathaniel-mage:
> 

Got it will change to cells instead of rows. For duplicate values/rows do we alsowant to change that to cells?

mage_ai/frontend/components/datasets/overview/utils.ts
```diff
@@ -15,14 +17,14 @@ export function createMetricsSample(statistics) {
...
+      let value = statistics[key];
       const order = HUMAN_READABLE_MAPPING[key];
       const index = METRICS_SORTED_MAPPING[key];
       if (PERCENTAGE_KEYS.includes(key)) {
-        value *= 100;
-        bar = [true, value];
-        value = `${value}%`;
+        value = value.toPrecision(2);
+        bar = [true, (value * 100)];
+        value = getPercentage(value);
```
johnson-mage, line 27: should this be calling the `getPercentage` method on the original `statistics[key]` value instead of the value modified with `toPrecision`?

johnson-mage (approved):
approved. just had one comment

mage_ai/frontend/components/datasets/overview/utils.ts
nathaniel-mage: For the original values that are not percentage keys we don't want to turn them into percentages. That's why we only call get percentage here. So we avoid values like 0.0 when it's 0 or values with e+3 when it's in the thousands.

PR #177, opened 2022-06-04
[nt] Add progress bar to cell and simple datatable

nathaniel-mage:
# Summary

### Add progress bar display to certain charts
- If a table is given a list inside it, it'll look for a pair. The first being a boolean to determine whether to include a progress bar, followed by the value.
- Updates cell file to display progress bars using the "showProgress" prop.
- Change default flex values to get the table to closer match figma, though the cell's padding makes it difficult for the chart to be exactly like Figma where the progress bar and percentage are both closer to each other.

# Tests

Ran on localhost 

### Max progress
image

### Good progress
image

### Bad progress
image

# Known issues/Improvements
* The progress bar is off center... I think this is due to how it renders inside a cell. 
* I also added the danger property for the red version of the progress bar when it's below average (C = 75%). We can extend this with a constant in the future when we do good/bad/average.

cc:

@johnson-mage @dy46 @tommydangerous

### DIFF
--- a/mage_ai/frontend/components/datasets/overview/index.tsx
+++ b/mage_ai/frontend/components/datasets/overview/index.tsx
@@ -236,7 +236,7 @@ function DatasetOverview({
             <Flex flex={1}>
               {metricSample && (
                 <SimpleDataTable
-                  columnFlexNumbers={[1, 1]}
+                  columnFlexNumbers={[2, 1, 2 ]}
                   columnHeaders={[{ label: 'Quality Metrics' }]}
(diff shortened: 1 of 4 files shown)


tommydangerous:
"The progress bar is off center... I think this is due to how it renders inside a cell."

Please fix. Try flex container align items center

tommydangerous (approved):
Approved only after you vertically align the progress bar

nathaniel-mage:
Thanks Tommy! Looks centered now
image

PR #179, opened 2022-06-04
[nt]  Basetable polish p2

nathaniel-mage:
# Summary

### BaseTable polishing
- Add rounded corners
- Removed fancy layout hooks
- Improved styling on the inside of the table
- Shrunk the size of table down for more apparent vertical scrolling

# Tests

Localhost 

### Ugly dataset with long values
image

### Dataset with long column names
image

### Small dataset
image

**Doesn't look that good since the borders don't align**

### Additions & improvements
- Change px to rem and em, 
- Find another way to style that works with the react-table without passing in a style prop.
- Display the datatype on top of the column
- Add a sticky column header.
- There is a bug where if the table is at max scroll it will scroll the page by the direction. Can be annoying at times.

cc:

@johnson-mage @dy46 @tommydangerous

### DIFF
--- a/mage_ai/frontend/oracle/components/Table/BaseTable.tsx
+++ b/mage_ai/frontend/oracle/components/Table/BaseTable.tsx
@@ -150,7 +150,7 @@ import { cutTextSize } from './helpers';
               return (
                 // eslint-disable-next-line react/jsx-key
                 <tr {...row.getRowProps()}>
-                  {row.cells.map((cell) => (
+                  {row.cells.map(cell => (
                     // eslint-disable-next-line react/jsx-key
                     <td
                       {...cell.getCellProps()}
--- a/mage_ai/frontend/oracle/components/Table/BaseTable.tsx
+++ b/mage_ai/frontend/oracle/components/Table/BaseTable.tsx
@@ -119,17 +119,14 @@ import { cutTextSize } from './helpers';
           { headerGroups.map(headerGroup => (
             // eslint-disable-next-line react/jsx-key
             <tr {...headerGroup.getHeaderGroupProps()}>
-              {headerGroup.headers.map((column, i, cols) => (
+              {headerGroup.headers.map(column => (
                 // eslint-disable-next-line react/jsx-key
                 <th
                   {...column.getHeaderProps()}
                   style={{
                     background: '#F9FAFC',
                     border: 'solid 1px #D8DCE3',
-                    borderLeft: (i === 0) && 'none',
-                    borderRight: (i === cols.length - 1) && 'none',
-                    borderTop: 'none',
-                    padding: '14px',
+                    padding: '4px',
                   }}
                 >
                   <TextStyle>
@@ -145,7 +142,7 @@ import { cutTextSize } from './helpers';
         </thead>
         {/* Rows: relative, overflow, black text, borders on everything but bottom except for last, skip bg */}
         <tbody {...getTableBodyProps()}>
-          {rows.map((row, i) => {
+          {rows.map((row, i, arr) => {
               prepareRow(row);
               return (
                 // eslint-disable-next-line react/jsx-key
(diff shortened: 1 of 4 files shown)


mage_ai/frontend/oracle/components/Table/BaseTable.tsx
```diff
@@ -160,27 +150,21 @@ import { cutTextSize, getColumnWidth } from './helpers';
               return (
                 // eslint-disable-next-line react/jsx-key
                 <tr {...row.getRowProps()}>
-                  {row.cells.map(cell => (
+                  {row.cells.map((cell) => (
```
tommydangerous, line 153: Read the style guide. If you are returning something immediately in a loop without an explicit return function, you don't need to wrap a single parameter in parentheses

mage_ai/frontend/oracle/components/Table/SimpleDataTable.tsx
```diff
@@ -15,6 +15,8 @@ import {
 } from './index.style';
 import { Check } from '@oracle/icons';
 import Spacing from '@oracle/elements/Spacing';
+import ProgressBar from '../ProgressBar';
```
tommydangerous, line 18: Why are you importing this

mage_ai/frontend/oracle/components/Table/Table.style.tsx
```diff
@@ -2,13 +2,19 @@ import styled from 'styled-components';
 
 import light from '@oracle/styles/themes/light';
 import { UNIT, PADDING_UNITS } from '@oracle/styles/units/spacing';
+import { BORDER_RADIUS_LARGE } from '@oracle/styles/units/borders';
 
 export const TableStyle = styled.div<any>`
   position: relative;
+  text-align: left;
+  position: relative;
+  border: solid 1px #D8DCE3;
```
tommydangerous, line 11: Don't hardcode this, get colors from the theme

mage_ai/frontend/oracle/components/Table/Table.style.tsx
```diff
@@ -2,13 +2,19 @@ import styled from 'styled-components';
 
 import light from '@oracle/styles/themes/light';
 import { UNIT, PADDING_UNITS } from '@oracle/styles/units/spacing';
+import { BORDER_RADIUS_LARGE } from '@oracle/styles/units/borders';
 
 export const TableStyle = styled.div<any>`
   position: relative;
+  text-align: left;
```
tommydangerous, line 9: Alphabetize the styles

mage_ai/frontend/oracle/components/Table/BaseTable.tsx
nathaniel-mage: Got it, will fix that. I forgot about it after I was using cell, i but removed the i.

mage_ai/frontend/oracle/components/Table/SimpleDataTable.tsx
nathaniel-mage: Strange, that shouldn't be there

johnson-mage:
@nathaniel-mage The designs have borders between the columns for the data table, right?. Can you add an optional `showBorders` prop or something like that to display the column borders in the table? Makes it easier to view the data with those border delineations.

johnson-mage:
@nathaniel-mage For the "small dataset" where there's not enough columns to fit the container width, please figure out how to make the container only fit the columns and the table border go around

johnson-mage (approved):
approved after addressing comments. we'll continue to improve the column width sizing in a separate PR.

nathaniel-mage:
I was able to fix the width sizing but at the cost of not surrounding the container with borders. Due to this the cells don't look rounded (i tried with borderradius styling and it didn't work). In the future I'd like to refactor the table cells to not use the html table elements and instead use the cells elements. Also because it's no longer inside a container, the bottom cells don't have borders until the user scrolls down. This can be useful as an indicator that there's more left in the dataset.
image

Also, there's a weird bug I'm looking into where the last column of some datasets aren't displaying. No idea why. Might need some help with this one before we launch in the morning @johnson-mage

PR #187, opened 2022-06-06
[sk] Support for custom cleaning actions

skunichetty:
# Summary
Backend transformer action library now supports custom cleaning actions. The cleaning action accepts primarily two forms of python code input:
- `custom_action` decorator: any function decorated with `@custom_action` will be executed after the action invocation. All 
   such functions must accept a Pandas data frame and return a Pandas data frame. Functions decorated with 
    `@custom_action` are executed in the order of their definition.
- Script style: entering a script into the action code menu will execute the script. The script will have access to the 
   dataframe through the `df` variable and must mutate it before completion in order to apply the correct action.

If both are provided, the scripted logic is ran before any `@custom_action` decorated functions.

An example of a decorator style cleaning action:
```python
@custom_action
def transform(df: pandas.DataFrame) -> pandas.DataFrame:
    df['total'] = df.sum(axis=0)
    return df
```

An example of a script style cleaning action:
 ```python
 df['total'] = df.sum(axis=0)
```

# Tests
Unit tests were written and ran, alongside testing on real datasets using the client.

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_ai/tests/data_cleaner/transformer_actions/test_custom_actions.py
+++ b/mage_ai/tests/data_cleaner/transformer_actions/test_custom_actions.py
@@ -1,0 +1,189 @@
+from mage_ai.data_cleaner.transformer_actions.custom_action import execute_custom_action
+from mage_ai.tests.base_test import TestCase
+from pandas.testing import assert_frame_equal
+import pandas as pd
+import numpy as np
+
+
+class CustomActionTests(TestCase):
+    def test_decorated_custom_action(self):
+        df = pd.DataFrame(
+            [
+                ['1', 1.000, '2021-10-01', 'Store 1', 23023],
+                ['1', None, '2021-10-01', 'Store 2', np.nan],
+                [np.nan, 1100, '', '', 90233],
+                ['2', None, None, 'Store 1', 23920],
+                ['2', 12.00, '2021-09-01', None, np.nan],
+                ['2', 125.0, '2021-09-01', 'Store 3', 49833],
+            ],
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: mage_ai/tests/data_cleaner/transformer_actions/test_custom_actions.py [hunk 1]
```diff
@@ -0,0 +1,189 @@
...
+                [2.0, None, None, 'Store 1', 23920, 2],
+                [2.0, 12.00, '2021-09-01', None, np.nan, 2],
+                [2.0, 125.0, '2021-09-01', 'Store 3', 49833, 2],
+            ],
+            columns=['group_id', 'price', 'group_churned_at', 'store', 'zip_code', 'new_col'],
+        )
+        action = {
+            'action_type': 'custom',
+            'axis': 'row',
+            'action_code': 'from mage_ai.data_cleaner.column_type_detector import NUMBER,'
```
wangxiaoyou1993, line 41: Let's write multiline text in the following format
''''
from mage_ai....
....
....
'''

wangxiaoyou1993:
Let's name the decorator "transformer_function".

mage_ai/tests/data_cleaner/transformer_actions/test_custom_actions.py
skunichetty: Fixed (albeit with strange formatting in order to adjust for indent)

wangxiaoyou1993 (approved):

PR #196, opened 2022-06-06
[sk] Statistics performance updates

skunichetty:
# Summary
Updates to speed up statistics calculation. 

# Performance Metrics
Performance was evaluated on dataset with 100000 rows and 28 columns (total of 2.8 million entries) with numbers, categories, boolean, and text columns. All times are averaged over 100 trials. 
**Baseline**: 1.463 seconds
- No multithreading - 1.408 seconds (likely because statistics calculation is CPU bound rather than IO bound)
- Deduplicated null value calculation between null_value_rate and null_value_count - 1.362 seconds      
- Switched to inbuilt mean calculation - 1.344 seconds
- Switched from quantile to inbuilt median calculation - 1.323 seconds
- Refactored timeseries check to avoid manually calculating the null value rate - 1.285 seconds
- Optimizing longest null sequence:
    - Switched from pandas series to Numpy ndarray - 1.211 seconds
    - Switched to generator of null values (implicit running mean) - 1.04 seconds
- Removed duplicate max_null_sequence calculation - 0.91 seconds

**Final Time:** 0.91 seconds (~38% decrease)

# Tests
Correctness was confirmed by running the code on unit tests, and also through local testing.

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_ai/data_cleaner/statistics/calculator.py
+++ b/mage_ai/data_cleaner/statistics/calculator.py
@@ -53,7 +53,6 @@ class StatisticsCalculator:
             arr_args_1 = ([df[col] for col in df.columns],)
             arr_args_2 = ([col for col in df.columns],)
 
-            # dicts = run_parallel(self.statistics_overview, arr_args_1, arr_args_2)
             dicts = map(self.statistics_overview, *arr_args_1, *arr_args_2)
 
             for d in dicts:
--- a/mage_ai/data_cleaner/statistics/calculator.py
+++ b/mage_ai/data_cleaner/statistics/calculator.py
@@ -172,7 +172,9 @@ class StatisticsCalculator:
         }
 
         data[f'{col}/max_null_seq'] = (
-            max(self.null_seq_gen(series.isna().to_numpy())) if series.size else 0
+            max(self.null_seq_gen(series.isna().to_numpy()))
+            if data[f'{col}/count'] != 0
+            else len(series)
         )
 
         if len(series_non_null) > 0:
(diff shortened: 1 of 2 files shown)


mage_ai/data_cleaner/statistics/calculator.py
```diff
@@ -176,16 +168,19 @@ def __statistics_overview(self, series, col):
             if series.size == 0
             else series.isnull().sum() / series.size,
             f'{col}/null_value_count': series.isnull().sum(),
-            f'{col}/max_null_seq': self.get_longest_null_seq(series),
             f'{col}/value_counts': df_top_value_counts.to_dict(),
         }
 
+        data[f'{col}/max_null_seq'] = (
```
wangxiaoyou1993, line 174: do we need to calculate this for columns with all null values?

wangxiaoyou1993:
You can also reduce the duplicate `value_counts()` calculation.

mage_ai/data_cleaner/statistics/calculator.py
skunichetty: Don't have to calculate, should've been `count()` instead of `size` there. Fixed now.

skunichetty:
I deduplicated the value counts call when computing the mode, reducing the runtime to 0.7568003163 when averaged over 100 trials.

wangxiaoyou1993 (approved):

mage_ai/data_cleaner/statistics/calculator.py
```diff
@@ -211,18 +208,22 @@ def __statistics_overview(self, series, col):
...
             if column_type not in NUMBER_TYPES:
                 if dates is not None:
-                    value_counts = dates.value_counts()
-                else:
-                    value_counts = series_non_null.value_counts()
-
-                mode = value_counts.index[0]
-                if column_type == DATETIME:
+                    df_value_counts = dates.value_counts()
+                    data[f'{col}/value_counts'] = (
```
wangxiaoyou1993, line 212: For datetime column, we need to convert the index to str. Otherwise, there'll be json serialization issue when writing stats to disk.
https://github.com/mage-ai/mage-ai/blame/master/mage_ai/data_cleaner/statistics/calculator.py#L158

skunichetty: https://github.com/mage-ai/mage-ai/pull/220

skunichetty: I accidentally tried converting the value counts instead of the index, fixed in this new PR.

PR #201, opened 2022-06-06
[nt] Add calculations for total cells and calculate percentages

nathaniel-mage:
# Summary

### Show percentage of invalid and missing values in Metrics table
- Adds a new parameter to the parse (needs column types to get number of columns)
- Handles edge case of 0 rows or cols, but 0 rows is more likely.

# Tests

Launched on Localhost 

### Blank dataset
image

### Normal dataset
image

cc:

@johnson-mage

### DIFF
--- a/mage_ai/frontend/components/datasets/overview/utils.ts
+++ b/mage_ai/frontend/components/datasets/overview/utils.ts
@@ -16,7 +16,7 @@ export function createMetricsSample(statistics, colTypes) {
   const stats = Object.keys(statistics);
   const types = Object.values(colTypes);
   const metricRows = Array(METRICS_KEYS.length).fill(0);
-  const total_cells = (statistics?.count === 0 || types?.length === 0)
+  const totalCells = (statistics?.count === 0 || types?.length === 0)
     ? 1 : statistics?.count * types?.length;
   stats.map((key) => {
     if (METRICS_KEYS.includes(key)) {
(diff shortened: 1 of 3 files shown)


nathaniel-mage:
Note that if you'd like for these metrics to also be shown with a progress bar you can update the constants file to include `  'total_invalid_value_count',
  'total_null_value_count'` inside `PERCENTAGE_KEYS`.

mage_ai/frontend/components/datasets/overview/utils.ts
```diff
@@ -7,14 +7,17 @@ import {
...
   WARN_KEYS,
 } from '../constants';
 
-export function createMetricsSample(statistics) {
+export function createMetricsSample(statistics, colTypes) {
   const stats = Object.keys(statistics);
+  const types = Object.values(colTypes);
   const metricRows = Array(METRICS_KEYS.length).fill(0);
-
+  const total_cells = (statistics?.count === 0 || types?.length === 0)
```
johnson-mage, line 19: camelCase

johnson-mage (approved):
check comment

mage_ai/frontend/components/datasets/overview/utils.ts
nathaniel-mage: ah good catch will fix.

PR #212, opened 2022-06-07
[sk] Fixed bugs with column type detector

skunichetty:
# Summary
Fixed two bugs with column type detection:
- Object types sometimes contained mixed types and so string functions failed to work on these objects. Fixed by casting object columns to string
- `infer_column_types` sometimes (incorrectly) identified a column as a phone number due being formatted as a phone number, despite column names suggesting otherwise. To resolve this problem, phone number and zip code detection now must match column names against a set of reserved keywords most often associated with phone number and zip code columns in order for the inference to report those columns as phone numbers or zip codes.

# Tests
Previous unit tests were reran to confirm correctness. New tests were written to test the new phone number and zip code detection system.

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_ai/data_cleaner/column_type_detector.py
+++ b/mage_ai/data_cleaner/column_type_detector.py
@@ -55,6 +55,16 @@ REGEX_PHONE_NUMBER = re.compile(REGEX_PHONE_NUMBER_PATTERN)
 REGEX_ZIP_CODE_PATTERN = r'^\d{3,5}(?:[-\s]\d{4})?$'
 REGEX_ZIP_CODE = re.compile(REGEX_ZIP_CODE_PATTERN)
 
+RESERVED_PHONE_NUMBER_WORDS = frozenset(['phone', 'landline'])
+RESERVED_ZIP_CODE_WORDS = frozenset(['zip', 'postal'])
+
+
+def str_in_set(string, string_set):
+    for entry in string_set:
+        if entry in string:
+            return True
+    return False
+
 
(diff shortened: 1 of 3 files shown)


mage_ai/data_cleaner/column_type_detector.py
```diff
@@ -55,6 +55,16 @@
 REGEX_ZIP_CODE_PATTERN = r'^\d{3,5}(?:[-\s]\d{4})?$'
 REGEX_ZIP_CODE = re.compile(REGEX_ZIP_CODE_PATTERN)
 
+RESERVED_PHONE_NUMBER_WORDS = frozenset(['phone', 'landline'])
+RESERVED_ZIP_CODE_WORDS = frozenset(['zip', 'postal', 'zipcode', 'postcode'])
+
+
+def str_in_set(string, string_set):
+    for entry in string_set:
```
wangxiaoyou1993, line 63: return any(entry in string for entry in string_set)

mage_ai/data_cleaner/column_type_detector.py
```diff
@@ -139,11 +149,16 @@ def infer_object_type(series, kwargs):
             if clean_series.str.contains(REGEX_FLOAT_NEW_SYM).sum():
                 mdtype = NUMBER_WITH_DECIMALS
             else:
+                lower = column_name.lower()
```
wangxiaoyou1993, line 152: nit: use a more descriptive name like column_name_lower

wangxiaoyou1993 (approved):

PR #239, opened 2022-06-07
[sk] Fixed datetime detection bug

skunichetty:
# Summary
Fixed bug where certain datetime formats were not being detected, such as "DD/MM/YYYY HH:MM:SS". Fixed by updating the regex string to handle this case.

# Tests
Unit tests were run to confirm correctness of code, alongside testing on the dataset associated with this bug to confirm the date is now parsed.

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_ai/data_cleaner/column_type_detector.py
+++ b/mage_ai/data_cleaner/column_type_detector.py
@@ -39,7 +39,7 @@ COLUMN_TYPES = frozenset(
     ]
 )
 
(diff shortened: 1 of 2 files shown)


wangxiaoyou1993:
Did you add a new unit test for this format?

skunichetty:
I did not modify the unit tests, I can quickly add a new test

wangxiaoyou1993 (approved):

PR #250, opened 2022-06-08
[nt]  column metric additions

nathaniel-mage:
# Summary

### Improve the column reports page
- Refactor cell and tables for undefined values
- Add progress bar to percentage metrics
- Add more quality metrics to the page.

# Tests

Went on localhost
### Long table
image

### Short Table
image

cc: @johnson-mage 

Note this has big changes to design as it removes the border from the cells for the datatable.

### DIFF
--- a/mage_ai/frontend/components/datasets/columns/ColumnReports.tsx
+++ b/mage_ai/frontend/components/datasets/columns/ColumnReports.tsx
@@ -11,6 +11,7 @@ import { COLUMN_TYPE_HUMAN_READABLE_MAPPING, COLUMN_TYPE_NUMBERS } from '@interf
 import { PADDING_UNITS } from '@oracle/styles/units/spacing';
 import { getFeatureSetStatistics } from '@utils/models/featureSet';
 import { getPercentage, transformNumber } from '@utils/number';
+import { roundNumber } from '@utils/string';
 
 type ColumnReportsProps = {
   column: string;
@@ -90,22 +91,22 @@ function ColumnReports({
     },
     {
       columnValues: [
-        'Max value', max, [false],
+        'Max value', roundNumber(max), [false],
       ],
     },
     {
       columnValues: [
-        'Min value', min, [false],
+        'Min value', roundNumber(min), [false],
       ],
     },
     {
       columnValues: [
-        'Median value', median, [false],
+        'Median value', roundNumber(median), [false],
       ],
     },
     {
       columnValues: [
-        'Average value', average, [false],
+        'Average value', roundNumber((average)), [false],
       ],
     },
     // {
--- a/mage_ai/frontend/components/datasets/columns/ColumnReports.tsx
+++ b/mage_ai/frontend/components/datasets/columns/ColumnReports.tsx
@@ -39,6 +39,7 @@ function ColumnReports({
     max,
     median,
     min,
+    // mode,
     null_value_count: nullValueCount,
     outlier_count: outlierCount,
     outliers,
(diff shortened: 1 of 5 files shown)


mage_ai/frontend/components/datasets/columns/ColumnReports.tsx
```diff
@@ -29,10 +29,15 @@ function ColumnReports({
...
   const {
+    average,
     completeness,
     count,
     count_distinct: countDistinct,
     invalid_value_count: invalidValueCount,
+    max,
+    median,
+    min,
+    // mode,
```
johnson-mage, line 40: Is `mode` purposely left out for a reason?

mage_ai/frontend/oracle/components/Table/Cell.tsx
```diff
@@ -117,8 +119,10 @@ function Cell({
         first={cellIndex === 0}
         showBackground={showBackground}
         small={small}
+        vanish={vanish}
       >
-        {cellEl}
+        { !vanish && <>{cellEl} </>
```
johnson-mage, line 124: wrapping `cellEl` in the empty brackets seems unnecessary

johnson-mage:
When there are a lot of values in the Outliers table, the Column Summary table has excessive white space. We want to borders to fit the children content. 
image

johnson-mage:
Average value should be rounded.

johnson-mage:
The second column gets cut off in Column Summary even on full screens. We can change the proportion of the screen that the Warnings/Outliers tables are taking up.

mage_ai/frontend/components/datasets/columns/ColumnReports.tsx
```diff
@@ -48,38 +53,64 @@ function ColumnReports({
...
+      ],
+    },
+    {
+      columnValues: [
+        'Average value', average, [false],
+      ],
+    },
+    // {
+    //   columnValues: [
+    //     'Mode value', mode, [false],
```
johnson-mage, line 111: Same here. Mode commented out?

johnson-mage:
SimpleDataTable looks very similar visually to RowDataTable, and refactoring it might work now to fit what you need it for, but it seems like it's quickly becoming unmaintainable. A better solution seems like it would have been to just use RowDataTable instead. If those comments I left can be fixed quickly, then we can merge it in, but the excessive empty space at the bottom isn't something we'd want in there.

mage_ai/frontend/components/datasets/columns/ColumnReports.tsx
nathaniel-mage: Yeah it didn't work so I thought we could do the metric later. My backends not returning it.

mage_ai/frontend/components/datasets/columns/ColumnReports.tsx
nathaniel-mage: Intentional, see commit comment.

johnson-mage (approved):
approved after addressing comments. there shudnt be the excess white space in the tables

johnson-mage:
> I couldn't get them to not appear, so I instead will display them as NaN. image
> 
> Also I reverted my border changes as they broke the tables on other pages that encapsulate them inside another component.

@nathaniel-mage Why not "N/A" instead of NaN?

johnson-mage:
> I couldn't get them to not appear, so I instead will display them as NaN. image
> 
> Also I reverted my border changes as they broke the tables on other pages that encapsulate them inside another component.

@nathaniel-mage Not sure if you're addressing this in another PR already, but the bottom corners are also clipped.

PR #253, opened 2022-06-08
[sk] Updated filtering code to accept actions with dirty formatting

skunichetty:
# Summary
Any filter action (including row removal imputation and outlier removal) now accepts dirty column names. More specifically, here are the rules for writing filtering code using column names now:
- If column names contain only alphanumeric characters and the special character '_', then the raw column name can used to filter code. Example: `age != null and price_of_item >= 50.23`
- If columns contain whitespace, newlines, and some of the following special characters: `+=-*&^%$! ?~|<>(){}[],.`, then the quoted column name can be used to filter code. Any quote character can be used. Example: `("+=  -*&^%$! ?~  |<>" == False and ' kas22d fe ($)' > 50.23) or ("dis>>> ??cou nted" contains "partial")`.
- If columns contain any other special characters apart from those listed above, the column cannot be filtered. This is a fundamental limitation of pandas, see [here](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.query.html?highlight=query#pandas.DataFrame.query:~:text=During%20parsing%20a,not%20be%20escaped) for more information.

As a result of these changes, the dependency resolution checking for the filter action has been removed. While the dependency resolution infrastructure remains, the check for valid column names (which was buggy to begin with) is no longer performed.

The cleaning suggestions for impute values and remove outliers were updated

# Tests
Unit tests were written to make sure that (a) old filtering code still works and (b) new filtering with new column names also works. Unit tests were also written to detect when a column name needs wrapping in quotes. In addition, filtering code was testing using sample datasets to make sure that filtering dirty column names worked. Here is a sample of the successful filtering using the frontend (tested on `salaries.csv`):
image

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_ai/data_cleaner/cleaning_rules/shared.py
+++ b/mage_ai/data_cleaner/cleaning_rules/shared.py
@@ -1,0 +1,7 @@
+COLUMN_NAME_QUOTE_CHARS = '+=-*&^%$! ?~|<>(){}[],.'
+
+
+def wrap_column_name(name: str) -> str:
+    if any(symbol in name for symbol in COLUMN_NAME_QUOTE_CHARS):
+        name = f'"{name}"'
+    return name
--- a/mage_ai/data_cleaner/cleaning_rules/shared.py
+++ b/mage_ai/data_cleaner/cleaning_rules/shared.py
@@ -1,7 +1,0 @@
-COLUMN_NAME_QUOTE_CHARS = '+=-*&^%$! ?~|<>(){}[],.'
-
-
-def wrap_column_name(name: str) -> str:
-    if any(symbol in name for symbol in COLUMN_NAME_QUOTE_CHARS):
-        name = f'"{name}"'
-    return name
--- a/mage_ai/data_cleaner/transformer_actions/action_code.py
+++ b/mage_ai/data_cleaner/transformer_actions/action_code.py
@@ -2,7 +2,7 @@ from mage_ai.data_cleaner.transformer_actions.constants import Operator
 import re
 
 ACTION_CODE_NAME = r'(?:([^\s()"\']+|[\'\"][^"\']+[\'\"]))'
-ACTION_CODE_CONDITION_PATTERN = re.compile(
+ACTION_CODE_PATTERN = re.compile(
     rf'{ACTION_CODE_NAME} ([!=<>]+|(?:contains)|(?:not contains)) {ACTION_CODE_NAME}'
 )
 ORIGINAL_COLUMN_PREFIX = 'orig_'
@@ -11,12 +11,12 @@ TRANSFORMED_COLUMN_PREFIX = 'tf_'
 
 
 def append_prefix(column_name, prefix):
-    isQuoted = False
+    is_quoted = False
     if column_name[0] == '\"' or column_name[0] == '\'':
-        isQuoted = True
+        is_quoted = True
         column_name = column_name.strip(QUOTES)
     column_name = f'{prefix}{column_name}'
-    if isQuoted:
+    if is_quoted:
         column_name = f'`{column_name}`'
     return column_name
 
@@ -80,7 +80,7 @@ def query_with_action_code(df, action_code, kwargs):
     column_set = set(df.columns)
 
     prev_end = 0
-    for match in ACTION_CODE_CONDITION_PATTERN.finditer(action_code):
+    for match in ACTION_CODE_PATTERN.finditer(action_code):
         column_name, operator, value = match.groups()
         column_name = column_name.strip(QUOTES)
         reconstructed_code.append(action_code[prev_end : match.start()])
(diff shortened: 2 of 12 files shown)


mage_ai/data_cleaner/cleaning_rules/shared.py
```diff
@@ -0,0 +1,7 @@
+COLUMN_NAME_QUOTE_CHARS = '+=-*&^%$! ?~|<>(){}[],.'
```
wangxiaoyou1993, line 1: It'll be better to move shared methods/files to this folder: https://github.com/mage-ai/mage-ai/tree/master/mage_ai/data_cleaner/shared

mage_ai/data_cleaner/transformer_actions/action_code.py
```diff
@@ -1,18 +1,30 @@
...
-    r'([^\s()]+) ([!=<>]+|(?:contains)|(?:not contains)) ([^\s()]+)'
+    rf'{ACTION_CODE_NAME} ([!=<>]+|(?:contains)|(?:not contains)) {ACTION_CODE_NAME}'
 )
 ORIGINAL_COLUMN_PREFIX = 'orig_'
+QUOTES = '\"\''
 TRANSFORMED_COLUMN_PREFIX = 'tf_'
 
 
+def append_prefix(column_name, prefix):
+    isQuoted = False
```
wangxiaoyou1993, line 14: nit: snake case

mage_ai/data_cleaner/transformer_actions/dependency_resolution.py
```diff
@@ -1,5 +1,5 @@
 from pandas import DataFrame
-from typing import Tuple
+from typing import Tuple, Dict
```
wangxiaoyou1993, line 2: nit: alphabetical order, `Dict, Tuple`

mage_ai/data_cleaner/transformer_actions/dependency_resolution.py
```diff
@@ -10,11 +10,19 @@
...
 
 
-def default_resolution(df: DataFrame) -> Tuple[bool, str]:
+def default_resolution(df: DataFrame, action: Dict) -> Tuple[bool, str]:
     return True, None
 
 
-def resolve_filter_action(df: DataFrame) -> Tuple[bool, str]:
+def isolate(action_code):
+    split_code = action_code.split(" and ")
```
wangxiaoyou1993, line 18: nit: single quote

mage_ai/data_cleaner/transformer_actions/dependency_resolution.py
```diff
@@ -10,11 +10,19 @@
 """
 
 
-def default_resolution(df: DataFrame) -> Tuple[bool, str]:
+def default_resolution(df: DataFrame, action: Dict) -> Tuple[bool, str]:
     return True, None
 
 
-def resolve_filter_action(df: DataFrame) -> Tuple[bool, str]:
+def isolate(action_code):
```
wangxiaoyou1993, line 17: where is it used?

mage_ai/data_cleaner/transformer_actions/action_code.py
```diff
@@ -1,18 +1,30 @@
 from mage_ai.data_cleaner.transformer_actions.constants import Operator
 import re
 
-
+ACTION_CODE_NAME = r'(?:([^\s()"\']+|[\'\"][^"\']+[\'\"]))'
```
wangxiaoyou1993, line 4: nit: alphabetical order

mage_ai/data_cleaner/transformer_actions/base.py
```diff
@@ -76,9 +76,9 @@ def axis(self):
     def execute(self, df, **kwargs):
         action_type = self.action['action_type']
         dependency = DEPENDENCIES.get(action_type, default_resolution)
-        dependencies_met, msg = dependency(df)
-        if not dependencies_met:
-            raise RuntimeError(f'Dependencies of this cleaning action are not completed: {msg}')
+        dependencies_met, msg = dependency(df, self.action)
```
wangxiaoyou1993, line 79: Shall we make DEPENDENCIES an empty dictionary?

wangxiaoyou1993 (approved):

mage_ai/data_cleaner/transformer_actions/base.py
skunichetty: Yeah, I can make it empty and uncomment the error message so in the future the dependencies will work correctly.

mage_ai/data_cleaner/transformer_actions/dependency_resolution.py
skunichetty: It's currently not used, I forgot to remove, will remove now.

PR #254, opened 2022-06-08
[sp] fix suggestions overflow

shrey-mage:
# Summary
- extremely long column name links now wrap instead of overflowing `SuggestionRow`

# Tests

There's still an issue with "Remove outliers" because it needs to display the super long column name variable, which I'm not sure if we can do much about. It should be a very rare edge case.

image

cc: @tommydangerous @johnson-mage

### DIFF
--- a/mage_ai/frontend/oracle/elements/Link/index.tsx
+++ b/mage_ai/frontend/oracle/elements/Link/index.tsx
@@ -30,6 +30,8 @@ export type LinkProps = {
   href?: string;
   inline?: boolean;
   large?: boolean;
+  minWidth?: string | number;
+  maxWidth?: string | number;
   muted?: boolean;
   noColor?: boolean;
   noHoverUnderline?: boolean;
@@ -51,7 +53,6 @@ export type LinkProps = {
   transparentBorder?: boolean;
   underline?: boolean;
   weightStyle?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8;
-  width?: number;
   wordWrap?: boolean;
 };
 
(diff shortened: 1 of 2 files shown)


mage_ai/frontend/oracle/elements/Link/index.tsx
```diff
@@ -202,15 +203,16 @@ export const SHARED_LINK_STYLES = css<any>`
     text-overflow: ${props.textOverflow};
   `}
 
-  ${props => props.width && `
+  ${props => (props.minWidth || props.maxWidth) && `
```
johnson-mage, line 206: Are these props needed? `SHARED_TEXT_STYLES` already has minWidth and maxWidth props.

johnson-mage: And `LinkStyle` contains `SHARED_TEXT_STYLES`.

johnson-mage: The only difference is the existing props in `mage-ai/mage_ai/frontend/oracle/elements/Text/index.tsx` does not contain `display: block`, so maybe whatever Link component had the issue needs a `block` prop.

shrey-mage: oh I didn’t realize that it uses shared styles, the only thing is that Link needs `display: block` like you said

shrey-mage: anyways we can’t use `display: block` for `SuggestionsRow` bc it breaks the design of inline links, so it was more of a fix for the future. I’ll remove and only add `display: block`

johnson-mage: We have a `block` prop in `Link`, so check that.

johnson-mage (approved):
approved after removing redundant props

PR #288, opened 2022-06-08
[nt] Change example placeholder to editable text

nathaniel-mage:
# Summary

### Change the custom code placeholder to be default value
- Removed placeholder
- Update `CODE_EXAMPLE` to be more instructional

# Tests

Localhost and ran some queries.
### Queries
image

### Default values (editable)
Screen Shot 2022-06-08 at 12 58 38 PM

_You can tell it's a value because it's colored_

cc: @johnson-mage 

Is this what you were thinking of @tommydangerous? Or did you want to display it somewhere else?

### DIFF
--- a/mage_ai/frontend/components/ActionForm/actions.ts
+++ b/mage_ai/frontend/components/ActionForm/actions.ts
@@ -25,7 +25,7 @@ const columns: {
       Add @transformer_action, then begin typing Python code inside a function. \
       You may refer to the current dataset as df.',
     code: {
-      custom: CODE_EXAMPLE,
+      default: CODE_EXAMPLE,
       multiline: true,
       values: USER_INPUT,
     },
@@ -144,7 +144,7 @@ const rows: {
       Add @transformer_action, then begin typing Python code inside a function. \
       You may refer to the current dataset as df.',
     code: {
-      custom: CODE_EXAMPLE,
+      default: CODE_EXAMPLE,
       multiline: true,
       values: USER_INPUT,
     },
--- a/mage_ai/frontend/components/ActionForm/actions.ts
+++ b/mage_ai/frontend/components/ActionForm/actions.ts
@@ -25,6 +25,7 @@ const columns: {
       Add @transformer_action, then begin typing Python code inside a function. \
       You may refer to the current dataset as df.',
     code: {
+      custom: CODE_EXAMPLE,
       multiline: true,
       values: USER_INPUT,
     },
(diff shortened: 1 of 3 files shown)


nathaniel-mage:
Found another bug, will fix don't review yet.

nathaniel-mage:
### Fixed Filter placeholder and display bug
image

nathaniel-mage:
And it still works on custom actions
image

Additionally, if you want to add your own default value you can include it.

mage_ai/frontend/components/ActionForm/actions.ts
```diff
@@ -25,6 +25,7 @@ const columns: {
       Add @transformer_action, then begin typing Python code inside a function. \
       You may refer to the current dataset as df.',
     code: {
+      default: CODE_EXAMPLE,
```
johnson-mage, line 28: if this `custom` property is the exact same in columns and rows constants, let's create a separate variable for it and  use it in both places. It's a hassle to have to update the code in two places every time we make a change.

johnson-mage (approved):
approved after addressing comment

mage_ai/frontend/components/ActionForm/actions.ts
nathaniel-mage: That's what code example is for, or am i misunderstanding what you mean?

johnson-mage: meant the whole `custom` prop: 
```
custom: {
    title: 'Custom code',
    description: 'Write your own custom cleaning function. \
      Add @transformer_action, then begin typing Python code inside a function. \
      You may refer to the current dataset as df.',
    code: {
      default: CODE_EXAMPLE,
      multiline: true,
      values: USER_INPUT,
    },
```

nathaniel-mage: One is for the dataset page and the other is for column specific page

nathaniel-mage: Should I write different copy for them then?

nathaniel-mage: This is the logic that selects which to look at: (axis === AxisEnum.ROW ? actions.rows : actions.columns)?.[actionType];
Adding a third variable might break it.

nathaniel-mage:
Merged but adding in the enhancement. Didn't include this in merge due to type errors resulting in failed build.

PR #291, opened 2022-06-08
[sk] Reformat values doesn't fail on bad input datatypes

skunichetty:
# Summary
Updated reformat action to accept all datatypes, but only perform the reformatting if the datatype is correct. 

Note: unit tests were added alongside reformatting them, causing a very messy git blame.

# Tests
Changes were tested locally on sample datasets by trying to apply custom actions to incorrect datatypes. Unit tests were also written to confirm that there is no failure on inputs with the wrong datatype.

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_ai/data_cleaner/transformer_actions/column.py
+++ b/mage_ai/data_cleaner/transformer_actions/column.py
@@ -13,6 +13,7 @@ from mage_ai.data_cleaner.transformer_actions.helpers import (
     get_time_window_str,
 )
 from mage_ai.data_cleaner.transformer_actions.udf.base import execute_udf
+from mage_ai.data_cleaner.transformer_actions.utils import generate_string_cols
 from keyword import iskeyword
 import pandas as pd
 import numpy as np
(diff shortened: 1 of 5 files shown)


wangxiaoyou1993:
what change did you make to analysis calculator?

mage_ai/data_cleaner/transformer_actions/column.py
```diff
@@ -147,21 +147,33 @@ def reformat(df, action, **kwargs):
     reformat_action = options['reformat']
     df.loc[:, columns] = df[columns].replace('^\s*$', np.nan, regex=True)
 
+    def gen_string_cols(columns):
```
wangxiaoyou1993, line 150: it can be put into a shared util file.

mage_ai/data_cleaner/transformer_actions/column.py
```diff
@@ -147,21 +147,33 @@ def reformat(df, action, **kwargs):
...
+            clean_col = df[column]
+            dropped = clean_col.dropna(axis=0)
+            exact_dtype = type(dropped.iloc[0]) if len(dropped) > 0 else None
+            if exact_dtype is str:
+                yield column
+
     if reformat_action == 'caps_standardization':
         capitalization = options['capitalization']
-        for column in columns:
+        for column in gen_string_cols(columns):
```
wangxiaoyou1993, line 160: Let's print out something for column that's not string type.

skunichetty:
Analysis calculator failed to generate time series charts for columns full of `pd.NaT` values (which happens often when poor conversion from a nondate column to a date occurs), which causes errors to be raised. I circumvented that change by adding a check to whether the charts generated are null, and if so skipping adding those null-valued charts.

Statistics calculator failed with a more general bug that after the mode for a datetime column is parsed to an ISO-formatted string, the mode cannot be used as a key in the value_counts dictionary. There are many ways to fix this error, I decided to not actually convert the `mode` variable to string but instead perform the conversion to ISO-formatting string only upon writing to statistics.

wangxiaoyou1993 (approved):

PR #294, opened 2022-06-08
[jk] Store custom code from transformer actions

johnson-mage:
# Summary
- See title.

# Tests
store custom code

cc:
@wangxiaoyou1993


wangxiaoyou1993:
Do we store it per feature set?

johnson-mage:
> Do we store it per feature set?

oh, we do not. let me update that

PR #325, opened 2022-06-10
[nt] ace code editor polish

nathaniel-mage:
# Summary

### Improvements to the code editor code
- Refactor the repeated code into a const file
- Flesh out more of the parameters
# Tests

### Export pages (no highlighting lines/gutter)
image

### Suggestions List
image
image

### Custom Code Actions
image

### Known bugs
- When going from custom to filter without closing the form it doesn't change the values
### Notes on implementation
- Due to the nature of NextJS and the dynamic call, are we able to change the Code Editor into a component instead?
- It feels really annoying to have to create a new dynamic function, import the same builds in order to get it 
- Also, I was unable to get the cursor to completely hide, but I made it slimmer so hopefully it's less noticeable. It still will highlight/box the current text it's on though, like a real code editor.

cc: @johnson-mage


nathaniel-mage:
The bug is on master not just this branch.

johnson-mage:
> are we able to change the Code Editor into a component instead?

why don't you try this?

johnson-mage:
> When going from custom to filter without closing the form it doesn't change the values

This is a pretty important bug to fix. did you add a story?

johnson-mage:
The cursor doesn't look any less visible after making it slimmer, but ideally, we still wanna hide it initially. Add a story for that if you havent already.

johnson-mage (approved):

mage_ai/frontend/components/datasets/Export/ace.style.tsx
```diff
@@ -0,0 +1,18 @@
+import styled, { css } from 'styled-components';
+
+import light from '@oracle/styles/themes/light';
+import { BORDER_RADIUS, BORDER_RADIUS_SMALL } from '@oracle/styles/units/borders';
+import { UNIT } from '@oracle/styles/units/spacing';
+
+
+export const AceStyle = css<any>`
```
johnson-mage, line 8: Is this used anywhere? If not, why do we have it?

johnson-mage: @nathaniel-mage

nathaniel-mage: I was experimenting with trying to access its children from the package. It didn't work. Will delete the file.

PR #327, opened 2022-06-10
[sk] Adding syntax error cleaning action

skunichetty:
# Summary
Added transformer action to clean syntax errors by replacing invalid values with an 'invalid' marker. Some notable changes:
- `find_syntax_errors` now supports cleaning integers and datetimes (however as of now they still can't be used with the cleaning suggestion since the dataset gets automatically cleaned).
- `find_syntax_errors` filters out null values (null values are null, not invalid)
- Added `invalid_value_distribution` entry in statistics which can be used to generate a visualization on the distribution of the invalid values in a column

# Tests
Unit tests were written to test that invalid values are recognized and replaced properly, and also the backend was tested with API calls to the pipeline to remove syntax errors.

cc:
@wangxiaoyou1993 @nathaniel-mage

### DIFF
--- a/mage_ai/data_cleaner/column_type_detector.py
+++ b/mage_ai/data_cleaner/column_type_detector.py
@@ -41,7 +41,7 @@ COLUMN_TYPES = frozenset(
 
 REGEX_DATETIME_PATTERN = r'^\d{2,4}-\d{1,2}-\d{1,2}$|^\d{2,4}-\d{1,2}-\d{1,2}[Tt ]{1}\d{1,2}:\d{1,2}[:]{0,1}\d{1,2}[\.]{0,1}\d*|^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$|^\d{1,4}[-\/]{1}\d{1,2}[-\/]{1}\d{1,2}$|^\d{1,2}[-\/]{1}\d{1,2}[-\/]{1}\d{1,4}$|^\d{1,2}[-\/]\d{1,2}[-\/]\d{2,4}[Tt ]{1}\d{1,2}:\d{1,2}[:]{0,1}\d{1,2}[\.]{0,1}\d*|(Jan(?:uary)?|Feb(?:ruary)?|Mar(?:ch)?|Apr(?:il)?|May|Jun(?:e)?|Jul(?:y)?|Aug(?:ust)?|Sep(?:tember)?|Oct(?:ober)?|Nov(?:ember)?|Dec(?:ember)?)\s+(\d{1,2})[\s,]+(\d{2,4})'
 REGEX_DATETIME = re.compile(REGEX_DATETIME_PATTERN)
-REGEX_EMAIL_PATTERN = r'^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$'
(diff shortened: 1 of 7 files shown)


mage_ai/data_cleaner/column_type_detector.py
```diff
@@ -66,17 +66,22 @@ def str_in_set(string, string_set):
...
-        return ~series.str.match(REGEX_EMAIL)
+        pattern = REGEX_EMAIL
     elif column_type == PHONE_NUMBER:
-        return ~series.str.match(REGEX_PHONE_NUMBER)
+        pattern = REGEX_PHONE_NUMBER
     elif column_type == ZIP_CODE:
-        str_series = series.astype(str)
-        return ~str_series.str.match(REGEX_ZIP_CODE)
+        pattern = REGEX_ZIP_CODE
+    elif column_type == NUMBER or column_type == NUMBER_WITH_DECIMALS:
```
wangxiaoyou1993, line 76: nit: column_type in NUMBER_TYPES

mage_ai/data_cleaner/transformer_actions/column.py
```diff
@@ -1,5 +1,6 @@
 from mage_ai.data_cleaner.column_type_detector import (
     DATETIME,
+    find_syntax_errors,
```
wangxiaoyou1993, line 3: nit: move it below constants

mage_ai/data_cleaner/transformer_actions/column.py
```diff
@@ -74,6 +75,16 @@ def first(df, action, **kwargs):
     return __agg(df, action, 'first')
 
 
+def fix_syntax_errors(df, action, **kwargs):
+    action_variables = action['action_variables']
+    columns = action['action_arguments']
+    for column in generate_string_cols(df, columns):
+        dtype = action_variables[column]['feature']['column_type']
+        mask = find_syntax_errors(df[column], dtype)
+        df.loc[mask, column] = 'invalid'
```
wangxiaoyou1993, line 84: for number, let's set it to NaN

skunichetty:
Note: in this latest push I refactored some code both for this cleaning action and the constant imputation cleaning action as they were both similar, and in hindsight there is a much simpler method to perform their logic.

mage_ai/data_cleaner/column_type_detector.py
skunichetty: Fixed

mage_ai/data_cleaner/transformer_actions/column.py
skunichetty: Fixed

mage_ai/data_cleaner/transformer_actions/column.py
skunichetty: Numbers now get converted to NaN.

mage_ai/data_cleaner/column_type_detector.py
```diff
@@ -66,17 +66,22 @@ def str_in_set(string, string_set):
 def find_syntax_errors(series, column_type):
     if len(series) == 0:
         return pd.Series([])
+    str_series = series.astype(str)
```
wangxiaoyou1993, line 69: just to avoid unnecessary regex matching, if the series.dtype is a number type or datetime, we don't need to do regex matching with the str.

wangxiaoyou1993 (approved):

PR #330, opened 2022-06-11
[nt]  invalid syntax detection

nathaniel-mage:
# Summary

### Highlight Invalid values on Datatable
- Cells are highlighted dark red with white text when they are invalid values.

# Tests

Localhost
### Show cells with updated danger color
image

_Checked values from the feature profiles that had red and checked the datatable for inspection.
cc: @johnson-mage

### DIFF
--- a/mage_ai/frontend/components/DataTable/index.tsx
+++ b/mage_ai/frontend/components/DataTable/index.tsx
@@ -49,6 +49,7 @@ type TableProps = {
 type DataTableProps = {
   columns: string[];
   rows: string[][] | number[][];
+  invalids: number[];
 } & SharedProps;
 
 const Styles = styled.div<{
@@ -353,6 +354,7 @@ function DataTable({
   columnHeaderHeight,
   columns: columnsProp,
   height,
+  invalids,
   renderColumnHeader,
   rows: rowsProp,
   width,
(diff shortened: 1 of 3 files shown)


mage_ai/frontend/components/DataTable/index.tsx
```diff
@@ -227,6 +235,11 @@ function Table({
             cellStyle.width = maxWidthOfFirstColumn;
           }
 
+          if (isInvalid) {
+            cellStyle.background = light.interactive.dangerBorder;
```
johnson-mage, line 239: This isn't the preferred way to style our components, hardcoding an inline style. It might work, but we should be using a React styled component with props.

mage_ai/frontend/utils/models/featureSet.ts
```diff
@@ -62,6 +62,33 @@ export function getFeatureSetStatistics(featureSet, featureUUID) {
   };
 }
 
+export function getFeatureSetInvalidValues(featureSet, featureUUID) {
```
johnson-mage, line 65: Don't we already get these stats in the `getFeatureSetStatistics` method? So calling this seems redundant. We could just add another property for `invalid_indices` since that one isn't included in the `getFeatureSetStatistics` method.

mage_ai/frontend/components/DataTable/index.tsx
nathaniel-mage: I'll see if I can do something else.

mage_ai/frontend/utils/models/featureSet.ts
nathaniel-mage: This function does a different thing than the `getFeatureSetInvalidValuesAll`. Also this function is unused and is part of the next story to display in a chart what the values are.

nathaniel-mage: I'll merge both into the getFeatureSetStatistics function

PR #343, opened 2022-06-13
[sk] Cleaning suggestion for multidimensional outliers

skunichetty:
# Summary
Added support for identifying multidimensional outliers and removing them. This is a fairly complex PR, here is the breakdown of new features:
## Outlier Identification and Removal
When the numerical data in a dataset is univariate, the previous method of removing data from 3 standard deviations and beyond the mean is used.

When the numerical data is multivariate, a different approach is used:
- If there are 2-5 numerical features, then the **[Local Outlier Factor](https://scikit-learn.org/stable/modules/outlier_detection.html#local-outlier-factor)** (LOF) algorithm is used. The LOF algorithm is used for these lower dimensional feature due its quadratic runtime, which ends up being generally faster on smaller inputs (Isolation forests have some overhead). The number of neighbors is chosen dynamically based on the number of entries in the dataset, where larger datasets use more neighbors. Using the anomaly scores produced by the algorithm, all entries with an anomaly score less than or equal to -1.5 are discarded.
- If there are 5-20 numerical features, then the **[Isolation Forest](https://scikit-learn.org/stable/modules/outlier_detection.html#isolation-forest)** (ITREE) algorithm is used. As the ITREE algorithm's runtime scales linearly with the input, this method completes much quicker on higher dimensional data than LOF. The number of estimators is fixed at 75, but can be optimized to scale dynamically as needed if performance becomes an issue. Using the anomaly scores produced by the algorithm, all entries with an anomaly score less than or equal to -0.1 are discarded.
- - If there are more than 20 such features, Principal Component Analysis (PCA) is used to embed the features into a 20 dimensional space. This is done to reduce complexity. This will likely not affect the performance of selecting outliers, as PCA selects the data representation along which variance is maximized, and outliers are most likely to arise along high variance dimensions anyway (this is the distinction between outliers and anomalies).

All thresholds specified above were determined through empirical testing by:
1. Generating data from some distribution (normal, multimodal normal, moon shaped, swiss roll)
2. Modifying some data points to be larger or smaller, making them similar to extreme values
3. Removing all values below the threshold
4. Visually inspecting which values were removed - target is to removal most of the "extreme" values and keep as much of the initial distribution as possible.

These thresholds can be modified later based on further testing.
## Other Changes
- Multidimensional outlier removal suggestion cannot give numerical boundaries to filter with as the regions of removal are nonlinear (and as such cannot be expressed using linear inequalities). To solve this, a new transformer action `remove` was added which removes rows given indices.
- Outlier removal suggestion was being infinitely added due to ITREE and LOF using relative measures of isolation to find outliers. To resolve this, two changes were made:
     - When in interactive mode (i.e., using the frontend interface), the `BasePipeline` class and `mage_ai.clean` function now take an optional parameter `rules` that specifies the cleaning suggestion rules to evaluate against. In `app.py` if a multidimensional outlier removal operation is performed, then the `RemoveOutliers` cleaning suggestion is removed from the set of rules to evaluate, preventing the suggestion from being made twice. This does not impact the case in which a univariate outlier removal cleaning suggestion is made. 
     - When directly using `mage_ai.clean`, an additional parameter was added to track if a multidimensional outlier removal operation is completed, in which case no further row removal operations are applied as transformations.

# Tests
Unit tests were written to test both the new multidimensional outlier removal functionality and the new row removal transformer action. Local testing was also completed on user data to confirm the expected behavior.

cc:
@wangxiaoyou1993


mage_ai/data_cleaner/cleaning_rules/remove_outliers.py
```diff
@@ -1,36 +1,118 @@
...
             )
+        return None
+
+    def evaluate(self):
+        suggestions = []
+        ndim = len(self.numeric_columns)
+        suggestion = None
+        if len(self.numeric_df) > 0:
+            if ndim == 1:
+                suggestion = self.one_dim_outlier_check(self.numeric_columns[0])
```
wangxiaoyou1993, line 113: we still want to provide one dimensional outlier check for all numeric columns.

skunichetty: Will we make suggestions for both multidimensional outliers and univariate outliers as well?

wangxiaoyou1993: Yes. We'll provide both suggestions.

skunichetty: Fixed, now both suggestions should be generated and suggestions for both update dynamically as outliers are removed

wangxiaoyou1993:
How long does the multidimensional outliers detection take?

skunichetty:
I've tested on some of the larger datasets, can take up to 4-5 seconds when there are ~100,000 rows and many dimensions. It takes a very long time, but it's a tradeoff for accuracy in identifying the correct outliers.

I should mention there is any alternative method that tends to be much faster, but it assumes every distribution is normally distributed, which can often lead to poor results in the case of multimodal or non-normal distributions

mage_ai/data_cleaner/cleaning_rules/remove_outliers.py
```diff
@@ -1,36 +1,119 @@
...
+            outlier_algorithm.fit(data)
+            outlier_scores = outlier_algorithm.decision_function(data)
+            outlier_mask = outlier_scores <= ITREE_ANOMALY_SCORE_THRESHOLD
+        outlier_count = outlier_mask.sum()
+        if outlier_count > 0:
+            return self._build_transformer_action_suggestion(
+                REMOVE_OUTLIERS_TITLE,
+                f'Remove {outlier_count} multidimensional outlier(s) to reduce the amount of noise in the data.',
+                ActionType.REMOVE,
+                action_options=dict(rows=df.index[outlier_mask].tolist()),
```
wangxiaoyou1993, line 102: I notice one issue of specifying row indexes in the transformer action. The transformer action won't be able to applied to a different dataset when executing the pipeline in production

skunichetty: Hrm, that presents an issue then since there is no easy way to "store" the decision boundary provided by LOF or ITREE (unless we could possibly store the model parameters themselves). They are nonlinear so there exists no easy inequality to describe the regions where outliers exist.

skunichetty: Actually, one solution would be to have the actual outlier detection and removal done in the cleaning action as well. A new transformer action such as `remove_outliers` can be created which takes in the choice of algorithm and settings for that algorithm, runs the algorithm chosen to produce anomaly scores and then remove the rows. Then the pipeline can be executed in production as the outliers to remove are found dynamically. The main downside though is that there would be duplicate computation of the outlier finding algorithms, which as discussed above can be costly.

wangxiaoyou1993: Yes, let's actually add it as a transformer action. User should be able to remove multivariate outliers through cleaning action in UI.

skunichetty: Do we want to keep the cleaning suggestion as well?

mage_ai/data_cleaner/cleaning_rules/remove_outliers.py
```diff
@@ -4,33 +4,58 @@
     ActionType,
     Axis,
 )
+from typing import Union, Dict
+import numpy as np
 
+ITREE_ANOMALY_SCORE_THRESHOLD = -0.10
```
wangxiaoyou1993, line 10: nit: remove duplicate constants and unused imports

mage_ai/data_cleaner/cleaning_rules/remove_outliers.py
```diff
@@ -4,33 +4,58 @@
...
-            std = self.statistics[f'{c}/std']
-            avg = self.statistics[f'{c}/average']
+    """
+    Checks dataframe for the existence of potential outliers, and generates a cleaning suggestion
+    to remove these outliers
+    """
+
+    def __init__(self, df, column_types, statistics):
+        super().__init__(df, column_types, statistics)
+        self.numeric_df, self.numeric_columns = self._filter_numeric_types()
```
wangxiaoyou1993, line 23: no need to filter numeric columns here anymore. it takes additional time to filter numeric types.

mage_ai/data_cleaner/transformer_actions/column.py
```diff
@@ -196,6 +197,22 @@ def remove_column(df, action, **kwargs):
     return df.drop(columns=drop_columns)
 
 
+def remove_outliers(df, action, **kwargs):
+    cols = action['action_arguments']
```
wangxiaoyou1993, line 201: Let's filter out non numerical columns here.

mage_ai/server/app.py
```diff
@@ -279,9 +280,15 @@ def update_pipeline(id):
     if feature_set_id is not None:
         feature_set = FeatureSet(id=feature_set_id)
         df_transformed = clean_pipeline.transform(feature_set.data_orig, auto=False)
+
+        # TODO: Come up with better way to support dynamic rulesets for pipelines
+        rules = [rule for rule in DEFAULT_RULES]
+        if clean_pipeline.outliers_removed:
+            rules.remove(RemoveOutliers)
```
wangxiaoyou1993, line 287: we don't need this logic here now since we don't include multidimensional outlier removal in suggestions.

mage_ai/data_cleaner/pipelines/base.py
```diff
@@ -62,11 +63,20 @@ def transform(self, df, auto=True):
...
+                action['title'] == REMOVE_OUTLIERS_TITLE
+                and payload['action_type'] == 'remove_outliers'
+                and self.outliers_removed
+            ):
+                df_transformed = BaseAction(payload).execute(df_transformed)
+                action['status'] = STATUS_COMPLETED
+                completed_queue.append(action)
+                # TODO: Come up with a better way to limit the execution of certain actions:
+                if payload['action_type'] == 'remove' and payload['axis'] == 'row':
+                    self.outliers_removed = True
```
wangxiaoyou1993, line 77: It this parameter still needed?

mage_ai/tests/data_cleaner/cleaning_rules/test_remove_outliers.py
```diff
@@ -1,6 +1,8 @@
+import statistics
```
wangxiaoyou1993, line 1: What's this library used for?

mage_ai/server/app.py
```diff
@@ -1,7 +1,8 @@
 from flask import Flask, render_template, request
 from flask_cors import CORS
+from mage_ai.data_cleaner.cleaning_rules.remove_outliers import RemoveOutliers
```
wangxiaoyou1993, line 3: Remove unused imports

mage_ai/data_cleaner/transformer_actions/column.py
```diff
@@ -196,6 +198,34 @@ def remove_column(df, action, **kwargs):
...
+        if column not in cols:
+            numeric_df.drop(column, axis=1, inplace=True)
+        else:
+            dtype = action['action_variables'][column]['feature']['column_type']
+            if dtype in NUMBER_TYPES:
+                numeric_df.loc[:, column] = numeric_df.loc[:, column].astype(float)
+            else:
+                numeric_df.drop(column, axis=1, inplace=True)
+    outlier_mask = numeric_df.notna().all(axis=1)
+    numeric_df = numeric_df.dropna(axis=0)
```
wangxiaoyou1993, line 214: Another option is to fillna with 0. Do you have any particular reason of choosing to dropna here?

skunichetty: The general problem with rows that have na values is that it is hard to infer whether they are outliers or not due to the missing information. If we fill the na values with zero and then identify that row as an outlier, it is possible that the row actually isn't an outlier and has been incorrectly removed. For most distributions which are likely not zero centered, these rows with NA values will likely get marked as outliers and removed despite there being no actual evidence to suggest that they are.

I thought that erring on the side of caution and not removing an outlier is better than accidentally removing a row that isn't an outlier, so I opted to drop the null valued rows instead.

mage_ai/data_cleaner/pipelines/base.py
skunichetty: Forgot to remove it, it is removed now

mage_ai/tests/data_cleaner/cleaning_rules/test_remove_outliers.py
skunichetty: Removed, it was auto imported unknowingly

mage_ai/server/app.py
skunichetty: Removed

mage_ai/data_cleaner/transformer_actions/column.py
wangxiaoyou1993: got it. that makes sense. the downside of dropna is if there're columns with mostly na, we'll drop the most of the rows. we can optimize for that case later.

mage_ai/data_cleaner/transformer_actions/column.py
```diff
@@ -196,6 +198,34 @@ def remove_column(df, action, **kwargs):
     return df.drop(columns=drop_columns)
 
 
+def remove_outliers(df, action, **kwargs):
+    cols = set(action['action_arguments'])
+    numeric_df = df.copy()
+    for column in df.columns:
+        if column not in cols:
+            numeric_df.drop(column, axis=1, inplace=True)
```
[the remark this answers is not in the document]
skunichetty: Fixed, didn't catch this optimization

mage_ai/server/app.py
```diff
@@ -1,7 +1,7 @@
 from flask import Flask, render_template, request
 from flask_cors import CORS
 from mage_ai.data_cleaner.data_cleaner import analyze, clean as clean_data
-from mage_ai.data_cleaner.pipelines.base import BasePipeline
+from mage_ai.data_cleaner.pipelines.base import DEFAULT_RULES, BasePipeline
```
wangxiaoyou1993, line 4: same here. remove unused imports.

wangxiaoyou1993 (approved):

mage_ai/server/app.py
wangxiaoyou1993: @skunichetty i left one last comment here.

skunichetty: Oops, didn't see. I'll fix that now.

PR #347, opened 2022-06-13
[nt] WIP reports version diff

nathaniel-mage:
# Summary

### Add the ProgressIcon component
- Small component that has the arrow+color used for displaying percentages
- Added color related props to Text and Button.
# Tests

Went on Storybook and looked at what the component looked like.

### Negative change
image

### Positive change
image

cc: @johnson-mage @dy46 

Thoughts on the current component, any styling improvements etc..
Waiting for new endpoint for percentages and SimpleTable refactor to add in the version diffs.


mage_ai/frontend/oracle/elements/Button/index.tsx
```diff
@@ -72,6 +74,15 @@ const ButtonStyle = styled.button<ButtonProps>`
...
 
+  ${props => props.negative && `
+    background-color: ${light.background.negative}; !important
+  `}
+
+  // TODO: Change to Lime 40 theme.
+  ${props => props.positive && `
+    background-color: ${light.background.positive}; !important
+  `}
+
```
dy46, line 85: i wonder if we should just add a `background-color` prop

mage_ai/frontend/oracle/components/ProgressIcon/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+  children,
+  ...props
+}: ProgressIconProps) => (
+  <ProgressIconStyle
+    {...props}
+  >
+    { props.danger
+      ?
+        <Button
+          afterIcon={<ArrowDown negative/>}
```
dy46, line 36: you should be able to pass in a fill color for the icon

nathaniel-mage:
> What's the new endpoint for percentage?

Sorry meant payload. Shrey told me that Sachit was adding a payload to get the percentage for the metric so the frontend wouldn't need to calculate this.

mage_ai/frontend/oracle/components/ProgressIcon/index.tsx
nathaniel-mage: I did that previously, if that's better I can revert to commit [1072b985d6c997c2ee01f5498d7b7cba1004e3b1](https://github.com/mage-ai/mage-ai/pull/347/commits/1072b985d6c997c2ee01f5498d7b7cba1004e3b1). I didn't know if I should be using the ThemeProps inside the component.

johnson-mage:
Arrow icon looks slightly large. Figma designs have it at 10x9. Perhaps size of 10 works better.

mage_ai/frontend/oracle/elements/Button/index.tsx
johnson-mage: I had mentioned to @nathaniel-mage and @shrey-mage to use props on the components for theming in case we add a dark theme later. Although it shud be used like `{(props.theme.background || light.background).positive};` instead of a hardcoded color.

johnson-mage: Also not sure if that `!important` is necessary

nathaniel-mage: > Arrow icon looks slightly large. Figma designs have it at 10x9. Perhaps size of 10 works better.
image

Size 10 was actually smaller. I set the Text to large to fit the shape better.

Here is Size 10 on a normal text
image

Which do you think fits better? Or should I look for a different combination?

johnson-mage:
> > Arrow icon looks slightly large. Figma designs have it at 10x9. Perhaps size of 10 works better.
> 
> image
> 
> @johnson-mage Size 10 was actually smaller. I set the Text to large to fit the shape better.
> 
> Here is Size 10 on a normal text image
> 
> Which do you think fits better? Or should I look for a different combination?

Try to match the figma designs. Normal text since that shud match whatever we use in the tables. And the icon is slightly different which is why I guess it's a bit smaller at a size of 10. Try a 10.5 or 11 then.

mage_ai/frontend/oracle/elements/Button/index.tsx
nathaniel-mage: For this one, I wasn't sure if we needed to add the props.theme... since everything else in Button doesn't have it. I'll add it as a story to clean up all the button styles.

johnson-mage: @nathaniel-mage Just because it was done that way for the other props doesn't necessarily mean it was the right way to do it.

nathaniel-mage:
image

image

More icon sizes

mage_ai/frontend/components/datasets/StatsTable/index.tsx
```diff
@@ -56,11 +60,23 @@ function StatsTable({ stats, title }: StatsTableProps) {
...
+            {progress
+              ?
+                <ProgressBar
+                  progress={rate * 100}
+                  {...warn}
+                />
+              :
+                <> </>
+            }
+            { change
```
johnson-mage, line 72: remove space

mage_ai/frontend/components/datasets/StatsTable/index.tsx
```diff
@@ -56,11 +60,23 @@ function StatsTable({ stats, title }: StatsTableProps) {
...
-                {...warn}
-              />
+            {progress
+              ?
+                <ProgressBar
+                  progress={rate * 100}
+                  {...warn}
+                />
+              :
+                <> </>
```
johnson-mage, line 70: Are these react fragments needed to maintain the column proportions?

mage_ai/frontend/components/datasets/overview/index.tsx
```diff
@@ -106,6 +109,107 @@ function DatasetOverview({
...
+        if (result !== 0) {
+          a[k] = result;
+        }
+        return a;
+      }, {});
+      setChanges(metricChanges);
+    }
+  }, [originalStatistics, showColumnsFromUrl, statistics]);
+
+  const qualityMetrics: StatRow[] = [
```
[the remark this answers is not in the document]
johnson-mage: similar to how we did for `createMetricsSample`

johnson-mage:
> Some changes I'm thinking of is to have color not match the arrow. For instance, when duplicate rows or invalid rows go up then it looks weird to see a green background.

Yeah, the color of the progress icon/background should be green if the change indicates something good (regardless of whether it's pointing up or down), and red if the change indicates something bad. So i would think an increase in "invalid cells" is bad and should be red, and a decrease in missing cells/values is good and should be green.

mage_ai/frontend/oracle/icons/BaseIcon.tsx
```diff
@@ -10,7 +10,9 @@ export type BaseIconProps = {
   earth?: boolean;
   fill?: string;
   muted?: boolean;
+  danger?: boolean;
```
tommydangerous, line 13: Sort these

tommydangerous (approved):

PR #351, opened 2022-06-14
[sk] Add more rate statistics

skunichetty:
# Summary
This PR adds more ratio computations as an added convenience for the frontend. Previously, the frontend would have to manually calculate percentages and ratios to display, including edge cases due to unclear access over row count (the number of non-null rows) and column count (the number of non-null entries per column). The backend now automatically calculates these statistics to simplify the front end code.

Adding these extra computations does not slow down the statistics calculation too much. After this change, the statistics calculation for a dataset with 2.8 million entries (100,000 rows, 28 columns) takes 0.685 seconds on average, compared to 0.665 seconds prior to the change (averaged over 100 trials).

# Tests
Changes were tested locally to confirm that the new statistics were calculated correctly.

cc:
@wangxiaoyou1993 @shrey-mage @nathaniel-mage


mage_ai/data_cleaner/statistics/calculator.py
```diff
@@ -63,14 +63,44 @@ def calculate_statistics_overview(self, df, is_clean=True):
 
             # Aggregated stats
             column_count = len(df.columns)
+            row_count = df.shape[0]
+
             data['total_null_value_count'] = sum(
                 data[f'{col}/null_value_count'] for col in df.columns
             )
+            data['total_null_value_rate'] = (
```
wangxiaoyou1993, line 71: let's create a method for calculating the rate. the logic of checking denominator is 0 is repeated.

wangxiaoyou1993 (approved):

PR #359, opened 2022-06-14
[nt] table invalid redirect

nathaniel-mage:
# Summary

### Clicking on the table cell of an invalid value will redirect to reports.
- The reports page has a table containing invalid values, accessed by the reports tab or clicking an invalid cell.
- In a future branch, we'll rework the reports page to be 'value': 'count' to increase readability/reduce duplicates.

# Tests

Tested on the Localhost app. I broke this off into two versions due to some UX decisions to get input.
### Version 1 (Remote branch)
View all is side by side with the cell, but the onClick is triggered when clicking on the cell
image

### Version 2 (Local branch)
View all is not side by side but the onClick is triggered when clicking on the text
image

Let me know which version is preferred and I can edit it accordingly.
cc: @dy46 @johnson-mage

### DIFF
--- a/mage_ai/frontend/components/DataTable/index.tsx
+++ b/mage_ai/frontend/components/DataTable/index.tsx
@@ -15,6 +15,7 @@ import {
 } from 'react-table';
 import { useSticky } from 'react-table-sticky';
 
+import FlexContainer from '@oracle/components/FlexContainer';
 import light from '@oracle/styles/themes/light';
 import Text from '@oracle/elements/Text';
 import {
@@ -25,8 +26,8 @@ import {
   REGULAR,
   REGULAR_LINE_HEIGHT,
 } from '@oracle/styles/fonts/sizes';
+import { TABS_QUERY_PARAM, SHOW_COLUMNS_QUERY_PARAM } from '@components/datasets/overview';
 import { UNIT } from '@oracle/styles/units/spacing';
-import FlexContainer from '@oracle/components/FlexContainer';
 
 const BASE_ROW_HEIGHT = (UNIT * 2) + REGULAR_LINE_HEIGHT;
 const DEFAULT_COLUMN_WIDTH = UNIT * 20;
@@ -239,7 +240,7 @@ function Table({
             cellStyle.width = maxWidthOfFirstColumn;
           }
 
-          let cellValue = original[idx - 1];
+          const cellValue = original[idx - 1];
           if (isInvalid) {
             cellStyle.color = light.interactive.dangerBorder;
           }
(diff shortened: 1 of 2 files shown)


nathaniel-mage:
Updated look to match Figma and design discussion we had.
image

mage_ai/frontend/components/DataTable/index.tsx
```diff
@@ -249,23 +253,37 @@ function Table({
...
-                  {cellValue === true && 'true'}
-                  {cellValue === false && 'false'}
-                  {(cellValue === null || cellValue === 'null') && 'null'}
-                  {cellValue !== true
+                <FlexContainer justifyContent={'space-between'}>
+                  <Text danger={isInvalid} wordBreak>
+                    {cellValue === true && 'true'}
+                    {cellValue === false && 'false'}
+                    {(cellValue === null || cellValue === 'null') && 'null'}
+                    {cellValue !== true
```
johnson-mage, line 261: fix the indenting below

mage_ai/frontend/components/DataTable/index.tsx
```diff
@@ -249,23 +253,37 @@ function Table({
...
                     && cellValue !== false
                     && cellValue !== null
                     && cellValue !== 'null'
                     && cellValue
                   }
-                </>
+                  </Text>
+                  {isInvalid && (
+                    <NextLink
+                      as={`/datasets/${slug}/?tabs[]=Reports&show_columns=1&column=${index}`}
```
johnson-mage, line 270: There are `TABS_QUERY_PARAM` and `SHOW_COLUMNS_QUERY_PARAM` constants from the `mage-ai/mage_ai/frontend/components/datasets/overview/index.tsx` file you can use here. No `COLUMN_QUERY_PARAM` but you can add it. May be helpful if we end up changing the query params down the road, will only need to change in one place.

mage_ai/frontend/components/DataTable/index.tsx
```diff
@@ -249,23 +253,37 @@ function Table({
             >
               {firstColumn && cell.render('Cell')}
               {!firstColumn && (
-                <>
-                  {cellValue === true && 'true'}
-                  {cellValue === false && 'false'}
-                  {(cellValue === null || cellValue === 'null') && 'null'}
-                  {cellValue !== true
+                <FlexContainer justifyContent={'space-between'}>
```
johnson-mage, line 256: string props dont need brackets, can use double quote

mage_ai/frontend/components/DataTable/index.tsx
```diff
@@ -249,23 +253,37 @@ function Table({
...
+                    </NextLink>
+                  )}
+                </FlexContainer>
               )}
             </div>
           );
         })}
       </div>
     );
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
johnson-mage, line 286: @nathaniel-mage Are we sure we shud just ignore the warning instead of updating the dependency array?

nathaniel-mage: I was worried about rerender issues, but I'll keep in all the dependencies.

johnson-mage (approved):

PR #365, opened 2022-06-14
[dy] Launch with API key

dy46:
# Summary

Launch the data cleaning tool with a group API key to sync pipelines with the Mage database. Calling `mage_ai.launch` with an api_key will kick off the `sync_pipelines()` method which will sync the pipelines between the user's local storage and Mage's database. Right now, we will automatically upload all of the local pipelines to Mage associated to the api key's group.

Usage:
```
import mage_ai

df = pd.read_csv('./testfiles/life_expectancy.csv')

mage_ai.launch(api_key="ErLVkWVZMJVSVDHDpembT7sPpom6cl0Vl8jR1dlg")
mage_ai.clean(df, remote_pipeline_uuid=1)
```

# Tests

tested locally

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/__init__.py
+++ b/mage_ai/__init__.py
@@ -60,11 +60,13 @@ def connect_data(df, name):
     return feature_set
 
 
-def clean(df, pipeline_uuid=None, pipeline_path=None):
+def clean(df, pipeline_uuid=None, pipeline_path=None, remote_pipeline_uuid=None):
     if pipeline_uuid is not None:
         df_clean = clean_df_with_pipeline(df, id=pipeline_uuid)
     elif pipeline_path is not None:
         df_clean = clean_df_with_pipeline(df, path=pipeline_path)
+    elif remote_pipeline_uuid is not None:
+        df_clean = clean_df_with_pipeline(df, remote_id=remote_pipeline_uuid)
     else:
         _, df_clean = clean_df(df)
     return df_clean
--- a/mage_ai/__init__.py
+++ b/mage_ai/__init__.py
@@ -25,6 +25,7 @@ def launch(
             host=iframe_host,
             port=iframe_port,
         )
+
     return thread
 
 
--- a/mage_ai/__init__.py
+++ b/mage_ai/__init__.py
@@ -61,13 +61,19 @@ def connect_data(df, name):
     return feature_set
 
 
-def clean(df, pipeline_uuid=None, pipeline_path=None, remote_pipeline_uuid=None):
+def clean(
+    df,
+    pipeline_uuid=None,
+    pipeline_path=None,
+    remote_pipeline_uuid=None,
+    api_key=None,
+):
     if pipeline_uuid is not None:
         df_clean = clean_df_with_pipeline(df, id=pipeline_uuid)
     elif pipeline_path is not None:
         df_clean = clean_df_with_pipeline(df, path=pipeline_path)
     elif remote_pipeline_uuid is not None:
-        df_clean = clean_df_with_pipeline(df, remote_id=remote_pipeline_uuid)
+        df_clean = clean_df_with_pipeline(df, remote_id=remote_pipeline_uuid, mage_api_key=api_key)
     else:
         _, df_clean = clean_df(df)
     return df_clean
--- a/mage_ai/__init__.py
+++ b/mage_ai/__init__.py
@@ -17,14 +17,14 @@ def launch(
     iframe_host=None,
     iframe_port=None,
     inline=True,
+    api_key=None,
 ):
-    thread = launch_flask()
+    thread = launch_flask(api_key)
     if inline:
         display_inline_iframe(
             host=iframe_host,
             port=iframe_port,
         )
-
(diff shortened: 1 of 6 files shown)


mage_ai/server/app.py
```diff
@@ -377,11 +388,18 @@ def launch() -> None:
         'host': host,
         'port': port,
     }
+    app_kwargs = {'port': SERVER_PORT, 'host': 'localhost', 'debug': False}
```
wangxiaoyou1993, line 391: duplicate app_kwargs

mage_ai/server/client/mage.py
```diff
@@ -0,0 +1,56 @@
+import json
+import requests
+from mage_ai.data_cleaner.shared.hash import merge_dict
```
wangxiaoyou1993, line 3: nit: put `from xxx import yyy` above `import xxx`

mage_ai/__init__.py
```diff
@@ -60,11 +61,13 @@ def connect_data(df, name):
     return feature_set
 
 
-def clean(df, pipeline_uuid=None, pipeline_path=None):
+def clean(df, pipeline_uuid=None, pipeline_path=None, remote_pipeline_uuid=None):
```
wangxiaoyou1993, line 64: when calling clean method with remote_pipeline_uuid, we need to pass in api_key as well. it'll be used independently in production environment without launching server.

mage_ai/server/app.py
```diff
@@ -93,7 +94,7 @@ def process():
     if not id:
         return
 
-    feature_set = FeatureSet(id=id)
+    feature_set = FeatureSet(id=id, api_key=api_key)
```
wangxiaoyou1993, line 97: Do we use the process endpoint anywhere?

mage_ai/server/client/mage.py
```diff
@@ -0,0 +1,56 @@
...
+                        'X-API-KEY': api_key,
+                    },
+                    url=f'{self.url_prefix}/data_cleaning_pipelines',
+                ).json()
+                pipeline_response = response['data_cleaning_pipeline']
+                pipeline.metadata = merge_dict(pipeline.metadata, {
+                    'remote_id': pipeline_response['id'],
+                })
+        except:
+            pass
```
wangxiaoyou1993, line 43: Let's print out a sync failure message. Same below.

mage_ai/server/client/mage.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def get_pipeline_actions(self, id, api_key):
```
wangxiaoyou1993, line 45: Same. Check whether api_key is None here.

mage_ai/__init__.py
dy46: I see

mage_ai/server/app.py
dy46: i'm not sure, but i can remove it in a future PR if we don't use this endpoint

dy46:
@wangxiaoyou1993 updated

mage_ai/server/app.py
```diff
@@ -314,16 +319,19 @@ def clean_df(df, name):
...
+def clean_df_with_pipeline(df, id=None, path=None, remote_id=None, mage_api_key=None):
     pipeline = None
     if id is not None:
-        pipeline = Pipeline(id=id)
+        pipeline = Pipeline(id=id).pipeline
     elif path is not None:
-        pipeline = Pipeline(path=path)
+        pipeline = Pipeline(path=path).pipeline
+    elif remote_id is not None:
+        final_key = mage_api_key if mage_api_key is not None else api_key
```
wangxiaoyou1993, line 329: api_key could be undefined.

mage_ai/server/data/models.py
```diff
@@ -226,6 +227,12 @@ def __init__(self, id=None, path=None, feature_set_id=None, pipeline=None):
         if pipeline is not None:
             self.pipeline = pipeline
 
+        if api_key is not None:
```
wangxiaoyou1993, line 230: Is this change needed?

dy46: If the api_key is None then I think it makes sense to not try to sync pipelines

dy46: You can initialize the `Pipeline` model class with a `pipeline`, so I guess I was thinking in that case, we would want to sync it. However, I don't think we actually ever initialize a `Pipeline` like that, so we could remove that if it's not necessary.

mage_ai/server/app.py
dy46: @wangxiaoyou1993 good call. I updated to set the `api_key` global variable.

mage_ai/server/data/models.py
wangxiaoyou1993: ok, that makes sense. maybe change it to `pipeline is not None and api_key is not None`

wangxiaoyou1993 (approved):

PR #379, opened 2022-06-16
[sp] action code editor fixes

shrey-mage:
# Summary
- fix backend error when user applies code example
- replace tabs with spaces
- add optional chaining to displayed action arguments

# Tests
- Click "New Action" > make no changes, click "Apply". Backend no longer returns an error, and code is applied.
- Write a line with four manual spaces. Write a line with a tab. Backend no longer returns an error, and code is applied.

cc: @johnson-mage @dy46 @tommydangerous @wangxiaoyou1993

### DIFF
--- a/mage_ai/frontend/components/ActionForm/index.tsx
+++ b/mage_ai/frontend/components/ActionForm/index.tsx
@@ -200,7 +200,7 @@ function ActionForm({
                 minLines={MIN_LINES_ACTIONS}
                 mode="python"
                 onChange={e => {
-                  payload['action_code'] = e;
+                  payload['action_code'] = e.replace('\t', ' '.repeat(4));
                   setActionCodeState(payload.action_code ?? actionCode);
                   setCustomCodeState({
                     actionType,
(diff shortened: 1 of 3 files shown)


mage_ai/frontend/components/ActionForm/index.tsx
```diff
@@ -199,7 +200,7 @@ function ActionForm({
                 minLines={MIN_LINES_ACTIONS}
                 mode="python"
                 onChange={e => {
-                  payload['action_code'] = e;
+                  payload['action_code'] = e.replace('\t', ' '.repeat(4));
```
shrey-mage, line 203: Concerns about this approach:
- Are there cases where a tab is actually preferred in Python source code over 4 spaces? I wouldn't think so, since code editors anyways insert 4 spaces in place of a tab.
- Is there any performance hit for doing this in `onChange`? Don't think so either.

johnson-mage: Seems fine

johnson-mage (approved):

tommydangerous:
This should be done on the backend.

PR #380, opened 2022-06-16
[sk] Added endpoint to download feature set

skunichetty:
# Summary
Added endpoint to download feature set CSVs from:
`http://localhost:xxxx/feature_sets/<id>/download`

This URL has one optional argument: `store_index`. If set to `true`, the downloaded dataset will include the previous dataset index (which may be useful to keep in the case that non-integer indices like timestamps are used). If unspecified, `store_index` defaults to `false` and index is not included. Example usage: `http://localhost:xxxx/feature_sets/<id>/download?store_index=true`

This endpoint checks to make sure the feature set ID passed is valid, and raises an error if not. If the feature set has name `name with spaces`, the feature set (with recent edits applied) is downloaded as `name_with_spaces.csv` (to make access via command line easier). This endpoint pulls the latest edits, so after cleaning suggestions are applied, the downloaded dataset will reflect the changes made.

# Tests
Changes were tested locally by applying transformations to stored feature sets and then downloading the edited feature sets using this new endpoint. Additional parameter was also tested.

cc:
@wangxiaoyou1993 @johnson-mage


wangxiaoyou1993:
Could you check whether we can use the `send_file` or `send_from_directory` method from flask library? I haven't tried them. Could you check there're any benefits of using those methods?

wangxiaoyou1993:
Also, could you test the endpoint with a feature set with 100k rows?

skunichetty:
I tested on a feature set with 100k rows, it finishes the download within 960 ms of sending the request. This could be problematic as currently the entire dataset is loaded into memory in order to write to file. What might be faster is to write both a CSV and Parquet file every time cleaning actions are applied, and then return the CSV from filesystem instead. This is where `send_file` and `send_from_directory` are useful, as they are designed to work with files that already exist on the filesystem.

wangxiaoyou1993:
> I tested on a feature set with 100k rows, it finishes the download within 960 ms of sending the request. This could be problematic as currently the entire dataset is loaded into memory in order to write to file. What might be faster is to write both a CSV and Parquet file every time cleaning actions are applied, and then return the CSV from filesystem instead. This is where `send_file` and `send_from_directory` are useful, as they are designed to work with files that already exist on the filesystem.

ok, 960ms is good. we load the entier dataset into memory when applying cleaning action anyway. no need to write both csv and parquet.

wangxiaoyou1993 (approved):

PR #383, opened 2022-06-16
[jk] Add  syntax error action

johnson-mage:
# Summary
- See title

# Tests
image

Title not included in PUT request payload for saving action:

Title IS included in PUT request payload:


johnson-mage:
@skunichetty Should the `title` property always be included in the pipelines PUT request payload so that the title appears in the pipeline actions response? We previously did not include it for non-suggested actions, but I noticed that even without including the title for certain actions, the backend would still return the title in the response.

For example in the gif below, "fix syntax errors" has a blank title, but "impute" does not:

PR #395, opened 2022-06-17
[sp] add text column stats

shrey-mage:
# Summary
- add word distribution chart to column-specific "Visualizations" tab
- display min/mean/max word counts in column-specific "Reports" tab
- display min/mean/max word counts in feature profiles

# Tests
image
image
image

cc: @johnson-mage @wangxiaoyou1993 @dy46


wangxiaoyou1993:
Can we also min/max/mean string length in the statistics?

johnson-mage:
wondering if there's a way to indicate that "min/max/mean" for a text column is word count in the feature profiles table. May not be clear to a user what "min" for a text col is.

shrey-mage:
> wondering if there's a way to indicate that "min/max/mean" for a text column is word count in the feature profiles table. May not be clear to a user what "min" for a text col is.

agreed, was thinking about this. Maybe we can put a label on hover? Maybe Kai's tooltip explanation design?

wangxiaoyou1993:
Or we should just use different stats name in features profiles. Something like this
image

mage_ai/frontend/components/datasets/columns/ColumnReports.tsx
```diff
@@ -101,35 +107,69 @@ function ColumnReports({
...
+        value: isNumeric(median) ? roundNumber(median) : median,
+      },
+      {
+        name: 'Average value',
+        value: isNumeric(average) ? roundNumber(average) : average,
+      },
+    );
+  }
+
+  if (columnType === ColumnTypeEnum.TEXT) {
```
johnson-mage, line 144: do we also need to check for other string column types like CATEGORY?

mage_ai/frontend/components/datasets/overview/index.tsx
```diff
@@ -317,7 +318,13 @@ function DatasetOverview({
               column={selectedColumn}
               features={features}
               insights={insightsOverview}
-              statisticsByColumn={statistics?.[`${selectedColumn}/value_counts`] || {}}
+              statisticsByColumn={
+                (colType === ColumnTypeEnum.TEXT 
+                  ?
+                  statistics?.[`${selectedColumn}/word_distribution`]
```
johnson-mage, line 324: can put on line above

mage_ai/frontend/components/datasets/overview/index.tsx
```diff
@@ -317,7 +318,13 @@ function DatasetOverview({
...
+                  :
+                  statistics?.[`${selectedColumn}/value_counts`]
```
johnson-mage, line 326: this too

johnson-mage (approved):

PR #419, opened 2022-06-21
[sk] Added support for list column type

skunichetty:
# Summary
Added support for list type in dataframe, including automatic type inference both from a list object and string representation of the list. This inference also supports loading tuples.

This ended being less trivial than first imagined, namely as:
- Pandas serializes lists in string format when saving a dataframe to csv
- The native Python `list` type is mutable, and thus not hashable by Pandas (making computing value counts and unique values more difficult)

To resolve these challenges, a few changes were added throughout the `data_cleaner` package:
- Instead of `list`, the `FrozenList` object is used from `pandas.core.indexes.frozen`, which acts as an immutable list (albeit not totally immutable). This allows a list-like object to be used as an index in a Pandas dataframe, while still allowing for a list-style string representation when pretty-printed (is pretty-printed with the format `"[...]"`)
- Automatic parsing of serialized lists was added, but by constraint can only work with python primitives and `np.nan` as of now. Example: `"['not string?', True, True, 8, False, np.nan, None]"` will get parsed to `['not string?', True, True, 8, False, np.nan, None]`
- Upon calling the pipeline update endpoint, the original dataframe loaded from file will have the list in string form. This is a problem, for example, when using custom code to write a filter on list only properties (user expects a list, but the data is actually a string). To resolve this, the dataframe is cleaned upon being loaded from file, before the pipeline actions are executed.

Of note, I placed a TODO marking that a new `ImputeValuesSubRule` should be created for list types, I'm not including it in this PR since this one is already pretty long. I added the datatype to `StringImputeSubRule` for now to avoid errors.

# Tests
Unit tests were written to check the correct column type inference behavior, alongside running on test data and executing cleaning actions to confirm that the frontend works correctly with list data.

cc:
@wangxiaoyou1993


mage_ai/server/app.py
```diff
@@ -302,7 +303,10 @@ def update_pipeline(id):
     feature_set_id = pipeline.metadata.get('feature_set_id')
     if feature_set_id is not None:
         feature_set = FeatureSet(id=feature_set_id)
-        df_transformed = clean_pipeline.transform(feature_set.data_orig, auto=False)
+        df = feature_set.data_orig
+        ctypes = feature_set.metadata['column_types']
```
wangxiaoyou1993, line 307: also use the ctypes in the clean_data method.

mage_ai/data_cleaner/shared/utils.py
```diff
@@ -59,6 +63,30 @@ def is_spark_dataframe(df):
...
+    if dtype is FrozenList:
+        return list_literal
+    elif dtype in (list, tuple):
+        return FrozenList(list_literal)
+    elif dtype is not str:
+        return None
+    list_literal = list_literal.strip('[]() ')
+    if list_literal == '':
+        return FrozenList([])
+    return FrozenList([__parse_element(element) for element in list_literal.split(', ')])
```
wangxiaoyou1993, line 87: what if there's no space between elements, like '1,2,3,4'?

mage_ai/tests/data_cleaner/statistics/test_calculator.py
```diff
@@ -1,4 +1,5 @@
 from faker import Faker
+from mage_ai.data_cleaner.column_types.column_type_detector import infer_column_types
```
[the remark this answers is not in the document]
skunichetty: Removed.

mage_ai/data_cleaner/shared/utils.py
skunichetty: Expanded the splitting algorithm to now count whitespace as well.

mage_ai/data_cleaner/column_types/column_type_detector.py
```diff
@@ -135,78 +137,76 @@ def infer_column_type(series, column_name, dtype, kwargs):
...
-                                mdtype = ColumnType.CATEGORY
-                            else:
-                                mdtype = ColumnType.CATEGORY_HIGH_CARDINALITY
-    return mdtype
+                return ColumnType.NUMBER
+    else:
+        matches = clean_series.str.match(REGEX_DATETIME).sum()
+        if matches / length >= DATETIME_MATCHES_THRESHOLD:
+            return ColumnType.DATETIME
+        correct_emails = clean_series.str.match(REGEX_EMAIL).sum()
```
wangxiaoyou1993, line 179: Let's add a TODO to refactor some of the code here. There's some duplicate logic and we don't need to do regex matching for all the syntax for each column.

mage_ai/data_cleaner/shared/utils.py
```diff
@@ -59,6 +65,30 @@ def is_spark_dataframe(df):
...
+            return eval(element)
+        except:
+            return None
+
+
+def parse_list(list_literal: Union[str, List[Any]]) -> FrozenList:
+    dtype = type(list_literal)
+    if dtype is FrozenList:
+        return list_literal
+    elif dtype in (list, tuple):
```
wangxiaoyou1993, line 82: Let's also add set here.

mage_ai/data_cleaner/shared/utils.py
```diff
@@ -59,6 +65,30 @@ def is_spark_dataframe(df):
...
+        return FrozenList(list_literal)
+    elif dtype is not str:
+        return None
```
wangxiaoyou1993, line 85: if it's not str, we can return FrozenList([list_literal])

wangxiaoyou1993 (approved):

mage_ai/data_cleaner/shared/utils.py
skunichetty: Added

mage_ai/data_cleaner/shared/utils.py
skunichetty: Added

mage_ai/data_cleaner/column_types/column_type_detector.py
skunichetty: Added TODO

mage_ai/server/app.py
```diff
@@ -302,10 +303,13 @@ def update_pipeline(id):
     feature_set_id = pipeline.metadata.get('feature_set_id')
     if feature_set_id is not None:
         feature_set = FeatureSet(id=feature_set_id)
-        df_transformed = clean_pipeline.transform(feature_set.data_orig, auto=False)
+        df = feature_set.data_orig
+        ctypes = feature_set.metadata.get('column_types', {})
+        df = clean_dataframe(df, ctypes, False)
```
wangxiaoyou1993, line 308: @skunichetty This throws error when deleting a "remove_column" action. The column_types doesn't include the deleted column name.

wangxiaoyou1993: Cool. Let's push the fix first.

PR #422, opened 2022-06-21
[sp] add column icons to feature profiles

shrey-mage:
# Summary
- see above

# Tests

Doesn't break max width text styling, so long column names still display correctly.

image

cc: @johnson-mage @dy46


johnson-mage:
@shrey-mage  can you add a `title` prop to the column type icon svgs so if you hover over it, it shows the text for column type? On the feature profiles table and also the main data table.

johnson-mage (approved):
approved after addressing comment

dy46 (commented):
Not sure what the designs look like, but the spacing looks a little cramped. Maybe add a little bit more padding between the icon and the text?

PR #429, opened 2022-06-22
[dy] Add instructions for api key

dy46:
# Summary

Add instructions for using API key in the export pipeline page.

without API key:
Screen Shot 2022-06-22 at 2 48 12 PM

with API key:
Screen Shot 2022-06-22 at 2 47 56 PM

# Tests

local

cc: @wangxiaoyou1993 @johnson-mage


wangxiaoyou1993:
I have 2 questions
* When user lands on the instruction page, the server is already launched. Calling mage_ai.launch() again will cause error.
* If the server is already launched with the API key, do we show instructions differently?
* If the pipeline is not synced to the cloud, what do we show in the remove_pipeline_uuid?

wangxiaoyou1993:
In the instructions, `mage_ai.remote_sync()` and `mage_ai.clean` will be used in two places. So maybe put them into two code blocks to avoid confusion. Also, the text is still showing "mage_ai.launch" command.

PR #435, opened 2022-06-23
[sp] add visualization for email data

shrey-mage:
# Summary
- add visualization for email data
- refactor column-specific logic to use lookup maps

# Tests
image

cc: @johnson-mage @dy46


johnson-mage:
add screenshot with tooltip

johnson-mage:
does this appear in the header column of the data table as well? we shud add it there, too.

johnson-mage (approved):

PR #447, opened 2022-06-27
[sk] Added SQL-like syntax for filtering

skunichetty:
# Summary
Filtering system now supports SQL-like syntax to improve ease of access for data scientists and machine learning engineers.

The only query currently supported is `SELECT` with the format:
```sql
SELECT [*|{column list}] FROM df [WHERE condition]
```
- `WHERE` condition supports using the `LIKE`, `BETWEEN`, and `IN` operation
    - `IN` currently only supports comparisons to sets of values: `column IN (1,2,3)`
    - `BETWEEN` supports numbers, strings, and datetimes. To perform datetime comparisons, wrap the iso-formatted string with the `date` function: `"Date of Birth" BETWEEN date("1970-1-1") AND date("2008-4-3")` (this generally applies to all comparisons)
    - `LIKE` supports the SQL-style wildcard matching
- If a column name has whitespace, newlines, or any of the following special characters: `+=-*&^%$! ?~|<>(){}[],.`, then wrap the column name in quotes (double quotes) to use it in the query. If any other character is used, the column name cannot be used in the query.
- Comparisons between columns can be performed: `column1 < column2`

# Tests
Three types of tests were performed:
- New unit tests were added to perform the conversion from SQL query to pandas filtering query
- Old unit tests were updated to check the successful filtering of sample datasets
- Local testing was performed with sample data to test integration with the frontend

cc:
@wangxiaoyou1993


tommydangerous:
@wangxiaoyou1993 is this still needed?

PR #460, opened 2022-06-29
[dy] Add execution logic for pipelines

dy46:
# Summary

Add execute logic for pipelines. Right now, the execution logic will run the blocks in parallel when possible.

# Tests

unit tests

cc:

@wangxiaoyou1993

### DIFF
--- a/mage_ai/data_preparation/models/pipeline.py
+++ b/mage_ai/data_preparation/models/pipeline.py
@@ -1,7 +1,7 @@
-from queue import Queue
 from mage_ai.data_cleaner.shared.utils import clean_name
 from mage_ai.data_preparation.models.block import Block
 from mage_ai.data_preparation.templates.utils import copy_templates
+from queue import Queue
 import asyncio
 import os
 import yaml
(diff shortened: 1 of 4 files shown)


mage_ai/data_preparation/models/pipeline.py
```diff
@@ -107,6 +132,7 @@ def add_block(self, block, upstream_block_uuids=[]):
         block.pipeline = self
         self.blocks_by_uuid[block.uuid] = block
         self.__save()
+        self.load_config_from_yaml()
```
wangxiaoyou1993, line 135: why do we need to reload the config?

mage_ai/data_preparation/models/pipeline.py
```diff
@@ -58,9 +60,32 @@ def get_all_pipelines(self, repo_path):
...
         return os.listdir(os.path.join(repo_path, PIPELINES_FOLDER))
 
-    def execute(self):
-        # TODO: implement execution logic
-        pass
+    # async function for parallel processing
+    async def execute(self):
+        tasks = dict()
+        blocks = Queue()
+        for b in self.block_configs:
```
wangxiaoyou1993, line 67: I guess you reload the config because self.block_configs is not updated. you can use self.blocks_by_uuid.values() to get updated blocks.

wangxiaoyou1993:
Could you rebase master?

wangxiaoyou1993 (approved):

PR #466, opened 2022-06-29
[dy] Add command line

dy46:
# Summary

Add command line commands for initializing repo and starting server.

Currently, the command is set as `mage-ai <command>` instead of `mage <command>` because that keyword seems to be used somewhere already.

# Tests

tested locally

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/command_line.py
+++ b/mage_ai/command_line.py
@@ -1,0 +1,20 @@
+from mage_ai.data_preparation.repo_manager import init_repo
+from mage_ai.server.server import main as start_server
+
+import asyncio
+import os
+
+def main():
+    import sys
+    command = sys.argv[1]
+
+    if command == 'init':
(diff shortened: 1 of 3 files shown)


mage_ai/data_preparation/repo_manager.py
```diff
@@ -1,12 +1,24 @@
...
 
-
 def init_repo(repo_path: str) -> None:
     """
     Initialize a repository under the current path.
     """
+
     if os.path.exists(repo_path):
         return
+    else:
```
wangxiaoyou1993, line 11: You might not need to add these since I already added the code to copy template files. You can add additional files to this folder if needed: https://github.com/mage-ai/mage-ai/tree/master/mage_ai/data_preparation/templates/repo

dy46: oh i see

wangxiaoyou1993 (approved):

PR #469, opened 2022-06-30
[sk] Added data loaders + templates for S3 and Redshift

skunichetty:
# Summary
Added data loaders and templates for loading data from either an S3 bucket or a Redshift cluster. These data loaders are more complex since AWS offers multiple different authentication and authorization methods. I've designed support for the more common methods.

To perform these connections, two new libraries were added:
- `boto3` is used to access S3 buckets
- `redshift-connector` is used to access Redshift clusters

Note: The `redshift-connector` library (used to connect to a Redshift cluster) requires an older version of the `requests` library, so I downgraded it from version 2.28 to version 2.27. No other conflicts have emerged so far, but I can revert this change if need be.

# Tests
Tests were performed by testing loading and querying data from S3 buckets and Redshift clusters.

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_ai/data_loader/s3.py
+++ b/mage_ai/data_loader/s3.py
@@ -14,6 +14,41 @@ class S3(BaseFile):
     - ".hdf5"
     """
 
+    def __init__(
+        self,
+        bucket_name: str,
+        object_name: str,
+        format: FileFormat = None,
+        **kwargs,
+    ) -> None:
+        """
+        Initializes data loader from an S3 bucket. If credentials are stored on
+        file, no further arguments are needed. Otherwise, use the factory method to construct
+        the data loader using manually specified credentials.
+
+        Args:
+            bucket_name (str): Bucket to load resource from
+            object_name (str): Object key name within bucket given
+            format (FileFormat, optional): File format of object. Defaults to None, in which
+            case the format is inferred.
+            **kwargs: all other keyword arguments to pass to the client
+        """
+        super().__init__(object_name, format)
+        self.bucket_name = bucket_name
+        self.client = boto3.client('s3', **kwargs)
+
+    def load(self, **kwargs) -> DataFrame:
+        """
+        Loads data from S3 into a Pandas data frame.
+
+        Returns:
+            DataFrame: The data frame specified by the bucket name and object name
+        """
+        buffer = BytesIO()
+        self.client.download_fileobj(self.bucket_name, self.filepath, buffer)
+        buffer.seek(0)
+        return self.reader(buffer, **kwargs)
+
     @classmethod
     def with_credentials(
         cls,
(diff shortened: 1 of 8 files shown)


mage_ai/data_preparation/templates/data_loaders/redshift.py
```diff
@@ -0,0 +1,23 @@
...
+def load_data_from_redshift() -> DataFrame:
+    """
+    Template code for loading data from Redshift cluster.
+
+    This template assumes that IAM profiles are specified in `~/.aws`.
+    - If IAM profiles are not setup, manually specify them as keyword arguments in `config`
+    - If using temporary database credentials, use `Redshift.with_temporary_credentials()`
+    to create loader.
+    """
+    config = {
```
wangxiaoyou1993, line 14: we also need to provide host and port, right? what configs are stored in IAM profiles?

mage_ai/data_loader/s3.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def load(self, **kwargs) -> DataFrame:
+        """
+        Loads data from S3 into a Pandas data frame.
+
+        Returns:
+            DataFrame: The data frame specified by the bucket name and object name
+        """
+        buffer = BytesIO()
+        self.client.download_fileobj(self.bucket_name, self.filepath, buffer)
```
wangxiaoyou1993, line 49: you can just use `get_object` method

mage_ai/data_preparation/templates/data_loaders/s3.py
```diff
@@ -0,0 +1,16 @@
...
+def load_from_s3_bucket() -> DataFrame:
+    """
+    Template code for loading data from S3 bucket.
+
+    This template assumes that user credentials are specified in `~/.aws`.
+    If not, use `S3.with_credentials()` to manually specify AWS credentials or use
+    AWS CLI to configure credentials on system.
+    """
+    bucket_name = 'your_s3_bucket_name'  # Specify S3 bucket name to pull data from
+    object_name = 'your_object_name'  # Specify object to download from S3 bucket
```
wangxiaoyou1993, line 14: You can call it object_key, which includes the path to the object.

mage_ai/data_loader/s3.py
skunichetty: Changed to use `get_object`

mage_ai/data_preparation/templates/data_loaders/s3.py
skunichetty: Refactored code to use `object_key` instead of `object_name` (which is strange since AWS officially calls the identifiers `object_key` but `boto3` instead uses `object_name` in some places).

mage_ai/data_preparation/templates/data_loaders/redshift.py
wangxiaoyou1993: I think the most common config is to provide the following configs: database, host, password, port, user

wangxiaoyou1993 (approved):

PR #472, opened 2022-06-30
[dy] Add other execution methods

dy46:
# Summary

Add pipeline execution from python method and command line 

# Tests

tested locally

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/__init__.py
+++ b/mage_ai/__init__.py
@@ -1,4 +1,5 @@
 from mage_ai.data_cleaner.shared.utils import is_spark_dataframe
+from mage_ai.data_preparation.models.pipeline import Pipeline
 from mage_ai.server.app import (
     clean_df,
     clean_df_with_pipeline,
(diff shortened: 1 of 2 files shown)


mage_ai/__init__.py
```diff
@@ -112,3 +115,11 @@ def clean(
     else:
         _, df_clean = clean_df(df, name=name, verbose=verbose)
     return df_clean
+
+# Data preparation methods
```
wangxiaoyou1993, line 119: Maybe make the comment more visible like
`# ------------------------- Data preparation methods ------------------------- #`

mage_ai/__init__.py
```diff
@@ -112,3 +115,11 @@ def clean(
...
+
+def run(pipeline_uuid: str, repo_name: str=None) -> None:
```
wangxiaoyou1993, line 121: nit: put space between and after `=`

wangxiaoyou1993 (approved):

PR #480, opened 2022-06-30
[dy] Add subprocess

dy46:
# Summary

Add subprocess to install dependencies before running pipeline. This could be a little annoying because it will run every time the pipeline runs, but I'm not sure how else to do it.

Also, I noticed that we will have a repo level `requirements.txt` file as well, but I don't see it being run anywhere. I'm not sure when we should install those dependencies either.

# Tests

local / unit tests

cc: @wangxiaoyou1993


wangxiaoyou1993:
Actually, let's use one requirements.txt for the whole repo instead of individual pipelines.

PR #494, opened 2022-07-02
[sk] Updated directory tree generation

skunichetty:
# Summary
Tree generation for frontend filesystem. Example payload: 
image

# Tests
Tests were performed locally to make sure that the correct tree structure is generated

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_ai/data_preparation/models/file.py
+++ b/mage_ai/data_preparation/models/file.py
@@ -1,4 +1,8 @@
+from typing import Dict
 import os
+
+MAX_DEPTH = 30
+INACCESSIBLE_DIRS = frozenset(['__pycache__'])
 
 
 class File:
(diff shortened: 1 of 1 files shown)


mage_ai/data_preparation/models/file.py
```diff
@@ -54,10 +43,45 @@ def delete(self):
...
+    disabled = name[0] == '.' or name in INACCESSIBLE_DIRS
+    children = []
+    entries = os.scandir(directory)
+    children.extend(
+        _traverse(entry, disabled, 1) for entry in sorted(entries, key=lambda entry: entry.name)
+    )
+    return dict(name=name, children=children)
+
+
+def _traverse(entry, disabled=True, depth=1) -> Dict:
```
wangxiaoyou1993, line 72: How are you using depth? Do you want to stop traverse when depth > MAX_DEPTH?

mage_ai/data_preparation/models/file.py
```diff
@@ -54,10 +43,45 @@ def delete(self):
...
+    Traverses directory tree and returns the underlying file structure
+
+    Args:
+        directory (str): Top-level directory to generate tree from
+
+    Returns:
+        Dict: Directory tree payload representation
+    """
+    name = os.path.splitext(os.path.basename(directory))[0]
+    disabled = name[0] == '.' or name in INACCESSIBLE_DIRS
```
wangxiaoyou1993, line 63: There are some duplicate code here.
You can define `_traverse` method as
```
def _traverse(name, path, disabled=True, depth=1) -> Dict:
    ....
```
Then call `_traverse(repo_name, repo_path, False, 0)`

skunichetty: Fixed, although function has a lot of parameters now

mage_ai/data_preparation/models/file.py
skunichetty: A check against max depth accidentally got deleted at some point, I added it back. I stop traversal when `depth > MAX_DEPTH`.

wangxiaoyou1993 (approved):

PR #508, opened 2022-07-05
[dy] Run blocks from websocket

dy46:
# Summary

Run custom code in blocks from messages received by the websocket

# Tests

tested locally

Screen Shot 2022-07-05 at 5 41 39 PM

cc: @wangxiaoyou1993


mage_ai/server/websocket.py
```diff
@@ -39,25 +46,63 @@ def on_message(self, raw_message):
...
+        error = msg_id_value['error']
+
+        output_dict = dict(uuid=uuid)
+        if error is not None:
+            output_dict['data'] = error
+            output_dict['msg_type'] = 'error'
+        elif len(output) > 0:
+            df = find(lambda val: type(val) == pd.DataFrame, output)
+            output_dict['data'] = [
+                df.columns.to_list(),
```
wangxiaoyou1993, line 95: is this format used by frontend now? otherwise, can we use this format for consistency? https://github.com/mage-ai/mage-ai/blob/master/mage_ai/data_preparation/models/block.py#L239-L242

mage_ai/server/websocket.py
```diff
@@ -39,25 +46,63 @@ def on_message(self, raw_message):
...
+                *df.values.tolist(),
+            ]
+            output_dict['msg_type'] = 'data_frame'
```
wangxiaoyou1993, line 98: what are some message types? can we define constants?

wangxiaoyou1993 (commented):
Could you add more testing details, like screenshot?

mage_ai/server/websocket.py
dy46: it's not used in the frontend right now, but tommy asked for it in this format. @tommydangerous is it ok if we change it to the above format for consistency?

mage_ai/server/websocket.py
dy46: yeah i'll add some constants. also i realized i used the wrong enum, so i'll fix that too

mage_ai/server/constants.py
```diff
@@ -2,3 +2,5 @@
 
 SERVER_HOST = os.getenv('HOST', 'localhost')
 SERVER_PORT = os.getenv('PORT', 5789)
+
+DATAFRAME_OUTPUT_SAMPLE_COUNT = 10
```
wangxiaoyou1993, line 6: We have multiple dataframe output sample count constants. Let's put them together. https://github.com/mage-ai/mage-ai/blob/master/mage_ai/data_preparation/models/constants.py#L16-L17

wangxiaoyou1993: Maybe, you can update the DATAFRAME_SAMPLE_COUNT_PREVIEW to 10.

mage_ai/server/websocket.py
```diff
@@ -40,25 +47,64 @@ def on_message(self, raw_message):
...
+        uuid = msg_id_value['block_uuid']
+        output = msg_id_value['block_output']
+        error = msg_id_value['error']
+
+        output_dict = dict(uuid=uuid)
+        if error is not None:
+            output_dict['traceback'] = error
+        elif len(output) > 0:
+            df = find(lambda val: type(val) == pd.DataFrame, output)
+            if msg_type == 'execute_input':
```
wangxiaoyou1993, line 95: Is there a enum for msg_type?

wangxiaoyou1993 (approved):

mage_ai/server/websocket.py
dy46: i don't think so, this msg_type is also set by the jupyter client which doesn't have an enum set

PR #532, opened 2022-07-06
[sp] highlight block when selected in file tree

shrey-mage:
# Summary
- outlines the selected block when clicked in file tree

# Tests

cc: @johnson-mage @tommydangerous

### DIFF
--- a/mage_ai/frontend/components/FileTree/index.tsx
+++ b/mage_ai/frontend/components/FileTree/index.tsx
@@ -10,9 +10,12 @@ import { ArrowDown, ArrowRight, FileFill, Folder } from '@oracle/icons';
 import { FileNodeType, getFileNodeColor, ReservedFolderEnum } from './constants';
 import { UNIT } from '@oracle/styles/units/spacing';
 import { equals } from '@utils/array';
+import { findBlockByPath } from './utils';
+import { useBlockContext } from '@context/Block';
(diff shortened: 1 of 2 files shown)


mage_ai/frontend/components/FileTree/index.tsx
```diff
@@ -101,12 +108,13 @@ function FileTree({
...
     const blockPath = path.slice(1).join('/');
     const blockEl = blockRefs.current[blockPath];
-    blockEl?.current?.scrollIntoView({ behavior: 'smooth' });
+    blockEl?.current.scrollIntoView({ behavior: 'smooth' });
   };
 
   const fileTreeHandler = (path, isFolder) => (e) => {
     e.preventDefault();
     scrollToBlock(path);
+    setSelectedBlock(findBlockByPath(blocks, path));
```
johnson-mage, line 117: not sure if there'd ever be an undefined type error here, could possibly use optional chaining

johnson-mage (approved):

mage_ai/frontend/components/FileTree/index.tsx
shrey-mage: I had this issue earlier, it doesn't cause an error. it just makes it so that no block is selected; since this case should never arise and doesn't crash the frontend, I think it's fine (the alternative is not selecting any block anyways, so I don't think it's an issue)

PR #534, opened 2022-07-06
[sk] Added IO config handler

skunichetty:
# Summary
This code enables the user to define data IO configuration settings via a configuration file. In every repo created with `mage 
init`, a new file named `io_config.yaml` will be created, in which all IO configurations can be stored (alternatively from manually specifying the arguments in code). This is more useful, for example, when storing secret keys and access tokens where the secrets can be handled separately from code execution.

Here is a sample config file (contains only the default profile with credentials only for AWS data sources):
```yaml
default:
  AWS:
    Redshift:
      database: your_redshift_database_name
      # Use the settings below if you are using temporary credentials
      user: your_authentication_user_name
      password: your_authentication_password
      host: your_redshift_cluster_hostname
      port: your_redshift_cluster_port
      # Use the settings below if you are using IAM credentials
      cluster_identifier: your_redshift_identifier
      db_user: your_redshift_username
      profile: your_aws_credentials_profile
      iam: True
    S3:
      bucket_name: your_bucket_name
      object_key: your_object_key
    # Omit these settings if aws credentials are already configured on file
    aws_access_key_id: your_aws_access_key_id
    aws_secret_access_key: your_secret_access_key_id
    region_name: your_aws_region
```

To use this configuration file, use the following example code below:
```python
with BaseLoader.with_config(IOConfig(CONFIG_PATH).use('default')) as loader:
     pass
```
which initializes the data loader with the credentials stored in the `default` profile in `io_config.yaml`. The path to the config file should be specified as input to the constructor of the `IOConfig` object. The `.use()` function allows the user to specify the profile they wish to use.

In the example above, the 'AWS' and 'Redshift' objects are read to initialize the `Redshift` data loader. Specifically, the shared credentials from 'AWS' object and the specific 'Redshift' credentials are combined as inputs to the `Redshift` data loader. The implementation for how parameters are read from the configuration file are data loader dependent.

# Tests
Testing was performed to make sure that all data loaders still worked while loading this configuration file.

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_ai/data_preparation/templates/data_exporters/bigquery.py
+++ b/mage_ai/data_preparation/templates/data_exporters/bigquery.py
@@ -33,7 +33,7 @@ def export_data_to_big_query(df: DataFrame) -> None:
     Alternatively, all parameters can be specified in the configuration file.
     """
     table_id = 'your-project.your_dataset.your_table_name'
-    config_path = 'path/to/your/io/config/file.yaml'
+    config_path = './default_repo/io_config.yaml'
     config_profile = 'default'
 
     BigQuery.with_config(IOConfig(config_path).use(config_profile)).export(
--- a/mage_ai/data_preparation/templates/data_exporters/bigquery.py
+++ b/mage_ai/data_preparation/templates/data_exporters/bigquery.py
@@ -1,4 +1,5 @@
 from mage_ai.io.bigquery import BigQuery
+from mage_ai.io.io_config import IOConfig
 from pandas import DataFrame
 
 if 'data_exporter' not in globals():
@@ -28,13 +29,14 @@ def export_data_to_big_query(df: DataFrame) -> None:
     ```
     BigQuery.with_credentials_object({'service_key': ...}, **kwargs)
     ```
+
+    Alternatively, all parameters can be specified in the configuration file.
     """
     table_id = 'your-project.your_dataset.your_table_name'
-    config = {
-        # Specify any other configuration settings here to pass to BigQuery client
-        'project': 'your_project_name',
-    }
-    return BigQuery(**config).export(
+    config_path = 'path/to/your/io/config/file.yaml'
+    config_profile = 'default'
+
+    BigQuery.with_config(IOConfig(config_path).use(config_profile)).export(
         df,
         table_id,
         if_exists='replace',  # Specify resolution policy if table name already exists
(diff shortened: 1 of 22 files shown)


skunichetty:
Don't review just yet, I need to rebase with master to update with data exporter changes.

mage_ai/data_preparation/templates/data_exporters/bigquery.py
```diff
@@ -25,13 +26,14 @@ def export_data_to_big_query(df: DataFrame) -> None:
...
+
+    Alternatively, all parameters can be specified in the configuration file.
     """
     table_id = 'your-project.your_dataset.your_table_name'
-    config = {
-        # Specify any other configuration settings here to pass to BigQuery client
-        'project': 'your_project_name',
-    }
-    return BigQuery(**config).export(
+    config_path = 'path/to/your/io/config/file.yaml'
```
wangxiaoyou1993, line 33: Let's set a default path here since we auto generated the io_config.yaml

skunichetty: Updated default path to use the autogenerated config path in templates.

mage_ai/data_preparation/templates/data_exporters/bigquery.py
```diff
@@ -25,13 +26,14 @@ def export_data_to_big_query(df: DataFrame) -> None:
...
+
+    Alternatively, all parameters can be specified in the configuration file.
     """
     table_id = 'your-project.your_dataset.your_table_name'
-    config = {
-        # Specify any other configuration settings here to pass to BigQuery client
-        'project': 'your_project_name',
-    }
-    return BigQuery(**config).export(
+    config_path = './default_repo/io_config.yaml'
```
wangxiaoyou1993, line 33: let's use `get_repo_path` from `mage_ai.data_preparation.repo_manager` to get the correct repo path.

mage_ai/io/bigquery.py
```diff
@@ -99,7 +99,11 @@ def export(
         self.client.load_table_from_dataframe(df, table_id, job_config=config).result()
 
     @classmethod
-    def with_credentials_file(cls, path_to_credentials: str, **kwargs):
+    def with_config(cls, config: Mapping[str, Any]) -> 'BigQuery':
+        return cls(**config['BigQuery'])
```
wangxiaoyou1993, line 103: Let's define those config keys as constants and raise exception with good error message when config is invalid.

wangxiaoyou1993 (approved):

mage_ai/data_preparation/templates/data_exporters/bigquery.py
skunichetty: Fixed, templates now use `get_repo_path`

mage_ai/io/bigquery.py
skunichetty: Moved config keys into an enum `IOConfigKeys`.

PR #563, opened 2022-07-07
[sk] Added option to print status of IO procedures

skunichetty:
# Summary
Data loading and exporting now has the option to print verbose output on the exact commands being executed at any point. This verbose output is disabled by default but can be enabled by specifying `verbose=True` when constructing any IO handler.

Example with Postgres

# Tests
Local tests were performed to check that the printing behavior works as expected.

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_ai/io/base.py
+++ b/mage_ai/io/base.py
@@ -45,11 +45,11 @@ class IOPrintHandler:
     def print_msg(self, msg):
         if self.verbose:
             if self.exists_previous_message:
-                print(f'\r├─ ', flush=True)
-            print(f'└─ {msg}...', end='', flush=True)
+                print(f'\r├─ ')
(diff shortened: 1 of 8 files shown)


mage_ai/io/base.py
```diff
@@ -30,11 +31,34 @@ class FileFormat(str, Enum):
 }
 
 
+class IOPrintHandler:
```
wangxiaoyou1993, line 34: Is this specific to the io module? Can it be generalized and put into the logger file? https://github.com/mage-ai/mage-ai/blob/master/mage_ai/shared/logger.py
Let's centralize the logging related classes and methods.

skunichetty: Moved into `logger`.

wangxiaoyou1993 (approved):

PR #587, opened 2022-07-08
[sk] Running pipeline from command line now prints block debug messages

skunichetty:
# Summary
Running the pipeline from the command line now all debug messages from the block. This was done by selectively enabling stdout redirection only when a server-side API call is made.

# Tests
Local tests were performed to confirm that the command line print debug messages as well as the frontend notebook.

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_ai/data_preparation/models/block.py
+++ b/mage_ai/data_preparation/models/block.py
@@ -18,6 +18,7 @@ from mage_ai.server.kernel_output_parser import DataType
 from mage_ai.shared.logger import VerboseFunctionExec
 import os
 import pandas as pd
+import sys
 
 
 class Block:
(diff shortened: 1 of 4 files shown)


dy46 (approved):

mage_ai/data_preparation/models/block.py
```diff
@@ -149,10 +150,10 @@ def delete(self):
             p.delete_block(p.get_block(self.uuid))
         os.remove(self.file_path)
 
-    async def execute(self, analyze_outputs=True, custom_code=None):
+    async def execute(self, analyze_outputs=True, custom_code=None, redirect_outputs=False):
         with VerboseFunctionExec(f'Executing {self.type} block: {self.uuid}'):
             try:
-                output = await self.execute_block(custom_code)
+                output = await self.execute_block(custom_code, redirect_outputs)
```
wangxiaoyou1993, line 156: nit: let's write `custom_code=custom_code, redirect_outputs=redirect_outputs` for keyword arguments.

skunichetty: Fixed.

PR #607, opened 2022-07-11
[dy] Run block execution code in kernel

dy46:
# Summary

Run block execution code in kernel so that we can interrupt the kernel if needed. Also, the code execution time is correct now.

# Tests

local testing. verified that we can interrupt the kernel, and the code block will stop.

(kernel error)
Screen Shot 2022-07-11 at 10 26 27 AM

cc: @wangxiaoyou1993 @skunichetty

### DIFF
--- a/mage_ai/server/utils/output_display.py
+++ b/mage_ai/server/utils/output_display.py
@@ -101,16 +101,19 @@ from mage_ai.data_preparation.repo_manager import get_repo_path
 from mage_ai.shared.array import find
 import pandas as pd
 
-pipeline_uuid=\'{pipeline_uuid}\'
-block_uuid=\'{block_uuid}\'
+def execute_custom_code():
+    pipeline_uuid=\'{pipeline_uuid}\'
+    block_uuid=\'{block_uuid}\'
+    pipeline = Pipeline(pipeline_uuid, get_repo_path())
+    block = pipeline.get_block(block_uuid)
 
-pipeline = Pipeline(pipeline_uuid, get_repo_path())
-block = pipeline.get_block(block_uuid)
-code = \'\'\'
+    code = \'\'\'
 {code}
-\'\'\'
-block_output = block.execute_sync(custom_code=code)
-output = block_output['output']
-df = find(lambda val: type(val) == pd.DataFrame, output)
-df
+    \'\'\'
+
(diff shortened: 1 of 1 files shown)


mage_ai/data_preparation/models/block.py
```diff
@@ -155,25 +155,31 @@ def delete(self):
...
-                self.status = BlockStatus.EXECUTED
-                if analyze_outputs:
-                    self.__analyze_outputs(variable_mapping)
-            except Exception as err:
-                self.status = BlockStatus.FAILED
-                raise err
-            finally:
-                self.__update_pipeline_block()
-        return output
+            return self.execute_sync(analyze_outputs, custom_code, redirect_outputs)
```
wangxiaoyou1993, line 173: nit: Let's pass keyword arguments with the argument names in case the order of the arguments get changed in the future.

mage_ai/data_preparation/models/block.py
```diff
@@ -155,25 +155,31 @@ def delete(self):
             p.delete_block(p.get_block(self.uuid))
         os.remove(self.file_path)
 
+    def execute_sync(self, analyze_outputs=True, custom_code=None, redirect_outputs=False):
+        try:
+            output = self.execute_block(
+                custom_code=custom_code, redirect_outputs=redirect_outputs
+            )
+            self.process_output(output, analyze_outputs)
```
wangxiaoyou1993, line 163: same here, `process_output(output, analyze_outputs=analyze_outputs)`

mage_ai/server/utils/output_display.py
```diff
@@ -93,3 +93,28 @@ def __custom_output():
...
+from mage_ai.data_preparation.repo_manager import get_repo_path
+from mage_ai.shared.array import find
+import pandas as pd
+
+pipeline_uuid=\'{pipeline_uuid}\'
+block_uuid=\'{block_uuid}\'
+
+pipeline = Pipeline(pipeline_uuid, get_repo_path())
+block = pipeline.get_block(block_uuid)
+final_output = []
```
wangxiaoyou1993, line 112: is it used?

dy46:

wangxiaoyou1993 (approved):

PR #609, opened 2022-07-11
[sp] add feature profiles to reports in sidekick

shrey-mage:
# Summary
- see above

# Tests
image

Light theme feature profiles still looks the same after styling tweaks:

image

cc: @johnson-mage @tommydangerous @dy46


johnson-mage:
@shrey-mage the purple columns and headers in dark theme kinda look like links. Can we make the font color the same as the row text?

johnson-mage:
@shrey-mage Is that background text color in the headers permanently like that or cuz the text is highlighted? 
image

johnson-mage:
> @shrey-mage Is that background text color in the headers permanently like that or cuz the text is highlighted? image

Looks kinda strange having it if it is permanently there

PR #612, opened 2022-07-11
[sk] Added support for runtime defined variables

skunichetty:
# Summary
When running a pipeline with `mage_ai.run`, users can now specify runtime specific variables as keyword arguments which become accessible by blocks at execution time. Any variable name can be used except those reserved at block execution runtime, which are:
- `data_loader` for data loaders
- `data_exporter` for data exporters
- `transformer` for transformers

If a name conflict with a reserved name is found, an error is raised and the pipeline execution fails.

## Example Usage
Consider the following data loader block defined by the user in a pipeline, and `filepath` is a runtime specific variable that the user will specify. Then the block code can be written as
```python
import pandas as pd

if 'data_loader' not in globals():
    from mage_ai.data_preparation.decorators import data_loader


@data_loader
def load_data_from_file(**kwargs) -> pd.DataFrame:
    return pd.read_csv(kwargs.get('filepath'))
```
and the pipeline that contains this block can be ran as:
```python
mage_ai.run('sample_pipeline', repo_name='default_repo', filepath='path/to/my/data.csv')
```
where the value specified for `filepath` will be used to populate the variable in the block.

# Tests
Changes were tested locally by creating a pipeline with runtime variables and executing the pipeline with the given parameters in `mage_ai.run`

cc:
@wangxiaoyou1993


wangxiaoyou1993:
We need to make sure there're no syntax/compilation errors when running individual block files

skunichetty:
I realized that too, I think that instead of providing variables just during runtime, there needs to be a more general variable management system. Ideally this would be a python variable management system that runs alongside the rest of the API. I can look into it further.

skunichetty:
Made a few updates:
- Websocket message now accepts a new parameter: `global_vars`, a dictionary of all the global variables to inject into the runtime. This parameter can be left as null to indicate that there are no global variables to inject. This allows the user at runtime to specify the global variables to inject into the runtime. Tagging @dy46 since `server/output_display.py` was modified and you would know if anything was broken.
- Pipeline execution request (POST) now will expect to receive a parameter in the body of the request of the form:
    ```json
    {
         "global_vars" : {}
    }
    ```
    which specifies the global variables to execute the pipeline with.

mage_ai/data_preparation/models/pipeline.py
```diff
@@ -103,7 +102,9 @@ def block_deletable(self, block, widget=False):
     async def execute(
         self,
         analyze_outputs=True,
+        global_vars=None,
         redirect_outputs=False,
+        run_all_blocks=False,
```
dy46, line 107: i don't think we need this anymore

mage_ai/server/utils/output_display.py
```diff
@@ -167,7 +169,11 @@ def execute_custom_code():
     if run_upstream:
         block.run_upstream_blocks()
 
-    block_output = block.execute_sync(custom_code=code)
+    global_vars = {global_vars}
```
dy46, line 172: i'm not sure if this would work. have you tested it locally?

wangxiaoyou1993:
Can you pass the variables in kwargs to the block methods? Referencing a non defined variable in a block file will show syntax errors.

mage_ai/data_preparation/models/block.py
```diff
@@ -359,7 +374,10 @@ def __validate_execution(self, decorated_functions, input_vars):
 
             return block_function
 
-    def execute_block(self, custom_code=None, redirect_outputs=False):
+    def execute_block(self, custom_code=None, redirect_outputs=False, global_vars=None):
+        if global_vars is None:
+            x = {}
```
wangxiaoyou1993, line 379: what's x?

mage_ai/data_preparation/models/block.py
```diff
@@ -390,7 +408,10 @@ def custom_code(function):
                     exec(file.read(), {self.type: block_decorator(decorated_functions)})
             block_function = self.__validate_execution(decorated_functions, input_vars)
             if block_function is not None:
-                outputs = block_function(*input_vars)
+                if global_vars is not None:
```
wangxiaoyou1993, line 411: also check whether it's empty

mage_ai/data_preparation/models/block.py
skunichetty: Typo, meant to type CTRL+x to remove that piece of code. Removed now.

mage_ai/data_preparation/models/block.py
skunichetty: Added check for empty dictionary.

wangxiaoyou1993 (approved):
Could you update code example in the PR description?

PR #633, opened 2022-07-12
[dy] Fix frontend routing

dy46:
# Summary

`/pipelines/{uuid}` was getting appended to the url every time I refreshed the page.

# Tests

tested locally

cc: @johnson-mage


johnson-mage:
@dy46 ah, thx david

PR #648, opened 2022-07-13
[sk] Data loaders now limit the number of rows that can be loaded.

skunichetty:
# Summary
Data loaders now limit loading data frames with at maximum 100000 rows (can be configurable through changing the `limit` parameter with `load`.

Specifically, if loading from:
- A SQL-style database, the query provided will be modified to add `LIMIT [limit]` if a user specified limit either (a) doesn't exist or (b) exceeds the limit specified by `limit`. 
- A file, then if the format is:
    - JSON or CSV - the first `limit` rows will be loaded from file
    - Parquet or HDF5 - the entire data frame will be loaded, but only the first `limit` rows will be returned. 

# Tests
Local testing was performed by querying data warehouses / data bases to load data and write it to file. Unit tests were written to test query modification.

cc:
@wangxiaoyou1993 @dy46


mage_ai/io/base.py
```diff
@@ -4,8 +4,13 @@
 from pandas import DataFrame
 from typing import IO, Any, Callable, Mapping, Union
 import pandas as pd
+import re
 import os
 
+QUERY_ROW_LIMIT = 100000
```
wangxiaoyou1993, line 10: nit: 100_000 for readability

mage_ai/io/base.py
```diff
@@ -40,6 +45,48 @@ def __init__(self, verbose=False) -> None:
...
+        PostgreSQL, Amazon Redshift, Snowflake, and Google BigQuery.
+
+        Args:
+            query (str): The SQL query to modify
+            limit (int): The limit on the number of rows to return.
+
+        Returns:
+            str: Modified query with limit on row count returned.
+        """
+        if query[:6].lower() != 'select':
```
wangxiaoyou1993, line 61: There could be empty lines or comments at the beginning.

wangxiaoyou1993: There could also be WITH at the beginning of the query.

mage_ai/io/base.py
```diff
@@ -4,8 +4,13 @@
 from pandas import DataFrame
 from typing import IO, Any, Callable, Mapping, Union
 import pandas as pd
+import re
```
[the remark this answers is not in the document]
skunichetty: Fixed.

mage_ai/io/base.py
skunichetty: Fixed.

mage_ai/io/base.py
skunichetty: Switched to the method below, this functionality no longer exists.

mage_ai/io/base.py
```diff
@@ -40,6 +45,48 @@ def __init__(self, verbose=False) -> None:
...
+            query (str): The SQL query to modify
+            limit (int): The limit on the number of rows to return.
+
+        Returns:
+            str: Modified query with limit on row count returned.
+        """
+        if query[:6].lower() != 'select':
+            return query
+
+        limit_match = REGEX_LIMIT.search(query)
```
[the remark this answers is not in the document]
skunichetty: I tested this out, seems to work with Redshift, Postgres, BigQuery, and Snowflake. Switched to this method.

mage_ai/io/base.py
```diff
@@ -3,8 +3,13 @@
 from mage_ai.shared.logger import VerbosePrintHandler
 from pandas import DataFrame
 from typing import IO, Any, Callable, Mapping, Union
-import pandas as pd
 import os
+import pandas as pd
+import re
+
+QUERY_ROW_LIMIT = 100_000
+REGEX_FOR = re.compile('FOR', flags=re.IGNORECASE)
```
wangxiaoyou1993, line 11: nit: remove unused constants

wangxiaoyou1993 (approved):

PR #649, opened 2022-07-13
[dy] Execute upstream blocks

dy46:
# Summary

Add option to execute upstream blocks. Right now, it will just run the upstream blocks in the same kernel before running the current block. There isn't a great UI indication that the upstream blocks are being run which I can improve in a future PR.

This PR contains both frontend and backend changes.

# Tests

tested locally

Screen Shot 2022-07-13 at 3 00 08 PM

cc: @wangxiaoyou1993 @johnson-mage @shrey-mage


mage_ai/data_preparation/models/block.py
```diff
@@ -443,6 +532,36 @@ def __store_variables(self, variable_mapping, override=False):
                     uuid,
                 )
 
+    def get_all_upstream_blocks(self) -> List['Block']:
```
wangxiaoyou1993, line 535: nit: let's move public methods above private methods

mage_ai/data_preparation/models/block.py
```diff
@@ -16,12 +18,87 @@
...
 import traceback
 
 
+async def run_blocks(
+    root_blocks: List['Block'],
+    analyze_outputs: bool = True,
+    redirect_outputs: bool = False,
+    selected_blocks: Set[str] = None,
+) -> None:
+    tasks = dict()
```
wangxiaoyou1993, line 34: are there ways to reduce the duplicate code between async and sync methods?

wangxiaoyou1993: ok

PR #651, opened 2022-07-14
[sk] Improved PostgreSQL Data Loader

skunichetty:
# Summary
This PR refactors the PostgreSQL IO handler to avoid using SQLAlchemy. There are a few benefits to this change:
- Reduces the number of project dependencies, especially since SQLAlchemy was only used in a single location in code
- Unifies interface between PostgreSQL IO Handler and handlers for Snowflake and Redshift (SQLAlchemy used a slightly different architecture than regular DB connectors)
- Faster export times. Testing on a data frame with 100,000 rows and columns of varying types (integers, dates, times, datetimes, strings, floats, categories) was reduced from (averaged over 30 trials):
     - 9.414 seconds → 1.228 seconds when the index of the data frame is included
     - 10.109 seconds → 1.219 seconds when the index of the data frame is excluded

The change made was to use PostgreSQLs `COPY TO` query to improve export times in place of `pandas.DataFrame.to_sql()`. To do this, a new framework was added to manage type conversion (mapping Pandas datatypes to PostgreSQL data types). This framework involves:
1. Inferring the pandas datatype of each column of the input data frame
2. Mapping the pandas datatype to a database relative data type (in this case, Postgres data types)
3. Cleaning the pandas data frame to ensure correct type mapping (e.g. `np.timedelta64` has no appropriate type in PostgreSQL, so it is mapped to an integer before exporting)

# Tests
Changes tested locally with PostgreSQL database alongside unit tests.

# cc
@wangxiaoyou1993 @dy46


mage_ai/io/postgres.py
```diff
@@ -73,14 +88,19 @@ def load(self, query_string: str, limit: int = QUERY_ROW_LIMIT, **kwargs) -> Dat
             return read_sql(self._enforce_limit(query_string, limit), self.conn, **kwargs)
 
     def export(
-        self, df: DataFrame, name: str, index: bool = False, if_exists: str = 'replace', **kwargs
+        self,
+        df: DataFrame,
+        table_name: str,
+        index: bool = False,
+        if_exists: str = 'replace',
```
wangxiaoyou1993, line 95: nit: let's create constants/enum for the options

mage_ai/io/postgres.py
```diff
@@ -1,8 +1,17 @@
+from io import StringIO
 from mage_ai.io.base import BaseSQL, QUERY_ROW_LIMIT
+from mage_ai.io.export_utils import (
+    BadConversionError,
+    clean_df_for_export,
+    PandasTypes,
```
wangxiaoyou1993, line 6: nit: order

mage_ai/io/postgres.py
```diff
@@ -73,14 +88,19 @@ def load(self, query_string: str, limit: int = QUERY_ROW_LIMIT, **kwargs) -> Dat
             return read_sql(self._enforce_limit(query_string, limit), self.conn, **kwargs)
 
     def export(
-        self, df: DataFrame, name: str, index: bool = False, if_exists: str = 'replace', **kwargs
+        self,
+        df: DataFrame,
+        table_name: str,
```
wangxiaoyou1993, line 93: I think you also need to specify schema_name to identify a table.

mage_ai/io/postgres.py
```diff
@@ -89,9 +109,118 @@ def export(
...
+
+                df.to_csv(buffer, index=False, header=False)
+                buffer.seek(0)
+                cur.copy_expert(
+                    f'COPY {table_name} FROM STDIN (FORMAT csv, DELIMITER \',\', NULL \'\');',
+                    buffer,
+                )
+            self.conn.commit()
+
+    def __table_exists(self, table_name: str) -> bool:
```
wangxiaoyou1993, line 141: same here. you need both schema_name and table_name to identify a table.

mage_ai/io/postgres.py
skunichetty: Forgot about schemas, added `schema_name` parameter + auto creation of schema if the given schema doesn't exist.

mage_ai/io/postgres.py
skunichetty: Fixed.

mage_ai/io/postgres.py
skunichetty: Added.

wangxiaoyou1993 (approved):

PR #654, opened 2022-07-14
more comfortable installing from docs

carlyjb17:
# Summary

When you have to install, everytime you copy the commend you have to remove the dollar sign, i remove them so is more comfortable and easier

# Tests

Well is a readme change i just had to render it
cc:


tommydangerous (approved):
🚢 Ship it!

PR #655, opened 2022-07-14
[dy] Fix start up issues

dy46:
# Summary

* try other ports if the initial port is not available.
* add `--service-ports` to the docker command which seems to fix localhost connection issue.

# Tests

tested locally

cc: @wangxiaoyou1993


wangxiaoyou1993 (approved):
We'll need to update frontend to use the port from backend as well.

dy46:
good call

PR #667, opened 2022-07-15
[sk] Updated templates + clients to use new IO configuration framework.

skunichetty:
# Summary
Integrated new configuration into templates and clients.

# Tests
Local testing was performed by querying data warehouses and databases using the new framework. Unit tests were updated.

cc:
@wangxiaoyou1993


wangxiaoyou1993:
When making changes to schema of the yaml file, we need to make sure it's backward compatible. If user creates a repo with the old config yaml, there'll be error loading the config.

mage_ai/io/s3.py
```diff
@@ -124,21 +124,33 @@ def export(
...
-            )
-        credentials = ['access_key_id', 'secret_access_key', 'region']
-        parameters = ['aws_access_key_id', 'aws_secret_access_key', 'region_name']
-        for credential, parameter in zip(credentials, parameters):
-            if credential in aws_config:
-                s3_config[parameter] = aws_config[credential]
-        return cls(**s3_config)
+    def with_config(
+        cls,
+        bucket_name: str,
```
wangxiaoyou1993, line 129: Did you test S3 data loader? The template uses `S3.with_config(ConfigFileLoader(config_path, config_profile)).load()` but the with_config method requires bucket_name and object_key.

wangxiaoyou1993:
> I'll add back the old IOConfig loader so that there is still an option for those who are using the old configuration file setup.

How do we use IOConfig since the with_config method is changed?

mage_ai/io/s3.py
skunichetty: Okay yeah I tested the S3 client, but not with the template so I missed this. I'll fix.

skunichetty:
I was thinking that I'll delete the old `IOConfig` but update the new `ConfigFileLoader` to process both formats as well so templates don't have to be changed.

mage_ai/io/file.py
```diff
@@ -38,10 +38,15 @@ def export(self, df: DataFrame, **kwargs) -> None:
...
 
     @classmethod
-    def with_config(cls, config: Mapping[str, Any]) -> 'FileIO':
-        try:
-            return cls(**config[IOConfigKeys.FILE])
-        except KeyError:
-            raise KeyError(
-                f'No configuration settings found for \'{IOConfigKeys.FILE}\' under profile'
-            )
+    def with_config(
```
wangxiaoyou1993, line 41: When are we going to use this method? The type of config is also different here.

skunichetty: Ah I think I messed a couple of things up with some of the interface design here. The problem is that this PR was initially a bit larger but I stripped it down after the fact. I'll have to refactor some of the abstract classes so that `File` doesn't require a `with_config` function.

wangxiaoyou1993:
Also, be careful about making breaking changes (like `with_config` method and the schema of the io config file) in the future PRs. You can add version to the io config file if you're going to update the schema. For now, it's not a big problem since there're not many people using the library.

wangxiaoyou1993 (approved):

PR #674, opened 2022-07-18
[jk] Transformer action templates UI

johnson-mage:
# Summary
- Allow user to select transformer action type when creating transformer block in order to generate template for that action.

# Tests
image

cc:
@shrey-mage  @skunichetty


johnson-mage:
@shrey-mage I fixed that z-index issue. At least, it works for now.
image

PR #679, opened 2022-07-19
[sk] Overflowing integers don't raise errors

skunichetty:
# Summary
Any overflowing integers (including integer strings) that exist will now get mapped to `CATEGORY_HIGH_CARDINALITY` (as often these integers are ID numbers that technically are categorical). Note: when considering overflow, only signed types are considered currently (unsigned integers can hold larger values, but require manipulation on frontend to handle these values since JavaScript will overflow these integers)

# Tests
Local testing was performed, along with the addition of more unit tests to check how large integers are mapped.

cc:
@wangxiaoyou1993 @dy46

### DIFF
--- a/mage_ai/data_cleaner/column_types/column_type_detector.py
+++ b/mage_ai/data_cleaner/column_types/column_type_detector.py
@@ -9,8 +9,6 @@ import pandas as pd
 import re
 
 DATETIME_MATCHES_THRESHOLD = 0.5
-MAX_SIGNED_INTEGER_VALUE = np.iinfo(np.int).max
-MAX_UNSIGNED_INTEGER_VALUE = np.iinfo(np.uint).max
 MAXIMUM_WORD_LENGTH_FOR_CATEGORY_FEATURES = 40
 MULTITHREAD_MAX_NUM_ENTRIES = 50000
 NUMBER_TYPE_MATCHES_THRESHOLD = 0.8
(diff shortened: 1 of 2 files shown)


mage_ai/data_cleaner/column_types/column_type_detector.py
```diff
@@ -172,7 +172,13 @@ def infer_object_type(series, column_name, kwargs):
             ):
                 return ColumnType.ZIP_CODE
             else:
-                return ColumnType.NUMBER
+                clean_series = clean_series.str.replace(r'\.0*', '')
+                try:
+                    clean_series.astype(int)
+                    return ColumnType.NUMBER
+                except OverflowError:
+                    return ColumnType.CATEGORY_HIGH_CARDINALITY
```
wangxiaoyou1993, line 180: Logic for determining categorical feature is in the branch below. You can reuse some of the code.
```
        if clean_series_nunique / length >= 0.8:
            return ColumnType.TEXT

        if clean_series_nunique <= kwargs.get('category_cardinality_threshold', 255):
            return ColumnType.CATEGORY
        else:
            return ColumnType.CATEGORY_HIGH_CARDINALITY
```

skunichetty: Fixed. I initially didn't perform this check on purpose since I assumed that it wouldn't be likely that an integer style string would occur frequently, but I realize  now that especially with user IDs there can actually be a lower cardinality set of integer-like categories.

wangxiaoyou1993 (approved):

PR #681, opened 2022-07-19
[dy] Support only updating one block

dy46:
# Summary

Support only updating one block when saving a pipeline. Right now when a block is executed, we are updating the entire pipeline metadata config which can overwrite updates that happened during the block execution. This is a temporary fix until we can have locks on pipeline editing.

# Tests

tested locally 

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/data_preparation/models/pipeline.py
+++ b/mage_ai/data_preparation/models/pipeline.py
@@ -10,8 +10,6 @@ from mage_ai.data_preparation.models.variable import Variable
 from mage_ai.data_preparation.models.widget import Widget
 from mage_ai.data_preparation.repo_manager import get_repo_path
 from mage_ai.data_preparation.templates.template import copy_template_directory
-from queue import Queue
-import asyncio
 import os
 import shutil
 import yaml
@@ -279,7 +277,6 @@ class Pipeline:
 
         self.save()
         return block
-        return block
 
     def get_block(self, block_uuid, widget=False):
         mapping = self.widgets_by_uuid if widget else self.blocks_by_uuid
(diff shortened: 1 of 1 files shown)


mage_ai/data_preparation/models/pipeline.py
```diff
@@ -311,7 +308,7 @@ def update_block(self, block, upstream_block_uuids=None):
                     ]
                 block.upstream_blocks = self.get_blocks(upstream_block_uuids)
         self.blocks_by_uuid[block.uuid] = block
-        self.save()
+        self.save(block_uuid=block.uuid)
```
wangxiaoyou1993, line 311: this is a little bit tricky. we also update the downstream_blocks field in other blocks in this method.

dy46: oh good point...

mage_ai/data_preparation/models/block.py
```diff
@@ -511,7 +511,6 @@ def update_content(self, content):
         if content != self.file.content():
             self.status = BlockStatus.UPDATED
         self.file.update_content(content)
-        self.__update_pipeline_block()
```
wangxiaoyou1993, line 514: Are we not updating the status here?

mage_ai/data_preparation/models/pipeline.py
dy46: ok i just updated it so that it'll save the entire pipeline if we're updating upstream blocks

mage_ai/data_preparation/models/pipeline.py
```diff
@@ -361,7 +361,17 @@ def delete_block(self, block, widget=False):
         self.save()
         return block
 
-    def save(self):
-        pipeline_dict = self.to_dict()
+    def save(self, block_uuid: str = None):
+        if block_uuid is not None:
+            current_pipeline = Pipeline(self.uuid, self.repo_path)
```
wangxiaoyou1993, line 366: you can do
```
current_pipeline.blocks_by_uuid[block_uuid] = self.get_block(block_uuid)
pipeline_dict = current_pipeline.to_dict()
```

wangxiaoyou1993 (approved):

PR #694, opened 2022-07-21
[dy] Show number of rows and columns in df

dy46:
# Summary

Display number of rows and columns in the dataframe output.

# Tests

Screen Shot 2022-07-21 at 11 20 36 AM

cc: @wangxiaoyou1993


wangxiaoyou1993:
We only return sample rows to UI. But we want to show the actual number of rows of the dataframe. We need to return that from backend.

dy46:
oh oops

dy46:
@wangxiaoyou1993 updated

wangxiaoyou1993 (approved):

PR #695, opened 2022-07-21
[dy] Run all blocks

dy46:
# Summary

Don't check block status when running a pipeline.

# Tests

tested locally

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/data_preparation/models/block.py
+++ b/mage_ai/data_preparation/models/block.py
@@ -34,7 +34,7 @@ async def run_blocks(
     log_func: Callable[[str], None] = None,
     redirect_outputs: bool = False,
     selected_blocks: Set[str] = None,
-    update_status: bool = False,
+    update_status: bool = True,
 ) -> None:
     tasks = dict()
     blocks = Queue()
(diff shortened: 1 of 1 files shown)


mage_ai/__init__.py
```diff
@@ -121,11 +121,10 @@ def clean(
...
 # --------------- Data preparation methods --------------- #
 
-
 def run(pipeline_uuid: str, project_path: str = None, **global_vars) -> None:
     project_path = os.getcwd() if project_path is None else os.path.abspath(project_path)
     sys.path.append(os.path.dirname(project_path))
     pipeline = Pipeline(pipeline_uuid, project_path)
     asyncio.run(
-        pipeline.execute(analyze_outputs=False, global_vars=global_vars, update_status=False)
+        pipeline.execute(analyze_outputs=False, global_vars=global_vars, update_status=True)
```
wangxiaoyou1993, line 129: I purposely set it to False because writing to metadata.yaml throws permission error when I run it in Airflow.

dy46: I see

dy46: Setting this as false can lead to some errors when running the pipeline if the blocks haven't been executed yet

wangxiaoyou1993: Should it just run all the blocks?

dy46: yeah it should, but there's some validation that will fail if the status is not set. i'm looking to see if we can skip that validation

wangxiaoyou1993 (approved):

PR #701, opened 2022-07-22
[sp] update collapsible blocks

shrey-mage:
# Summary
- remove background on "Expand output" caret
- make "Expand output" text clickable in addition to caret
- persist collapsed blocks on page reload

# Tests

cc: @johnson-mage @tommydangerous

### DIFF
--- a/mage_ai/frontend/components/CodeBlock/index.tsx
+++ b/mage_ai/frontend/components/CodeBlock/index.tsx
@@ -65,6 +65,7 @@ import {
 import { PADDING_UNITS, UNIT } from '@oracle/styles/units/spacing';
 import { SINGLE_LINE_HEIGHT } from '@components/CodeEditor/index.style';
 import { executeCode } from '@components/CodeEditor/keyboard_shortcuts/shortcuts';
+import { get, set } from '@storage/localStorage';
 import { indexBy } from '@utils/array';
 import { onError, onSuccess } from '@api/utils/response';
 import { onlyKeysPresent } from '@utils/hooks/keyboardShortcuts/utils';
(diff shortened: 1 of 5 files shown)


mage_ai/frontend/components/PipelineDetail/utils.ts
```diff
@@ -35,3 +36,17 @@ export function initializeContentAndMessages(blocks: BlockType[]) {
     messages: messagesInit,
   };
 }
+
+export function updateCollapsedBlocks(blocks: BlockType[], pipelineUUID: string, newPipelineUUID: string) {
+  blocks.forEach((b) => {
+    set(
+      `${newPipelineUUID}/${b.uuid}/codeCollapsed`,
```
johnson-mage, line 43: Instead of setting individual keys for every single block, can you set a single key for the collapsed blocks and a property in that object for each block?

mage_ai/frontend/pages/pipelines/[...slug].tsx
```diff
@@ -385,17 +385,21 @@ function PipelineDetailPage({
     },
   );
 
-  const [updatePipelineNoContent] = useMutation(api.pipelines.useUpdate(pipelineUUID));
-  const updatePipelineName = useCallback((name: string) => (
```
Comment (johnson-mage) [inline mage_ai/frontend/pages/pipelines/[...slug].tsx:389]: Did removing this useCallback fix any existing bugs?

mage_ai/frontend/components/CodeBlock/index.tsx
```diff
@@ -139,8 +140,12 @@ function CodeBlockProps({
   const [runEndTime, setRunEndTime] = useState<number>(null);
   const [runStartTime, setRunStartTime] = useState<number>(null);
 
-  const blockMenuRef = useRef(null);
+  useEffect(() => {
+    setCodeCollapsed(get(`${pipeline.uuid}/${block.uuid}/codeCollapsed`, false));
```
johnson-mage, line 144: `${pipeline.uuid}/${block.uuid}/codeCollapsed` and `${pipeline.uuid}/${block.uuid}/outputCollapsed` are repeated a lot. What if we use a simple util method to create the key names or at least create a variable in this file that can be reused?

johnson-mage (approved):
approved after addressing comments

mage_ai/frontend/pages/pipelines/[...slug].tsx
Comment (shrey-mage) [reply mage_ai/frontend/pages/pipelines/[...slug].tsx:389]: This was me going back to refactor my `updatePipelineName` implementation, since I had done it in two different statements that could've been combined in one. I wasn't as familiar with the design pattern but I thought this was a good time to revise it

PR #735, opened 2022-07-26
[dy] Update df size

dy46:
# Summary

Move the rows and columns indicator into the extra content block. Also, show the indicator on refresh.

# Tests
Screen Shot 2022-07-26 at 11 27 09 AM

cc: @wangxiaoyou1993 for backend changes. not sure if the way I calculated the df shape is the best

### DIFF
--- a/mage_ai/data_preparation/models/block.py
+++ b/mage_ai/data_preparation/models/block.py
@@ -548,12 +548,13 @@ class Block:
                 )
                 stats = analysis.get('statistics', {})
                 column_types = analysis.get('metadata', {}).get('column_types', {})
+                row_count = stats.get('original_row_count', stats.get('count'))
                 data = dict(
                     sample_data=dict(
                         columns=data.columns.tolist(),
                         rows=data.to_numpy().tolist(),
                     ),
-                    shape=[stats.get('count'), len(column_types)],
+                    shape=[row_count, len(column_types)],
                     type=DataType.TABLE,
                     variable_uuid=v,
                 )
@@ -669,6 +670,7 @@ class Block:
                 try:
                     analysis = clean_data(
                         data_for_analysis,
+                        df_original=data,
                         transform=False,
                         verbose=False,
                     )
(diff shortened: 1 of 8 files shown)


mage_ai/data_preparation/models/block.py
```diff
@@ -540,11 +540,20 @@ def get_outputs(self, sample_count=None):
...
+                    variable_type=VariableType.DATAFRAME_ANALYSIS,
+                )
+                stats = analysis.get('statistics', {})
+                column_types = analysis.get('metadata', {}).get('column_types', {})
                 data = dict(
                     sample_data=dict(
                         columns=data.columns.tolist(),
                         rows=data.to_numpy().tolist(),
                     ),
+                    shape=[stats.get('count'), len(column_types)],
```
wangxiaoyou1993, line 556: we set the maximum row count to 100k when analyzing the dataframe. we might want to store the the accurate row count in dataframe analysis first.

dy46: @wangxiaoyou1993 i updated it so that the original df is passed down into the statistics calculator. right now it's only used for calculating the row count, but i thought it could be useful for other calculations potentially

mage_ai/data_preparation/models/block.py
```diff
@@ -540,11 +540,20 @@ def get_outputs(self, sample_count=None):
...
+                    variable_type=VariableType.DATAFRAME_ANALYSIS,
+                )
+                stats = analysis.get('statistics', {})
+                column_types = analysis.get('metadata', {}).get('column_types', {})
                 data = dict(
                     sample_data=dict(
                         columns=data.columns.tolist(),
                         rows=data.to_numpy().tolist(),
                     ),
+                    shape=[stats.get('original_row_count'), len(column_types)],
```
wangxiaoyou1993, line 556: let's fallback to `stats.get('count')` since the new stats wasn't available in old analysis.

wangxiaoyou1993 (approved):

PR #743, opened 2022-07-26
[sk] Deleting a block also removes downstream charts

skunichetty:
# Summary
When deleting a block, all downstream charts are removed. If a block has both downstream blocks and downstream charts, the block cannot be deleted until all downstream blocks are removed. Chart block files are deleted when their upstream block is deleted.

# Tests
Tested locally.

cc:

### DIFF
--- a/mage_ai/data_preparation/models/pipeline.py
+++ b/mage_ai/data_preparation/models/pipeline.py
@@ -428,7 +428,7 @@ class Pipeline:
                 os.remove(block.file_path)
         shutil.rmtree(self.dir_path)
 
-    def delete_block(self, block, widget=False):
+    def delete_block(self, block, widget=False, commit=True):
         mapping = self.widgets_by_uuid if widget else self.blocks_by_uuid
 
         if block.uuid not in mapping:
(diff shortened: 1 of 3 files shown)


mage_ai/data_preparation/models/pipeline.py
```diff
@@ -434,11 +434,17 @@ def delete_block(self, block, widget=False):
...
+            downstream_block_uuids = [
+                b.uuid for b in block.downstream_blocks if b.type != BlockType.CHART
+            ]
+            if len(downstream_block_uuids) > 0:
+                raise Exception(
+                    f'Block(s) {downstream_block_uuids} are depending on block {block.uuid}'
+                    '. Please remove the downstream blocks first.'
+                )
+            for downstream_block in block.downstream_blocks:
+                downstream_block.delete()
```
wangxiaoyou1993, line 446: we update pipeline config file (self.save()) every time we delete a block. for this case, i think we can delay updating the file after deleting all downstream blocks. so maybe add a param called `commit` to the delete method, set it to False when deleting downstream blocks.

skunichetty: Added commit parameter to `delete`  and `delete_block`. Of note, when checking if the block is used in other pipelines, I had to filter out the current pipeline manually since the pipeline config is not immediately updated.

wangxiaoyou1993 (approved):

PR #759, opened 2022-07-28
[sk] Added transformer actions documentation

skunichetty:
# Summary
Added documentation for using the transformer actions library in the data prep tool.

cc:
@wangxiaoyou1993


docs/actions/transformer_actions/README.md
```diff
@@ -0,0 +1,772 @@
...
+```
+
+Let's break this code down:
+- `build_transformer_action()` creates a transformer action **payload**, which is a dictionary that describes settings for the action to complete. The arguments passed in are
+  - `df` - the data frame to transform, used to infer column types
+  - `action_type` - specifies the action to perform. In this case, we are performing an `IMPUTE` action.
+  - `arguments` - specifies the columns to perform the action on. Currently, the action is configured to impute all columns of the data frame. Since we only want to transform "Cabin", "Age", and "Embarked", we will update this entry to only contain these column names.
+  - `axis` - specifies which axis to perform the action on. Imputing values is done on a per column basis, so this is a column transformation.
+  - `option` - these are extra settings to specify for this action. For an impute transformation, the main option to specify is `'strategy'`, dictating the imputation strategy to use. See [Fill in Missing Values](#fill-in-missing-values) for details on the valid imputation strategies. For now we will use `MODE`, which specifies to fill empty values with the most frequent value per column.
```
wangxiaoyou1993, line 73: nit: options

docs/actions/transformer_actions/README.md
```diff
@@ -0,0 +1,772 @@
...
+| action_code      | Text field for specifying any code such as filters.                                                                                                                                 |
+| action_arguments | The columns to perform the transformation on.                                                                                                                                       |
+| action_options   | Other settings to provide to the transformer action. See [Transformer Action Reference](#transformer-action-reference) for specific details on which actions require which options. |
+| action_variables | Variable metadata for each input argument (column). Specifies column name and type.                                                                                                 |
+| outputs          | Variable metadata for each output column. Specifies column name and type.                                                                                                           |
+
```
wangxiaoyou1993, line 223: Let's also explain what column_type is.

docs/actions/transformer_actions/README.md
```diff
@@ -0,0 +1,772 @@
...
+            "column_type": "number"
+        },
+        "name": {
+            "uuid": "name",
+            "column_type": "text"
+        }
+    }
+}
+```
+The above transformer action imputes missing values in the 'age' and "name" columns with the constant value `32`. Here is a list of the keys used in transformer action payloads:
```
wangxiaoyou1993, line 212: nit: age uses single quote and name uses double quotes. let's make it consistent here. also filling in name with a number 32 might be confusing.

wangxiaoyou1993 (approved):

docs/actions/transformer_actions/README.md
skunichetty: I changed "name" to "hourly_salary" so a number makes sense, and both now use double quotes.

docs/actions/transformer_actions/README.md
skunichetty: Added an appendix with table of all possible column types and an explanation.

docs/actions/transformer_actions/README.md
skunichetty: Fixed.

PR #766, opened 2022-07-29
[sk] File loading now includes checks for limits

skunichetty:
# Summary
Added limit check for reading files, this must've gotten deleted at some point so I added it back.

# Tests
Tested locally with running FileIO and S3 loader

cc:
@wangxiaoyou1993


tommydangerous:
@wangxiaoyou1993 is this still needed?

PR #792, opened 2022-08-02
[dy] Kernel selection

dy46:
# Summary

Allow users to configure their kernel from the frontend.

For the backend, I created an `active_kernel` module that keeps track of which kernel is currently active. It provides helper methods for switching the active kernel as well as other methods for interacting with the kernel.

# Tests

tested locally

Screen Shot 2022-08-02 at 10 26 23 AM

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/server/websocket.py
+++ b/mage_ai/server/websocket.py
@@ -5,8 +5,13 @@ from mage_ai.data_preparation.models.constants import (
 )
 from mage_ai.data_preparation.models.pipeline import Pipeline
 from mage_ai.data_preparation.repo_manager import get_repo_config, get_repo_path
+from mage_ai.server.active_kernel import (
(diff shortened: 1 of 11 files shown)


mage_ai/server/websocket.py
```diff
@@ -41,32 +44,32 @@ def on_close(self):
...
-        manager = KernelManager(**connection, session=session)
-        return manager.client()
+        return get_active_kernel_client()
 
     def on_message(self, raw_message):
         message = json.loads(raw_message)
         custom_code = message.get('code')
         output = message.get('output')
         if output:
+            print('output')
```
wangxiaoyou1993, line 58: nit: remove this

wangxiaoyou1993 (approved):

PR #854, opened 2022-08-12
[dy] Add extras

dy46:
# Summary

Add extras so that users can selectively choose which dependencies they want to add.

# Tests

tested locally

cc: @wangxiaoyou1993


setup.py
```diff
@@ -10,6 +10,8 @@ def readme():
 requirements = []
 with open('requirements.txt') as f:
     for line in f.read().splitlines():
+        if line.startswith('# extras'):
```
dy46, line 13: not sure if it'd be better to do this or create another requirements file for the extras

setup.py
```diff
@@ -35,5 +37,20 @@ def readme():
             'mage=mage_ai.command_line:main',
         ],
     },
-    extras_require={},
+    extras_require={
+        'spark': ['botocore==1.27.19', 'boto3==1.24.19'],
+        'bigquery': ['google-cloud-bigquery==3.2.0', 'db-dtypes==1.0.2'],
+        'postgres': ['psycopg2-binary==2.9.3'],
+        'redshift': ['boto3==1.24.19', 'redshift-connector==2.0.907'],
+        'snowflake': ['snowflake-connector-python==2.7.9'],
```
wangxiaoyou1993, line 45: Let's also add s3 here.

wangxiaoyou1993 (approved):

PR #858, opened 2022-08-13
[dy] Add runtime variables to CLI

dy46:
# Summary

Allow users to configure runtime variables from mage CLI.

Also, update the README/help command to reflect changes.

# Tests

local

example command: `mage run default_repo long_leaf --runtime-vars iterations 7`

cc: @wangxiaoyou1993


dy46:
@wangxiaoyou1993 PTAL. I also updated the format of the CLI commands a bit

mage_ai/cli/main.py
```diff
@@ -1,3 +1,6 @@
+from mage_ai.cli.utils import parse_arguments, parse_runtime_variables
+from mage_ai.data_preparation.variable_manager import get_global_variables
```
wangxiaoyou1993, line 2: Let's put this import statement in its if-else branch so that we don't need to import it for commands that don't use variables.

mage_ai/cli/utils.py
```diff
@@ -0,0 +1,40 @@
+from typing import Dict, List
+import json
+
+
+# currently only parses single value arguments
+def parse_arguments(args: List[str]) -> Dict[str, any]:
```
wangxiaoyou1993, line 6: You can use library argparse: https://docs.python.org/3/library/argparse.html

dy46:
@wangxiaoyou1993 updated to use argparse

mage_ai/cli/main.py
```diff
@@ -0,0 +1,118 @@
...
+        from mage_ai.data_preparation.pipeline_executor import PipelineExecutor
+        from mage_ai.data_preparation.variable_manager import get_global_variables
+        from mage_ai.shared.hash import merge_dict
+
+        parser = argparse.ArgumentParser(description='Run pipeline.')
+        parser.add_argument('repo_path', metavar='project_path', type=str)
+        parser.add_argument('pipeline_uuid', type=str)
+        parser.add_argument('--runtime-vars', nargs="+")
+
+        if len(sys.argv) >= 3:
```
wangxiaoyou1993, line 85: do we throw error when there're fewer than 3 args?

mage_ai/cli/utils.py
wangxiaoyou1993, line 6: is this method still needed?

wangxiaoyou1993 (approved):

PR #861, opened 2022-08-15
Subsample Large DFs to Create Charts

ChadGueli:
# Summary
Rather than use the whole dataset, this PR takes a subsample of 10,000 observations from the data to build the charts. This expedites the chart creation process without dramatically hurting the visual information conveyed by any chart.

Note the more uneven the frequencies in a categorical column, the higher the probability of inaccurate representation. It might be good to add a warning about that somewhere.  

# Tests
Ran the unittest suite.

cc:
@tommydangerous

### DIFF
--- a/mage_ai/data_cleaner/analysis/calculator.py
+++ b/mage_ai/data_cleaner/analysis/calculator.py
@@ -21,6 +21,7 @@ TIMESERIES_COLUMN_TYPES = frozenset(
     ]
 )
 VERBOSE = False
+SUBSAMPLE_COUNT = 10_000
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


tommydangerous:
👀

wangxiaoyou1993:
@ChadGueli Thanks for your PR! We currently already sample 100k rows for data analysis and charts: https://github.com/mage-ai/mage-ai/blob/master/mage_ai/__init__.py#L83-L89
Do you have any particular reasons to further sample it down to 10k rows?

ChadGueli:
@wangxiaoyou1993 I was unaware that this sampling was already taking place. I think that 100k is probably better in the general setting particularly because that recommendations are based on the subsample.

PR #1004, opened 2022-09-06
[dy] Update server routes

dy46:
# Summary

Add routes

# Tests

cc:


wangxiaoyou1993:
Can we add a route for `/images` as well?

PR #1022, opened 2022-09-08
[td] Don’t loop forever

tommydangerous:
# Summary

# Tests

cc:


mage_ai/data_preparation/models/block/__init__.py
```diff
@@ -121,8 +122,18 @@ def run_blocks_sync(
 
     while not blocks.empty():
         block = blocks.get()
-        if block.type == BlockType.SCRATCHPAD:
+
+        if block.type in [BlockType.CHART, BlockType.SCRATCHPAD]:
```
wangxiaoyou1993, line 126: You can use NON_PIPELINE_EXECUTABLE_BLOCK_TYPES

PR #1031, opened 2022-09-08
[dy] Support cron expressions

dy46:
# Summary

Add support for cron expressions for `schedule_interval` field. The UI right now just allows users to write their own cron syntax at the moment.

# Tests

Screen Shot 2022-09-08 at 1 04 38 PM

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/orchestration/db/models.py
+++ b/mage_ai/orchestration/db/models.py
@@ -15,7 +15,6 @@ from sqlalchemy.sql import func
 from typing import Dict, List
 import enum
 
-
 Base = declarative_base()
 Base.query = Session.query_property()
 Base.select = Session.query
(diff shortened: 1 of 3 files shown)


mage_ai/orchestration/db/models.py
```diff
@@ -128,8 +144,9 @@ def current_execution_date(self) -> datetime:
                 timedelta(days=now.weekday())
         elif self.schedule_interval == '@monthly':
             return now.replace(second=0, microsecond=0, minute=0, hour=0, day=1)
-        # TODO: Support cron syntax
-        return None
+        else:
```
wangxiaoyou1993, line 147: can schedule_interval be None?

dy46: @wangxiaoyou1993 what would happen if schedule_interval is None

wangxiaoyou1993: It shouldn't schedule if schedule_interval is None. We should return None in current_execution_date here if schedule_interval is None.

dy46: 👍

wangxiaoyou1993 (approved):

PR #1049, opened 2022-09-14
[dy] Add sensor block type

dy46:
# Summary

Add sensor block type initial implementation

# Tests

cc:


dy46:
@wangxiaoyou1993 PTAL

mage_ai/orchestration/run_status_checker.py
```diff
@@ -0,0 +1,41 @@
+from datetime import datetime
+from mage_ai.orchestration.db.models import BlockRun, PipelineRun, PipelineSchedule
+
+
+def check_status(
+    pipeline_uuid: int,
```
wangxiaoyou1993, line 6: pipeline_uuid is a str

mage_ai/orchestration/run_status_checker.py
```diff
@@ -0,0 +1,41 @@
...
+    if trigger_id is not None:
+        pipeline_run = (
+            PipelineRun
+            .query
+            .filter(PipelineRun.pipeline_schedule_id == trigger_id)
+            .filter(PipelineRun.execution_date == partition)
+            .first()
+        )
+    else:
+        pipeline_schedule = (
```
wangxiaoyou1993, line 20: you can do
```
pipeline_run = (
  PipelineRun
  .query
  .join(PipelineRun.pipeline_schedule)
  .filter(PipelineSchedule.pipeline_uuid == pipeline_uuid)
  .filter(...)
  .first()
)
```

mage_ai/io/s3.py
```diff
@@ -109,6 +109,15 @@ def export(
                 buffer.seek(0)
                 self.client.put_object(Body=buffer, Bucket=bucket_name, Key=object_key)
 
+    def exists(
```
wangxiaoyou1993, line 112: let's also add this method to base class

mage_ai/orchestration/run_status_checker.py
```diff
@@ -0,0 +1,41 @@
...
+    if pipeline_run is None:
+        return False
+
+    if block_uuid is not None:
+        block_run = next(filter(lambda run: run.block_uuid == block_uuid, pipeline_run.block_runs))
+        if block_run is None:
+            return False
+        return block_run.status == BlockRun.BlockRunStatus.COMPLETED
+    else:
+        return pipeline_run.status == PipelineRun.PipelineRunStatus.COMPLETED
```
wangxiaoyou1993, line 41: we probably want to set sensor status to failed if its dependent pipeline/block run's status is failed.

mage_ai/data_preparation/templates/sensors/default.jinja
```diff
@@ -0,0 +1,30 @@
...
+    from mage_ai.data_preparation.decorators import sensor
+{% endblock %}
+
+{% block content %}
+@sensor
+def check_condition(**kwargs) -> None:
+    """
+    Template code for checking if block or pipeline run completed.
+    """
+    # Method to check status of a pipeline or block run
```
wangxiaoyou1993, line 15: The multi line comments should be wrapped with """

mage_ai/data_preparation/templates/sensors/s3.py
```diff
@@ -0,0 +1,32 @@
+{% extends "testable.jinja" %}
```
wangxiaoyou1993, line 1: We might not need test for sensor block since return value is just true or false.

mage_ai/orchestration/run_status_checker.py
dy46: i'll add this as a follow up

PR #1081, opened 2022-09-21
Fix #! in shell scripts

tdhood:
# Summary
My default sh is dash, which does not support things like empty array assignment. Not sure exactly why.
	Ex: POSITIONAL=()
The space in '# !/bin/bash' caused the shebang line to be ignore, and ran using the default sh.

# Tests

```
~ % cat /tmp/bar.sh           
# !/bin/bash

ps aux | grep "$$"

~ % /tmp/bar.sh          
kestrel   260579  0.0  0.0   2704   912 pts/1    S+   23:02   0:00 sh /tmp/bar.sh
kestrel   260581  0.0  0.0   6336  2104 pts/1    S+   23:02   0:00 grep 260579

~ % cat /tmp/foo.sh
#!/bin/bash

ps aux | grep "$$"

~ % /tmp/foo.sh  
kestrel   260726  0.0  0.0   7028  3240 pts/1    S+   23:02   0:00 /bin/bash /tmp/foo.sh
kestrel   260728  0.0  0.0   6336  2232 pts/1    S+   23:02   0:00 grep 260726
```

### DIFF
--- a/scripts/deploy/cloudformation/ecs.sh
+++ b/scripts/deploy/cloudformation/ecs.sh
@@ -1,4 +1,4 @@
-# !/bin/bash
+#!/bin/bash
 
 SUBNET="$1"
 PARAMETERS=ParameterKey=SubnetID,ParameterValue=${SUBNET}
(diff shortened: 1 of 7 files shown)


tommydangerous:
Thank you so much! Looks good. 🚢

tommydangerous (approved):

PR #1151, opened 2022-10-07
[dy] Multi dev environment backend

dy46:
# Summary

Backend and infra changes for supporting multiple dev environments with AWS ECS. I'm still working on the some of the python code to handle errors when fetching and parsing data from AWS.

For the terraform changes, I copied the existing terraform files from the `aws` folder, and only updated the `main.tf` file to add a new dev task definition. The idea is that when we create new dev environments, we will clone the dev task definition, update the `containerPath`, and create a new task based on that updated task definition. 

# Tests

Screen Shot 2022-10-07 at 2 50 41 PM

tested locally

cc: @wangxiaoyou1993


mage_ai/cluster_manager/aws/ecs_task_manager.py
```diff
@@ -0,0 +1,111 @@
...
+    def create_task(self, name: str):
+        region_name = os.getenv('AWS_REGION_NAME', 'us-west-2')
+        config = Config(region_name=region_name)
+        ec2_client = boto3.client('ec2', config=config)
+        ecs_client = boto3.client('ecs', config=config)
+
+        # create new task definition
+        base_td = \
+            ecs_client.describe_task_definition(
+                taskDefinition='mage-dev-dev-task'
```
wangxiaoyou1993, line 49: let's not hard code it.

mage_ai/cluster_manager/aws/ecs_task_manager.py
```diff
@@ -0,0 +1,111 @@
...
+            'registeredBy',
+            'requiresAttributes',
+            'revision',
+            'status',
+            'taskDefinitionArn',
+        ]
+        for arg in remove_args:
+            base_td.pop(arg)
+
+        container_path = f'/home/src/{name}'
```
wangxiaoyou1993, line 63: we don't need to update the container path here. we can mount EFS to the same container path and just use the different folders in the FS.

mage_ai/server/api/environments.py
```diff
@@ -0,0 +1,27 @@
+from .base import BaseHandler
+
+
+class ApiEnvironmentsHandler(BaseHandler):
```
wangxiaoyou1993, line 4: We should give the endpoint a better name. This endpoint is to manage the instances in a cluster instead of managing envrionments.

mage_ai/services/aws/ecs/ecs.py
```diff
@@ -9,3 +12,45 @@ def run_task(command: str, ecs_config: EcsConfig) -> None:
     client = boto3.client('ecs')
     response = client.run_task(**ecs_config.get_task_config(command=command))
     print(json.dumps(response, indent=4, default=str))
+
+
+def create_environment_task(
```
wangxiaoyou1993, line 17: this method can be combined with the run_task method.

mage_ai/server/api/environments.py
```diff
@@ -0,0 +1,27 @@
...
+    def get(self, env_type):
+        environments = []
+        if env_type == 'ecs':
+            from mage_ai.cluster_manager.aws.ecs_task_manager import EcsTaskManager
+            ecs_instance_manager = EcsTaskManager('mage-dev-development-cluster')
```
wangxiaoyou1993, line 9: Let's not hardcode the cluster name.

scripts/deploy/terraform/aws-dev/main.tf
```diff
@@ -0,0 +1,227 @@
...
+
+  vars = {
+    aws_access_key_id     = var.AWS_ACCESS_KEY_ID
+    aws_secret_access_key = var.AWS_SECRET_ACCESS_KEY
+    lambda_func_arn       = "${aws_lambda_function.terraform_lambda_func.arn}"
+    lambda_func_name      = "${aws_lambda_function.terraform_lambda_func.function_name}"
+  }
+}
+
+resource "aws_ecs_task_definition" "manager-task" {
```
wangxiaoyou1993, line 44: We'll need a different entry point for manager task. You can override the command for manager task once you add the management page.

dy46:
@wangxiaoyou1993 PTAL

mage_ai/server/server.py
```diff
@@ -420,7 +419,7 @@ def make_app():
         (r'/api/block_runs/(?P<block_run_id>\w+)/outputs', ApiBlockRunOutputHandler),
         (r'/api/block_runs/(?P<block_run_id>\w+)/logs', ApiBlockRunLogHandler),
         (r'/api/clusters/(?P<cluster_type>\w+)', ApiClustersHandler),
-        (r'/api/environments/(?P<env_type>\w+)', ApiEnvironmentsHandler),
+        (r'/api/clusters/(?P<cluster_type>\w+)/instances', ApiInstancesHandler),
```
dy46, line 422: I wasn't sure if I should make this include the cluster name, but it would make the format of the endpoint kind of weird

mage_ai/services/aws/ecs/config.py
```diff
@@ -15,6 +15,7 @@ class EcsConfig(BaseConfig):
     cluster: str
     security_groups: List[str]
     subnets: List[str]
+    tags: List
```
wangxiaoyou1993, line 18: Let's give a default value to tags if it's optional.

mage_ai/server/server.py
wangxiaoyou1993: I think putting the cluster name in the payload is fine.

mage_ai/cluster_manager/aws/ecs_task_manager.py
```diff
@@ -0,0 +1,85 @@
...
+        return tasks
+
+    def create_task(self, name: str, task_definition: str, container_name: str):
+        region_name = os.getenv('AWS_REGION_NAME', 'us-west-2')
+        config = Config(region_name=region_name)
+        ec2_client = boto3.client('ec2', config=config)
+
+        # create new task
+        response = list_tasks(self.cluster_name)['tasks']
+        network_interface = self.__get_network_interfaces(response, ec2_client)[0]
```
wangxiaoyou1993, line 51: If we only need the first network interface, why do we fetch all network interfaces? there'll be unnecessary api calls.

wangxiaoyou1993:
It's hard to review the terraform code since I can't compare the difference.
I think we we should be reusable terraform module to reuse the code. It can be done in a separate PR if it's not easy.

dy46:
I can do it in a future PR

wangxiaoyou1993 (approved):

PR #1169, opened 2022-10-12
[jk] Allow user to create streaming pipelines

johnson-mage:
# Summary
- Update UI to support creating streaming pipelines, in addition to standard (batch) pipelines.
- Initially only supporting loading data from Kafka and exporting data to OpenSearch.

# Tests
Add streaming pipeline:
Add streaming pipeline

Create trigger for streaming pipeline (only "Schedule" type triggers supported for streaming pipelines, not "Event" or "API" triggers):
Create trigger for streaming pipeline

cc:
@wangxiaoyou1993


wangxiaoyou1993:
Looks good! Let's hide SQL in transformers for now since we don't support it yet.

johnson-mage:
> Looks good! Let's hide SQL in transformers for now since we don't support it yet.

Okay

PR #1204, opened 2022-10-18
[jk] Show logs in reverse chronological order

johnson-mage:
# Summary
- See title.

# Tests


wangxiaoyou1993:
We should click the button at the top show older logs.

PR #1211, opened 2022-10-19
Simplify the code related to the Catalog class

csharplus:
# Summary

Some code in the Catalog class was not very intuitive, which was cleaned up.

# Tests

CI
cc:


tommydangerous (approved):
You are amazing!

PR #1274, opened 2022-10-28
[dy] Data integration transformer block backend

dy46:
# Summary

# Tests

cc:


mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -14,6 +14,7 @@
 from mage_ai.orchestration.execution_process_manager import execution_process_manager
 from mage_ai.orchestration.notification.config import NotificationConfig
 from mage_ai.orchestration.notification.sender import NotificationSender
+from mage_ai.data_preparation.models.constants import BlockType
```
wangxiaoyou1993, line 17: why do you import it twice?

mage_integrations/mage_integrations/sources/base.py
```diff
@@ -193,7 +193,7 @@ def process(self) -> None:
             ))
             raise Exception(message)
 
-    def process_stream(self, stream):
+    def process_stream(self, stream, properties: Dict = None):
```
wangxiaoyou1993, line 196: Could you update the method comment as well?

mage_integrations/mage_integrations/sources/base.py
```diff
@@ -309,7 +312,7 @@ def sync_stream(self, stream) -> List[Dict]:
 
         return rows
 
-    def sync(self, catalog: Catalog) -> None:
+    def sync(self, catalog: Catalog, properties: Dict = None) -> None:
```
wangxiaoyou1993, line 315: same. update method comments.

mage_integrations/mage_integrations/sources/utils.py
```diff
@@ -219,7 +219,7 @@ def parse_args(required_config_keys):
         help='Sources will log using Singer logger by default. Set his flag to True to log to sys.stdout.',
     )
 
-    args = parser.parse_args()
+    args, unknown = parser.parse_known_args()
```
wangxiaoyou1993, line 222: nit: `args, _ = ...`

mage_integrations/mage_integrations/transformers/base.py
```diff
@@ -0,0 +1,177 @@
...
+    def process(self, input_buffer) -> None:
+        if self.to_df:
+            self.transform_input(input_buffer)
+        else:
+            catalog = self.catalog or self.discover(streams=self.selected_streams)
+            dtypes = {k: dict(type=[v]) for k, v in infer_dtypes(self.df).items()}
+
+            self.sync(catalog, dtypes)
+
+    def __emit_state(self, state):
```
wangxiaoyou1993, line 164: why don't you update this method in parent class to be protected method?

mage_integrations/mage_integrations/transformers/base.py
```diff
@@ -0,0 +1,177 @@
...
+            self.block_uuid,
+            'df',
+            partition=self.execution_partition,
+            variable_type=VariableType.DATAFRAME,
+        )
+
+    def process_schema(
+        self,
+        stream: str,
+        schema: dict,
```
wangxiaoyou1993, line 80: nit: let's use Dict as the type

mage_integrations/mage_integrations/transformers/base.py
```diff
@@ -0,0 +1,177 @@
+from datetime import datetime
+from mage_ai.data_preparation.models.pipeline import Pipeline
```
wangxiaoyou1993, line 2: there could be circular dependencies if you reference mage_ai library here since mage_ai library depends on mage_integrations as well.

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -299,7 +300,7 @@ def run_integration_pipeline(
             _start_date=start_date,
         )
 
-        for idx, block_run in enumerate([data_loader_block_run, data_exporter_block_run]):
+        for idx, block_run in enumerate(block_runs):
```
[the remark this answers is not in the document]
dy46: @wangxiaoyou1993 doesn't the pipeline scheduler also just get the blocks in this order and execute them?

dy46:
@wangxiaoyou1993 address your comments PTAL

mage_ai/orchestration/pipeline_scheduler.py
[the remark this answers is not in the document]
wangxiaoyou1993: no, pipeline scheduler doesn't guarantee to get the block runs in sequence

mage_integrations/mage_integrations/destinations/base.py
```diff
@@ -118,6 +119,10 @@ def config(self) -> Dict:
             self.logger.exception(message)
             raise Exception(message)
 
+    @config.setter
+    def config(self, config):
```
wangxiaoyou1993, line 123: where is it used?

mage_integrations/mage_integrations/destinations/base.py
```diff
@@ -128,6 +133,10 @@ def settings(self) -> Dict:
 
         return {}
 
+    @settings.setter
+    def settings(self, settings):
```
wangxiaoyou1993, line 137: same here. where is it used?

mage_integrations/mage_integrations/destinations/base.py
```diff
@@ -331,3 +329,14 @@ def __validate_and_prepare_record(
         self.validators[stream].validate(record)
 
         return flatten_record(record)
+
+    def _emit_state(self, state):
```
wangxiaoyou1993, line 333: no need to move this method below `_validate_and_prepare_record`

wangxiaoyou1993:
Did you use commands to test locally or UI? Could you provide an example?

mage_integrations/mage_integrations/destinations/base.py
dy46: @wangxiaoyou1993 this is a bit a hack because the transformer class inherits from both `Source` and `Destination`. The source class sets the `config` variable, but it won't be able to set it in the transformer if the destination class's `config` is immutable. Since the transformer only is used as either a source or destination at the moment, I left it like this.

dy46:
yeah i was able to test it locally. it's a bit hard to show through screenshots, but i can add some

dy46:
but also it could be helpful to test with some more extensive transformations, i'm not sure what kind of transformations people are doing through the tool

mage_integrations/mage_integrations/destinations/base.py
wangxiaoyou1993: ok

wangxiaoyou1993: could you add comment for this?

wangxiaoyou1993 (approved):
Looks good after addressing comments.

PR #1362, opened 2022-11-11
[dy] Add partition keys to schema for Bigquery

dy46:
# Summary

Support partition keys for big query. Bigquery only supports one partition key field, but other destinations can support more, so the `partition_keys` field is an array. This adds the `partition_keys` field to the schema field in the data loader yaml.

# Tests

tested locally

Screenshot 2022-11-10 at 9 31 17 PM

Screenshot 2022-11-10 at 9 27 23 PM

cc: @wangxiaoyou1993


wangxiaoyou1993 (approved):

wangxiaoyou1993:
Could you also test the bigquery destination to make sure the insert command works?

dy46:
@wangxiaoyou1993 I checked the destination table, and the rows seem to be inserted correctly.

PR #1449, opened 2022-11-23
One hot encode added to transform

ajatau:
One Hot Encoding is used to convert numerical categorical variables into binary vectors. This is generally performed during the data preprocessing step.
Mage-Ai includes some common used transformations but one hot encoding is not a part of the tool right now. 

User is able to select “One Hot Encoding” from the “Transformer” option on the tool. Once the use selects it, the transformation step is added in the pipeline and return the data including  the one hot encoded (get dummies) is performed.

**Additional context**
I intend to implement using the scikit learn library

### DIFF
--- a/mage_ai/data_preparation/templates/transformers/transformer_actions/column/one_hot_encode.py
+++ b/mage_ai/data_preparation/templates/transformers/transformer_actions/column/one_hot_encode.py
@@ -1,0 +1,18 @@
+{% extends "transformers/transformer_actions/action.jinja" %}
+{% block imports %}
+{{ super() -}}
+{% endblock %}
+{% block action %}
+    """
+    Execute Transformer Action: ActionType.ONE_HOT_ENCODE
+
+    Docs: https://github.com/mage-ai/mage-ai/blob/master/docs/actions/transformer_actions/README.md#one-hot-encode
+    """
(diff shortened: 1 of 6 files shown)


tommydangerous:
👀

mage_ai/data_preparation/templates/transformers/transformer_actions/column/one_hot_encode.py
```diff
@@ -0,0 +1,18 @@
+{% extends "transformers/transformer_actions/action.jinja" %}
+{% block imports %}
+{{ super() -}}
+{% endblock %}
+{% block action %}
+    """
+    Execute Transformer Action: ActionType.ONE_HOT_ENCODE
+
+    Docs: https://github.com/mage-ai/mage-ai/blob/master/docs/actions/transformer_actions/README.md#one-hot-encode
```
wangxiaoyou1993, line 9: Could you add doc for this transformer action in this doc: https://github.com/mage-ai/mage-ai/blob/master/docs/guides/transformer-blocks.mdx#transformer-action-reference? And you can use doc link https://docs.mage.ai/guides/transformer-blocks#one-hot-encode here.

wangxiaoyou1993:
Could you include how you tested this PR in PR description?

PR #1474, opened 2022-11-29
[dy] Add option to run integration streams in parallel

dy46:
# Summary

Run streams in parallel if they are marked as parallel. I added the `run_in_parallel` config to the stream catalog entry.

I updated the `run_integration_pipeline` method in `pipeline_scheduler` to queue a job for each parallel stream and another job for all non-parallel streams. The code that is run for each stream should be unchanged, but the structure of an integration pipeline run is different.

Screenshot 2023-07-12 at 6 05 58 PM

# Tests

tested locally

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/orchestration/pipeline_scheduler.py
+++ b/mage_ai/orchestration/pipeline_scheduler.py
@@ -642,6 +642,10 @@ def run_integration_pipeline(
             pipeline_scheduler,
             pipeline_run_id,
             variables,
+            all_block_runs,
+            block_runs,
+            executable_block_run_ids,
+            pipeline_run,
         )
     )
 
@@ -678,22 +682,35 @@ def run_integration_stream(
     pipeline_scheduler,
     pipeline_run_id,
     variables,
+    all_block_runs,
+    block_runs,
+    executable_block_run_ids,
+    pipeline_run,
 ):
     tap_stream_id = stream['tap_stream_id']
     destination_table = stream.get('destination_table', tap_stream_id)
 
-    
-    block_runs_for_stream = (
-        BlockRun.query
-        .filter(BlockRun.id.in_(executable_block_runs))
-        .filter(BlockRun.block_uuid.like(f'%{tap_stream_id}%'))
-    )
+    block_runs_for_stream = list(filter(lambda br: tap_stream_id in br.block_uuid, block_runs))
+    if len(block_runs_for_stream) == 0:
+        return
+
     indexes = [0]
     for br in block_runs_for_stream:
         parts = br.block_uuid.split(':')
         if len(parts) >= 3:
             indexes.append(int(parts[2]))
     max_index = max(indexes)
+
+    all_block_runs_for_stream = list(filter(
+        lambda br: tap_stream_id in br.block_uuid,
+        all_block_runs,
+    ))
+    all_indexes = [0]
+    for br in all_block_runs_for_stream:
+        parts = br.block_uuid.split(':')
+        if len(parts) >= 3:
+            all_indexes.append(int(parts[2]))
+    max_index_for_stream = max(all_indexes)
 
     for idx in range(max_index + 1):
         block_runs_in_order = []
@@ -702,8 +719,9 @@ def run_integration_stream(
         while True:
             block_runs_in_order.append(
                 find(
-                    lambda b: b.block_uuid == f'{current_block.uuid}:{tap_stream_id}:{idx}',
-                    block_runs_for_stream,
+                    lambda b: b.block_uuid ==
+                    f'{current_block.uuid}:{tap_stream_id}:{idx}',  # noqa: B023
+                    all_block_runs,
                 )
             )
             downstream_blocks = current_block.downstream_blocks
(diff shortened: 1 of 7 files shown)


dy46:
@wangxiaoyou1993 I updated this PR, can you take a look?

wangxiaoyou1993:
> @wangxiaoyou1993 I updated this PR, can you take a look?

Yes, will take a look

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -472,32 +486,156 @@ def __schedule_blocks(self, block_runs: List[BlockRun] = None) -> None:
                 self.__build_tags(**tags),
             )
 
-    def __schedule_integration_pipeline(self, block_runs: List[BlockRun] = None) -> None:
-        if job_manager.has_pipeline_run_job(self.pipeline_run.id):
-            return
-
-        block_runs_to_schedule = self.initial_block_runs if block_runs is None else block_runs
-        block_runs_to_schedule = self.__fetch_crashed_block_runs() + block_runs_to_schedule
+    def __schedule_integration_streams(self, block_runs: List[BlockRun] = None) -> None:
```
wangxiaoyou1993, line 489: can we add comments to this method since it's a long method?

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -472,32 +486,156 @@ def __schedule_blocks(self, block_runs: List[BlockRun] = None) -> None:
...
+        if block_runs is not None:
+            block_runs_to_schedule = block_runs
+        else:
+            # Fetch all "in progress" blocks to handle crashed block runs
```
wangxiaoyou1993, line 493: This code below doesn't fetch crashed block runs. It just fetches blocks that are not completed.

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -472,32 +486,156 @@ def __schedule_blocks(self, block_runs: List[BlockRun] = None) -> None:
...
+            if (not sequential_streams or job_manager.has_pipeline_run_job(self.pipeline_run.id)) \
+                    and len(parallel_streams_to_schedule) == 0:
+                return
+
+            variables = self.pipeline_run.get_variables(extra_variables={
+                'pipeline.name': self.pipeline.name,
+                'pipeline.uuid': self.pipeline.uuid,
+                'pipeline_uuid': self.pipeline.uuid,
+            })
+            integration_pipeline = IntegrationPipeline.get(self.pipeline.uuid)
```
wangxiaoyou1993, line 535: Is it same with self.pipeline?

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -472,32 +486,156 @@ def __schedule_blocks(self, block_runs: List[BlockRun] = None) -> None:
...
+                _execution_date=execution_date.isoformat(),
+                _execution_partition=self.pipeline_run.execution_partition,
+                _start_date=start_date,
             )
 
+            data_loader_block = integration_pipeline.data_loader
+            data_exporter_block = integration_pipeline.data_exporter
+
+            all_block_runs = BlockRun.query.filter(BlockRun.pipeline_run_id == self.pipeline_run.id)
+            block_runs = list(filter(lambda br: br.id in executable_block_runs, all_block_runs))
```
wangxiaoyou1993, line 584: It seems `all_block_runs` and `block_runs` are not needed in this method.

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -472,32 +486,156 @@ def __schedule_blocks(self, block_runs: List[BlockRun] = None) -> None:
...
+
+            for stream in parallel_streams_to_schedule:
+                tap_stream_id = stream.get('tap_stream_id')
+                block_runs_for_stream = list(filter(
+                    lambda br: tap_stream_id in br.block_uuid,
+                    block_runs_to_schedule,
+                ))
+                for b in block_runs_for_stream:
```
wangxiaoyou1993, line 592: A batch update is more efficient here.
```
    BlockRun.batch_update_status(
        [b.id for b in block_runs_for_stream],
        BlockRun.BlockRunStatus.QUEUED,
    )
```

wangxiaoyou1993: Also, I'm not sure whether it's necessary to update the status of the block runs to QUEUED since the block runs in the same stream are executed in sequence.

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -472,32 +486,156 @@ def __schedule_blocks(self, block_runs: List[BlockRun] = None) -> None:
...
+                    len(sequential_streams) == 0:
+                return
+
+            for stream in sequential_streams:
+                tap_stream_id = stream.get('tap_stream_id')
+                block_runs_for_stream = list(filter(
+                    lambda br: tap_stream_id in br.block_uuid,
+                    block_runs_to_schedule,
+                ))
+                for b in block_runs_for_stream:
```
wangxiaoyou1993, line 621: Same here. Is it necessary to update the block run statuses?

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -566,222 +704,186 @@ def __run_heartbeat(self) -> None:
...
-def run_integration_pipeline(
+def run_integration_streams(
+    streams,
+    *args,
+):
+    for stream in streams:
+        run_integration_stream(stream, *args)
+
+
+def run_integration_stream(
```
wangxiaoyou1993, line 715: Could you add comments/docstring to this method as well?

mage_ai/orchestration/pipeline_scheduler.py
dy46: yeah we actually don't need it

mage_ai/orchestration/pipeline_scheduler.py
dy46: yes, this is on purpose to handle cases where block runs never completed because their parent process crashed.

dy46:
@wangxiaoyou1993 sorry for the delay, I addressed your comments

wangxiaoyou1993 (approved):

PR #1477, opened 2022-11-30
[jk] Limit log fetching

johnson-mage:
# Summary
- Fetch logs in intervals instead of all at once.
- Fixed issue where clicking on a log detail would reset the logs viewed if viewing past logs.

# Tests
logs

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_ai/server/api/logs.py
+++ b/mage_ai/server/api/logs.py
@@ -62,7 +62,7 @@ class ApiPipelineLogListHandler(BaseHandler):
             select(*columns).
             join(b, a.pipeline_schedule_id == b.id).
             filter(b.pipeline_uuid == pipeline_uuid).
-            order_by(a.execution_date.desc())
+            order_by(a.created_at.desc())
         )
 
         if len(pipeline_schedule_ids):
--- a/mage_ai/server/api/logs.py
+++ b/mage_ai/server/api/logs.py
@@ -88,6 +88,8 @@ class ApiPipelineLogListHandler(BaseHandler):
                 filter(a.execution_date <= end_timestamp)
             )
 
+        total_pipeline_run_log_count = query.count()
+
         pipeline_run_logs = []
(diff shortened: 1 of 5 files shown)


mage_ai/server/api/logs.py
```diff
@@ -60,7 +61,8 @@ def get(self, pipeline_uuid):
             PipelineRun.
             select(*columns).
             join(b, a.pipeline_schedule_id == b.id).
-            filter(b.pipeline_uuid == pipeline_uuid)
+            filter(b.pipeline_uuid == pipeline_uuid).
+            order_by(a.execution_date.desc())
```
johnson-mage, line 65: @wangxiaoyou1993 If I want to fetch the latest logs first, is this the right way to order them? It's so I can set the `_offset` and `_limit` initially at 0 and 20, respectively, and then increase them in intervals. But there seems to be some mix of logs not chronologically dated correctly when fetching log files.

For example in the summary gif, the oldest log in the first request is from 11/1/22, but when fetching the next (older) interval of logs, it includes logs from 11/9/22 as well.

mage_ai/server/api/logs.py
```diff
@@ -109,9 +117,11 @@ def get(self, pipeline_uuid):
             ])).
             join(a, a.id == c.pipeline_run_id).
             join(b, a.pipeline_schedule_id == b.id).
-            filter(b.pipeline_uuid == pipeline_uuid)
+            filter(b.pipeline_uuid == pipeline_uuid).
+            order_by(c.started_at.desc())
```
johnson-mage, line 121: Related to https://github.com/mage-ai/mage-ai/pull/1477/files#r1035480073

wangxiaoyou1993: this works

wangxiaoyou1993 (approved):

PR #1489, opened 2022-12-01
[dy] Refactor block

dy46:
# Summary

Refactor the `block/__init__.py` file into different block types. There was a lot of custom logic in the `Block` class for different types of blocks which made the code pretty hard to read and maintain, so I tried to create block subclasses for all the different types. This PR focuses primarily on the `execute_block` method.

# Tests

testing locally for all block types
- [x] python
- [x] integration
- [x] dbt
- [ ] sql
- [ ] r   

cc: @wangxiaoyou1993


mage_ai/data_preparation/models/block/__init__.py
```diff
@@ -345,6 +302,31 @@ def after_create(self, block, **kwargs):
...
+        if block_type == BlockType.CHART:
+            return Widget
+        if pipeline and PipelineType.INTEGRATION == pipeline.type:
+            if BlockType.DATA_LOADER == block_type:
+                return SourceBlock
+            elif BlockType.DATA_EXPORTER == block_type:
+                return DestinationBlock
+            else:
+                return TransformerBlock
+        elif BlockLanguage.SQL == language:
```
wangxiaoyou1993, line 324: The BlockLanguage for DBT block is also SQL. Did you test DBT blocks?

dy46: hmm I did test DBT blocks, but let me double check

PR #1490, opened 2022-12-01
[jk] Show recent pipeline run status on Triggers page

johnson-mage:
# Summary
- See title.

# Tests
image

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_ai/server/api/orchestration.py
+++ b/mage_ai/server/api/orchestration.py
@@ -378,7 +378,8 @@ class ApiPipelineScheduleListHandler(BaseHandler):
                         order_by(PipelineRun.created_at.desc())
                     )
                     latest_run = runs.first()
-                    schedule['latest_run_status'] = latest_run.status
+                    if latest_run is not None:
+                        schedule['latest_run_status'] = latest_run.status
         except Exception as err:
             raise err
             collection = []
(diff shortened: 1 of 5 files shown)


mage_ai/server/api/orchestration.py
```diff
@@ -369,6 +369,17 @@ def get(self, pipeline_uuid=None):
                 collection = [r.to_dict(include_attributes=['event_matchers', 'pipeline_runs_count'])
                               for r in results]
                 collection.sort(key=lambda d: d['id'], reverse=True)
+
+                for schedule in collection:
```
johnson-mage, line 373: @wangxiaoyou1993 Can you check my backend code? I had to add the latest pipeline run status to the triggers list so we could get that data with one API request. I feel like there may be a cleaner way to write this code though.

mage_ai/server/api/orchestration.py
```diff
@@ -369,6 +369,17 @@ def get(self, pipeline_uuid=None):
...
+                    runs = (
```
wangxiaoyou1993, line 374: no need to do additional DB queries. you can do `schedule.pipeline_runs` to get all pipeline runs.
You can sort them with the code like
```
sorted(ps.pipeline_runs, key=lambda x: x.created_at)
```

johnson-mage: thank you xiaoyou!

PR #1500, opened 2022-12-02
[dy] Update for Chargebee and BigQuery

dy46:
# Summary

Small updates from stress testing Chargebee and BigQuery.

Removed `table` and added `location` to the BigQuery destination config.

# Tests

tested locally

cc: @wangxiaoyou1993

### DIFF
--- a/mage_integrations/mage_integrations/destinations/bigquery/templates/config.json
+++ b/mage_integrations/mage_integrations/destinations/bigquery/templates/config.json
@@ -2,5 +2,5 @@
   "path_to_credentials_json_file": "path_to_credentials_json_file",
   "project_id": "",
   "dataset": "",
-  "table": ""
+  "location": ""
 }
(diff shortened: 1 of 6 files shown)


mage_integrations/mage_integrations/destinations/bigquery/templates/config.json
```diff
@@ -2,5 +2,5 @@
   "path_to_credentials_json_file": "path_to_credentials_json_file",
   "project_id": "",
   "dataset": "",
-  "table": ""
```
wangxiaoyou1993, line 5: actually, let's keep 'table' here. we still need it when we test destination individually via bash command.

dy46: @wangxiaoyou1993 you should still be able to add it to the config, it just won't be shown to the user through the UI.

wangxiaoyou1993 (approved):

PR #1515, opened 2022-12-03
[jk] Update once trigger with completed run to inactive

johnson-mage:
# Summary
- Updates pipeline schedule pertaining to a pipeline run to `inactive` status if the trigger was scheduled to run only once and was `active` status. Changes made on the backend.
- Added refresh interval to Triggers list page so it periodically checks for updates to the trigger and latest run statuses.

# Tests

### DIFF
--- a/mage_ai/orchestration/pipeline_scheduler.py
+++ b/mage_ai/orchestration/pipeline_scheduler.py
@@ -95,13 +95,13 @@ class PipelineScheduler:
                     )
 
                 pipeline_schedule = PipelineSchedule.query.get(self.pipeline_run.pipeline_schedule_id)
-                if (pipeline_schedule is not None):
-                    if (pipeline_schedule.status == PipelineSchedule.ScheduleStatus.ACTIVE \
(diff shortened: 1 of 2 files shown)


mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -94,6 +94,14 @@ def schedule(self, block_runs: List[BlockRun] = None) -> None:
                         tags=merge_dict(tags, dict(metrics=self.pipeline_run.metrics)),
                     )
 
+                pipeline_schedule = PipelineSchedule.query.get(self.pipeline_run.pipeline_schedule_id)
+                if (pipeline_schedule is not None):
```
dy46, line 98: you don't need the parentheses in python

dy46: you can also combine the two if statements if you want

dy46 (approved):

mage_ai/orchestration/pipeline_scheduler.py
dy46: Yeah it’ll evaluate the first statement first and if it fails it’ll just move on

PR #1532, opened 2022-12-07
[jk] Avoid duplicate pipeline and block run logs

johnson-mage:
# Summary
- See title.

# Tests
Before (note # of "logs found"):
image
before - nondeduped logs

After:
image
after - deduped logs

### DIFF
--- a/mage_ai/server/api/logs.py
+++ b/mage_ai/server/api/logs.py
@@ -98,13 +98,16 @@ class ApiPipelineLogListHandler(BaseHandler):
             else:
                 rows = query.all()
 
+            processed_pipeline_run_log_files = set()
             for row in rows:
                 model = PipelineRun()
                 model.execution_date = row.execution_date
                 model.pipeline_schedule_id = row.pipeline_schedule_id
                 model.pipeline_uuid = row.pipeline_uuid
-
-                pipeline_run_logs.append(model.logs)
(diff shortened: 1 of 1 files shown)


mage_ai/server/api/logs.py
```diff
@@ -98,13 +98,16 @@ def get(self, pipeline_uuid):
...
 
+            processed_pipeline_run_log_files = set()
             for row in rows:
                 model = PipelineRun()
                 model.execution_date = row.execution_date
                 model.pipeline_schedule_id = row.pipeline_schedule_id
                 model.pipeline_uuid = row.pipeline_uuid
-
-                pipeline_run_logs.append(model.logs)
+                pipeline_log_file_path = model.logs.get('path')
```
johnson-mage, line 107: @dy46 So the queries were returning multiple results of the same `execution_date`, `pipeline_schedule_id`, and `pipeline_uuid` and causing the same log file to be rendered multiple times (once for each time the pipeline was re-run). I couldn't get some sort of "distinct" query to return the same row only once, so I did it this way instead.

It works, but do you happen to know a better way off a quick look? Same thing happening for the `pipeline_run_logs` and `block_run_logs`. Thank you!

dy46: this seems ok to me, i don't know off the top of my head if there's a better way

PR #1534, opened 2022-12-07
[dy] Detect first sync

dy46:
# Summary

Added logic to detect if the current sync is a "first sync". The condition was proposed by Tommy to check if the source table has over 2 times the number of records as the destination table. If so, then use the primary key for the order by instead of the bookmark properties.

I'm not sure how to test if this makes a performance difference, do you have any ideas @wangxiaoyou1993?

# Tests
Tested locally with Postgres and Bigquery. 

cc: @wangxiaoyou1993

### DIFF
--- a/mage_integrations/mage_integrations/sources/sql/base.py
+++ b/mage_integrations/mage_integrations/sources/sql/base.py
@@ -232,11 +232,14 @@ WHERE table_schema = '{schema}'
         order_by_columns = []
 
         # This order is very important, don’t change
-        for arr in [
-            bookmark_properties,
+        order_by_properties = [
             key_properties,
             unique_constraints,
-        ]:
+        ]
+        if not self.is_first_sync:
+            order_by_properties = [bookmark_properties] + order_by_properties
+
+        for arr in order_by_properties:
             if not arr:
                 continue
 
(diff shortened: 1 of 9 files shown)


mage_integrations/mage_integrations/sources/sql/base.py
```diff
@@ -232,11 +232,14 @@ def __fetch_rows(
         order_by_columns = []
 
         # This order is very important, don’t change
-        for arr in [
-            bookmark_properties,
+        order_by_properties = [
```
wangxiaoyou1993, line 235: There're some issues here.
1. When we order the table by key_properties, we need to use that key as bookmark properties as well because the original bookmark properties will not be logged in order.
2. After the first sync, we need to fetch the max values of the original bookmark properties and then update the state file for future incremental sync.
3. If there're no key properties, we still need to order the rows by bookmark properties.
4. Regarding performance differences, we tested on a dataset with 16 million rows. Ordering by key columns and fetching sub batch took 30 seconds while ordering by a non-key columns took 3~4 minutes. So i think this strategy will help speed up syncing large datasets.

There're some non-trivial corner cases. Let me know if anything is unclear.

PR #1539, opened 2022-12-08
[dy] Update partition key column name

dy46:
# Summary

Add `clean_column_name` to the partition_key. Also added a try/except to the sql connection execute method because right now, the error doesn't get logged.

# Tests

tested locally

cc: @wangxiaoyou1993

### DIFF
--- a/mage_integrations/mage_integrations/connections/sql/base.py
+++ b/mage_integrations/mage_integrations/connections/sql/base.py
@@ -48,10 +48,13 @@ class Connection(BaseConnection):
             self.logger.info(f'{message} started.')
             now1 = datetime.utcnow().timestamp()
 
-            cursor.execute(clean_query(query_string))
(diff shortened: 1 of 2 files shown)


mage_integrations/mage_integrations/connections/sql/base.py
```diff
@@ -48,10 +48,13 @@ def get_data_from_query_strings(self, cursor, query_strings):
...
-            description = cursor.description
-            if description:
-                data.append(cursor.fetchall())
+            try:
+                cursor.execute(clean_query(query_string))
+                description = cursor.description
+                if description:
+                    data.append(cursor.fetchall())
+            except Exception as err:
+                self.logger.error(f'Error while executing query: {str(err)}')
```
wangxiaoyou1993, line 57: Let's also raise the exception instead of letting it complete.

PR #1543, opened 2022-12-09
[xy] Not commit for every model update.

wangxiaoyou1993:
# Summary

Not commit for every model update.

# Tests

tested locally

cc:

### DIFF
--- a/mage_ai/orchestration/db/models.py
+++ b/mage_ai/orchestration/db/models.py
@@ -66,15 +66,16 @@ class BaseModel(Base):
                 self.session.rollback()
                 raise e
 
-    def update(self, **kwargs) -> None:
+    def update(self, commit=True, **kwargs) -> None:
         for key, value in kwargs.items():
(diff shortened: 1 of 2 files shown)


mage_ai/orchestration/db/models.py
```diff
@@ -66,15 +66,16 @@ def save(self, commit=True) -> None:
                 self.session.rollback()
                 raise e
 
-    def update(self, **kwargs) -> None:
+    def update(self, commit=True, **kwargs) -> None:
```
Mrrobi, line 69: Can you put some docstring? So that it can be more transparent to others?
I guess rather than that its fine

Mrrobi (changes requested):

mage_ai/orchestration/db/models.py
wangxiaoyou1993: Thanks for the suggestions. We will add docstring to the methods.

PR #1625, opened 2022-12-28
[dy] Kubernetes multi dev

dy46:
# Summary

Support Kubernetes for multi dev. In order to create the Kubernetes stateful sets from the manage page, the `KUBE_NAMESPACE` environment variable needs to be set.

# Tests

tested on cloud

cc: @wangxiaoyou1993


requirements.txt
```diff
@@ -32,6 +32,7 @@ google-cloud-bigquery~=2.34.4
 google-cloud-iam==2.9.0
 google-cloud-run==0.5.0
 google-cloud-storage==2.5.0
+kubernetes==25.3.0
```
wangxiaoyou1993, line 35: @dy46 Could you also add this package to `setup.py` `all` extra_requires?
I got this error when calling `/api/status` with latest docker image.
image

PR #1706, opened 2023-01-08
[jl] Update default query limit in docs

jlondonobo:
# Summary
Update the documentation for the **default row limit when loading data**. Increased it from 100,000 to 10,000,000 due to changes in #779.


tommydangerous:
Taking a look, thank you.

wangxiaoyou1993 (approved):
Thank you for fixing the doc!

PR #1714, opened 2023-01-09
[dy] Create job manager

dy46:
# Summary

Create worker pool/manager for running block runs. The worker manager manages a set of workers that will pick up jobs off a multiprocessing `Queue`. Right now, the # of workers is set to be equal to the number of cpu cores, but we can make that number configurable. I created my own worker manager instead of using the multiprocessing `Pool` because I wanted to have more control over the individual processes.

Right now, the worker pool is only being used for standard pipeline block runs since integration pipelines just create 1 multiprocessing process for the pipeline run, and then create subprocesses for the streams. I can work on integration pipelines in the future if this works for standard pipelines.

# Tests

i'm still debugging some issues, but wanted to get some feedback on it.

cc: @wangxiaoyou1993


mage_ai/orchestration/execution_process_manager.py
```diff
@@ -90,3 +92,152 @@ def clean_up_processes(self, include_child_processes=True):
...
+            if not self.queue.empty():
+                args = self.queue.get()
+
+                self.job_uuid = args[0]
+                run_args = args[1:]
+
+                try:
+                    start_session_and_run(*run_args)
+                finally:
+                    self.job_uuid = None
```
wangxiaoyou1993, line 139: If there're exceptions, the process will be terminated. Setting job_uuid does nothing.

mage_ai/orchestration/execution_process_manager.py
```diff
@@ -90,3 +92,152 @@ def clean_up_processes(self, include_child_processes=True):
...
+    def run(self):
+        while True:
+            if not self.queue.empty():
+                args = self.queue.get()
+
+                self.job_uuid = args[0]
+                run_args = args[1:]
+
+                try:
+                    start_session_and_run(*run_args)
```
wangxiaoyou1993, line 137: we don't need to repeatedly start db session for different jobs.

mage_ai/orchestration/execution_process_manager.py
```diff
@@ -90,3 +92,152 @@ def clean_up_processes(self, include_child_processes=True):
...
+            if worker.job_uuid == job_uuid:
+                worker.terminate()
+                self.workers.remove(worker)
+                new_worker = Worker(self.queue)
+                new_worker.start()
+                self.workers.append(new_worker)
+                break
+
+    def is_alive(self, job_uuid) -> bool:
+        for worker in self.workers:
```
wangxiaoyou1993, line 177: the worker(process) could be dead.

mage_ai/orchestration/execution_process_manager.py
wangxiaoyou1993: if the process dies, we'll need to create a new one.

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -269,39 +272,35 @@ def queued_block_runs(self) -> List[BlockRun]:
...
-            proc = create_process(run_block, (
+            job_uuid = worker_manager.add_job(run_block, (
                 self.pipeline_run.id,
                 b.id,
                 get_variables(self.pipeline_run),
                 self.__build_tags(**tags),
             ))
-            execution_process_manager.set_block_process(self.pipeline_run.id, b.id, proc)
-            proc.start()
+            execution_job_manager.set_block_job(self.pipeline_run.id, b.id, job_uuid)
```
wangxiaoyou1993, line 303: You'll need to update the places that calls `execution_process_manager.has_block_process`

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -269,39 +272,35 @@ def queued_block_runs(self) -> List[BlockRun]:
...
-                block_run.update(status=BlockRun.BlockRunStatus.QUEUED)
                 queued_block_runs.append(block_run)
 
         return queued_block_runs
 
     def __schedule_blocks(self, block_runs: List[BlockRun] = None) -> None:
         # TODO: implement queueing logic
         block_runs_to_schedule = self.queued_block_runs if block_runs is None else block_runs
-        block_runs_to_schedule = self.__fetch_crashed_block_runs() + block_runs_to_schedule
+        # block_runs_to_schedule = self.__fetch_crashed_block_runs() + block_runs_to_schedule
```
wangxiaoyou1993, line 282: why do you comment out this?

mage_ai/orchestration/execution_process_manager.py
```diff
@@ -90,3 +92,152 @@ def clean_up_processes(self, include_child_processes=True):
...
+        self.terminate_block_job(pipeline_run_id, block_run_id)
+        if pipeline_run_id not in self.block_jobs:
+            self.block_jobs[pipeline_run_id] = dict()
+        self.block_jobs[pipeline_run_id][block_run_id] = job_uuid
+
+    def terminate_block_job(self, pipeline_run_id: int, block_run_id: int) -> None:
+        if self.has_block_job(pipeline_run_id, block_run_id):
+            worker_manager.cancel_job(self.block_jobs[pipeline_run_id][block_run_id])
+
+    def clean_up_jobs(self, include_child_processes=True):
```
wangxiaoyou1993, line 216: where do you call this method?

mage_ai/orchestration/execution_process_manager.py
dy46: oh good point

PR #1769, opened 2023-01-17
[dy] Add secret var

dy46:
# Summary

Add `secret_var` variable. Syntax is `"secret_var('secret_id', True/False)"`. Right now it will only check AWS secret manager. I was wondering if we should name the variable like `aws_secret_var` to specify AWS because there may be other secret managers in the future.

I also added caching because every request to the API costs money, but you can fetch directly from AWS with by setting the second argument as `False`. 

# Tests

Screenshot 2023-01-17 at 3 51 36 PM

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/data_integrations/utils/config.py
+++ b/mage_ai/data_integrations/utils/config.py
@@ -1,6 +1,5 @@
 from jinja2 import Template
 from mage_ai.data_integrations.utils.parsers import NoDatesSafeLoader
-from mage_ai.services.aws.secrets_manager.secrets_manager import get_secret
 from mage_ai.shared.dates import n_days_ago
 from mage_ai.shared.hash import merge_dict
 from mage_ai.shared.parsers import encode_complex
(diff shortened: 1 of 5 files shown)


mage_ai/data_integrations/utils/config.py
```diff
@@ -1,5 +1,6 @@
 from jinja2 import Template
 from mage_ai.data_integrations.utils.parsers import NoDatesSafeLoader
+from mage_ai.services.aws.secrets_manager.secrets_manager import get_secret
```
wangxiaoyou1993, line 3: in the future, we'll add support to get secrets from gcp secret manager as well. how can we differentiate secrets from aws and gcp?

wangxiaoyou1993: Just saw your description in PR. Yes, let's do `aws_secrets_var`.

dy46: @wangxiaoyou1993 updated

wangxiaoyou1993: If we import `aws.secrets_manager.secrets_manager` here, `boto3` and `botocore` are required packages. How can we make those packages optional?

dy46: @wangxiaoyou1993 I moved the imports to inside the methods. I think that should work

wangxiaoyou1993: Could you check whether `SecretsManager.__init__` is called when you import `get_secret` method?

dy46: you're right, it does call __init__

dy46: @wangxiaoyou1993 ok i updated it again. i tested it on a new virtual env and it worked

wangxiaoyou1993 (approved):

PR #1791, opened 2023-01-19
[dy] Update snowflake convert array

dy46:
# Summary

The queries with arrays in them were failing because the strings weren't being enclosed by quotes. Now the syntax should look something like this `('value1', 'value2', ...)`

# Tests

tested locally

cc: @wangxiaoyou1993

### DIFF
--- a/mage_integrations/mage_integrations/destinations/snowflake/__init__.py
+++ b/mage_integrations/mage_integrations/destinations/snowflake/__init__.py
@@ -21,8 +21,8 @@ from typing import Dict, List, Tuple
 
 def convert_array(value, column_settings):
     if type(value) is list and value:
-        value_string = ', '.join([str(i) for i in value])
-        return value_string
+        value_string = ', '.join([f"'{str(i)}'" for i in value])
+        return f'({value_string})'
 
     return 'NULL'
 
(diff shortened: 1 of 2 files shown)


mage_integrations/mage_integrations/destinations/snowflake/__init__.py
```diff
@@ -21,8 +21,8 @@
 
 def convert_array(value, column_settings):
     if type(value) is list and value:
-        value_string = ', '.join([str(i) for i in value])
-        return value_string
+        value_string = ', '.join([f"'{str(i)}'" for i in value])
```
wangxiaoyou1993, line 24: Could the value be numbers?

dy46: @wangxiaoyou1993 it can be any value, but it doesn't seem like we really check for that

dy46: and it seems like we're already trying to convert everything to a string

dy46:
@wangxiaoyou1993 can you take another look?

wangxiaoyou1993 (approved):

PR #1852, opened 2023-01-27
[dy] Add setting to skip current pipeline run

dy46:
# Summary

If any previous run is still in progress, then the current run will be cancelled.

Screenshot 2023-01-26 at 8 55 00 PM

# Tests

tested locally

cc: @wangxiaoyou1993


mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -716,23 +716,34 @@ def schedule_all():
...
+            running_pipeline_run = find(
+                lambda r: r.status in [
+                    PipelineRun.PipelineRunStatus.INITIAL,
+                    PipelineRun.PipelineRunStatus.RUNNING,
+                ],
+                pipeline_schedule.pipeline_runs
+            )
+            if pipeline_schedule.settings.get('skip_if_previous_running') \
+                    and running_pipeline_run is not None:
+                pipeline_run = PipelineRun.create(**payload)
```
wangxiaoyou1993, line 728: no need to create block run in this case.

mage_ai/server/api/orchestration.py
```diff
@@ -170,7 +170,7 @@ def process_pipeline_runs(
         if additional_results[0].id != results[0].id:
             filter_dates.append(collection[0]['execution_date'])
 
-        last_execution_date = results[-1].execution_date
+        last_execution_date = results[results.count() - 1].execution_date
```
wangxiaoyou1993, line 173: what's the difference here?

dy46: @wangxiaoyou1993 it was erroring for some reason. possibly due to the sqlalchemy upgrade? but i just changed it in case we upgraded

wangxiaoyou1993 (approved):

PR #1893, opened 2023-01-31
Minor code refinements for the BigQuery source

csharplus:
# Summary

Made several minor changes to the code:

* Added type hints to the function, properties and class definition.
* Refined the way the query variable is updated to use the += operator instead of reassigning a new value to the variable.
* Changed the list(map(lambda column: f'{column}', columns)) to a list comprehension [f'{column}' for column in columns]

# Tests

CI

cc:

@tommydangerous


tommydangerous (approved):
Thank you @HenryWuMage !

PR #1915, opened 2023-02-02
[jk] Add return value to mock

johnson-mage:
# Summary
- Fix unit test that was inconsistently throwing an error (though passed most of the time).

# Tests
- Reran `test_sync` unit test many times in order for the error to appear. Reran test many times after updating it, and the error stopped appearing.

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_integrations/mage_integrations/tests/sources/test_base.py
+++ b/mage_integrations/mage_integrations/tests/sources/test_base.py
@@ -292,7 +292,7 @@ class BaseSourceTests(unittest.TestCase):
         catalog.get_selected_streams.return_value = build_sample_streams_catalog_entries()
         source = Source()
         with patch.object(source, 'process_stream', return_value=None) as mock_process_stream:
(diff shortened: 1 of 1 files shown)


mage_integrations/mage_integrations/tests/sources/test_base.py
```diff
@@ -292,7 +292,7 @@ def test_sync(self):
         catalog.get_selected_streams.return_value = build_sample_streams_catalog_entries()
         source = Source()
         with patch.object(source, 'process_stream', return_value=None) as mock_process_stream:
-            with patch.object(source, 'sync_stream') as mock_sync_stream:
+            with patch.object(source, 'sync_stream', return_value=dict()) as mock_sync_stream:
```
wangxiaoyou1993, line 295: return value should be a number https://github.com/mage-ai/mage-ai/blob/master/mage_integrations/mage_integrations/sources/base.py#L377

wangxiaoyou1993 (approved):

PR #1923, opened 2023-02-03
[jk] Separate data integration catalog into json file

johnson-mage:
# Summary
- Save data integration catalog into json file (`data_integration_catalog.json` in individual pipeline folder).
- If user has existing data integration pipeline with catalog data in the data loader block, the `catalog` in the data loader block will be set to an empty object, and the pipeline will read the catalog data from the json file moving forward.

# Tests
New catalog json file created when selecting streams:
catalog json created when selecting streams

Pipeline is successfully ran after saving catalog data to json file:
pipeline run executed using catalog json


mage_ai/data_integrations/utils/config.py
```diff
@@ -33,6 +38,16 @@ def __get_settings(absolute_file_path, variables: Dict = {}, pipeline: 'Pipeline
...
         )
+    elif pipeline and not settings.get('catalog'):
+        catalog_config_path = os.path.join(
+            get_repo_path(),
+            PIPELINES_FOLDER,
+            pipeline.uuid,
+            DATA_INTEGRATION_CATALOG_FILE,
+        )
+        with open(catalog_config_path, 'r') as f:
+            catalog = interpolate_variables(f.read(), variables)
```
johnson-mage, line 49: @wangxiaoyou1993 Though everything seems to run fine when I tested it, this line I was a little unsure if I was calling the  `interpolate_variables` method correctly. Or if `f` versus `f.read()` was the correct first argument to pass.

wangxiaoyou1993:
Could you add unit tests for saving and loading integration pipeline catalog data?

mage_ai/data_preparation/models/pipeline.py
```diff
@@ -729,6 +772,10 @@ def save(self, block_uuid: str = None, widget: bool = False):
                 current_pipeline.blocks_by_uuid[block_uuid] = block
             pipeline_dict = current_pipeline.to_dict()
         else:
+            if self.data_integration is not None:
+                with open(self.catalog_config_path, 'w') as fp:
+                    json.dump(self.data_integration, fp)
+                self.data_integration = None
```
wangxiaoyou1993, line 778: why do we set it to None? you can exclude it from the `to_dict()` method. setting it to None could cause problems.

mage_ai/data_preparation/models/pipelines/integration_pipeline.py
```diff
@@ -111,6 +111,8 @@ def source_file_path(self) -> str:
     @property
     def settings_file_path(self) -> str:
         if self.data_integration and 'catalog' in self.data_integration:
+            if os.path.exists(self.catalog_config_path):
```
wangxiaoyou1993, line 114: simply changing the file path could still cause problem since the file format also changes. we'll need to update mage_integration module to read from json file instead. https://github.com/mage-ai/mage-ai/blob/master/mage_integrations/mage_integrations/sources/utils.py#L279-L281

mage_ai/data_preparation/models/pipeline.py
```diff
@@ -43,7 +45,7 @@ def __init__(self, uuid, repo_path=None, config=None, repo_config=None):
         if config is None:
             self.load_config_from_yaml()
         else:
-            self.load_config(config)
+            self.load_config(config, catalog)
```
wangxiaoyou1993, line 48: for keyword arguments, let's include key in the method call `catalog=catalog`

mage_ai/data_preparation/models/pipeline.py
johnson-mage: Got it. I was testing it to make sure there weren't any issues, but better to not set it to `None`. Was trying to figure out a way to exclude the data_integration prop from the dictionary without having to modify the `to_dict()` method.

mage_ai/data_integrations/utils/config.py
wangxiaoyou1993: this seems useless. you already read the catalog file and set it in `pipeline.data_integration` when loading the pipeline.

wangxiaoyou1993 (approved):

PR #1977, opened 2023-02-10
[jk] Add new custom block

johnson-mage:
# Summary
- User can add a new "custom" block and choose its color in the Pipeline Editor.
- Custom blocks are a way for users to organize their pipeline with blocks that may not fit into one of the other block types (e.g. `data_loader`, `transformer`, `data_exporter`, etc.). These blocks are run as part of the pipeline, unlike `scratchpad` blocks.
- Custom blocks are saved to the `custom` directory.

# Tests
Adding custom block:
Custom block

Pipeline completed with custom blocks (includes both python and sql):
custom pipeline run


johnson-mage:
@wangxiaoyou1993 Could you please review my backend changes? It's mainly adding an additional attribute (`color`) and type (`custom`) to the Block model so we can allow users to choose colors for their custom blocks.

There were also a few lines I had to change so the name of the custom block directory would be `custom` (since `customs` sounded weird and had a different meaning).

mage_ai/data_preparation/models/block/__init__.py
```diff
@@ -210,6 +211,7 @@ def __init__(
         name: str,
         uuid: str,
         block_type: BlockType,
+        block_color: BlockColor = None,
```
dy46, line 214: I think it'd make more sense to just create a new Block for the custom block, and override some of the methods, kind of like the `SensorBlock`. https://github.com/mage-ai/mage-ai/blob/master/mage_ai/data_preparation/models/block/__init__.py#L1791

johnson-mage: ah ok

PR #1996, opened 2023-02-14
[dy] Add mage secrets

dy46:
# Summary

Support creating encrypted secrets within Mage. I originally tried to put the secrets in the variables tab of the sidekick, but they are actually quite different from variables, so I moved them to a separate tab.

Screenshot 2023-02-14 at 10 56 12 AM

# Tests
tested locally

cc: @wangxiaoyou1993


mage_ai/data_preparation/repo_manager.py
```diff
@@ -120,3 +130,51 @@ def set_repo_path(repo_path: str) -> None:
 
 def get_variables_dir(repo_path: str = None) -> str:
     return get_repo_config(repo_path=repo_path).variables_dir
+
+
+def create_secret(name: str, value: str):
```
wangxiaoyou1993, line 135: Move the secrets related methods into a separate file. It doesn't seem that they should be in this repo_manager.py.

mage_ai/data_preparation/repo_manager.py
```diff
@@ -120,3 +130,51 @@ def set_repo_path(repo_path: str) -> None:
...
+    from mage_ai.orchestration.db.models import Secret
+    secrets_dir = os.path.join(
```
wangxiaoyou1993, line 137: The secrets dir should not be in the repo (code)'s folder. We want to separate the code and the data. For example, in the [local to cloud](https://docs.mage.ai/production/ci-cd/local-cloud/repository-setup) mode, user deploys their prebuilt docker image to cloud, any change to the repo's folder won't be persisted on cloud.
We can put the secret key in data's folder.

mage_ai/data_preparation/shared/utils.py
```diff
@@ -0,0 +1,20 @@
+from mage_ai.data_preparation.repo_manager import get_secrets
+from typing import Callable, Dict
+import os
+
+
+def get_template_vars() -> Dict[str, Callable]:
+    kwargs = dict(
+        env_var=os.getenv,
+        mage_secret_var=lambda x: get_secrets().get(x),
```
wangxiaoyou1993, line 9: Getting all the secrets is too expensive. We can just query the secret by name.

dy46:
@wangxiaoyou1993 addressed comments

wangxiaoyou1993 (approved):

PR #2004, opened 2023-02-15
added pipedrive source ( still in dev)

Luishfs:
# Summary
Adds pipedrive source for data integration pipelines ( not working)

# Tests

# Issues
just a quick question here, i'm sending the replication_method both on metadata and on all CatalogEntry entries. I'm seeing each replication_method appear on the Catalog, but when i execute the pipelina via trigger, i receive this error https://github.com/mage-ai/mage-ai/blob/76f2615526e28c004bb91e04762bb7fdfe7e726d/mage_integrations/mage_integrations/destinations/sql/base.py#L153

I checked the replication_method and replication_methods variable via logging and they appear as None and {"schema_name:None} respectivly, i tried forcing an FULL_TABLE method using the function, but got no results as well. Any ideia on why its receving a None in place of REPLICATION_METHOD?

cc:
@tommydangerous


Luishfs:
closing PR to overwrite with new changes.

PR #2020, opened 2023-02-16
[dy] Datadog source

dy46:
# Summary

Datadog source. Waiting on customer to add `metrics` stream.

# Tests

tested locally

cc:


mage_integrations/mage_integrations/sources/datadog/streams/__init__.py
```diff
@@ -0,0 +1,290 @@
...
+    https://docs.datadoghq.com/api/latest/users/#list-all-users
+    """
+    TABLE = 'users'
+    ENTITY = 'user'
+    API_METHOD = 'GET'
+    SCHEMA = 'users'
+
+    current_page = 0
+
+    def get_url(self):
```
wangxiaoyou1993, line 282: nit: if base url is same for all streams, you can specify a base_url constant and only customize path for each stream.

wangxiaoyou1993 (approved):

PR #2050, opened 2023-02-22
[jk] Update block templates

johnson-mage:
# Summary
- Update data loader, transformer, and data exporter templates to not require DataFrame.

# Tests
- Updated unit tests. Created new blocks and confirmed they are using the updated templates.

### DIFF
--- a/mage_ai/data_preparation/templates/data_exporters/default.jinja
+++ b/mage_ai/data_preparation/templates/data_exporters/default.jinja
@@ -1,16 +1,14 @@
-from pandas import DataFrame
-
 if 'data_exporter' not in globals():
     from mage_ai.data_preparation.decorators import data_exporter
 
 
 @data_exporter
-def export_data(df: DataFrame, **kwargs):
+def export_data(*args, **kwargs):
     """
     Exports data to some source
 
     Args:
-        df (DataFrame): Data frame to export to
+        df (optional): Data frame to export to
 
     Output (optional):
         Optionally return any object and it'll be logged and
--- a/mage_ai/data_preparation/templates/data_exporters/default.jinja
+++ b/mage_ai/data_preparation/templates/data_exporters/default.jinja
@@ -8,7 +8,7 @@ def export_data(*args, **kwargs):
     Exports data to some source
 
     Args:
-        df (optional): Data frame to export to
+        args: The input variables from upstream blocks
 
     Output (optional):
         Optionally return any object and it'll be logged and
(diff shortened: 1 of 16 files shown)


mage_ai/tests/data_preparation/test_templates.py
```diff
@@ -268,85 +266,37 @@ def test_template_generation_data_loader_streaming(self):
...
-        config2 = {'axis': 'row'}
-        config3 = {}
-        new_template1 = fetch_template_source(BlockType.TRANSFORMER, config1)
-        new_template2 = fetch_template_source(BlockType.TRANSFORMER, config2)
-        new_template3 = fetch_template_source(BlockType.TRANSFORMER, config3)
-        self.assertEqual(expected_template, new_template1)
-        self.assertEqual(expected_template, new_template2)
-        self.assertEqual(expected_template, new_template3)
-
-    def test_template_generation_transformer_action_default(self):
```
johnson-mage, line 316: Removed because this was  a duplicate unit test.

mage_ai/data_preparation/templates/data_exporters/r/default.jinja
```diff
@@ -1,5 +1,5 @@
-export_data <- function(df_1, ...) {
+export_data <- function(arg, ...) {
```
wangxiaoyou1993, line 1: You don't need to update the code templates for R block.

mage_ai/data_preparation/templates/data_exporters/default.jinja
```diff
@@ -1,16 +1,14 @@
...
 
 @data_exporter
-def export_data(df: DataFrame, **kwargs):
+def export_data(*args, **kwargs):
     """
     Exports data to some source
 
     Args:
-        df (DataFrame): Data frame to export to
+        df (optional): Data frame to export to
```
wangxiaoyou1993, line 11: `df` is not in the args anymore. So you can write
```
args: The input variables from upstream blocks
```

mage_ai/data_preparation/templates/transformers/default.jinja
```diff
@@ -1,27 +1,25 @@
...
     Add more parameters to this function if this block has multiple parent blocks.
     There should be one parameter for each output variable from each parent block.
 
     Args:
-        df (DataFrame): Data frame from parent block.
+        df (optional): Data frame from parent block.
 
     Returns:
-        DataFrame: Transformed data frame
+        Anything (e.g. df)
```
wangxiaoyou1993, line 20: nit: `e.g. data frame, dictionary, array, int, str, etc.`

mage_ai/data_preparation/templates/data_exporters/default.jinja
johnson-mage: Okay, I had left the `df` in there as an "optional" arg so users would know they can still pass in data frames, but that should be implied.

mage_ai/data_preparation/templates/transformers/default.jinja
johnson-mage: Will update in all the places where it says "Returns anything"

PR #2063, opened 2023-02-23
[dy] Add block callbacks

dy46:
# Summary

Add block callbacks for non integration blocks. The callback is shown in the UI as another code editor section beneath the main code block. To write a callback, a user would use the `@on_success` and `@on_failure` decorators. I was thinking we could potentially remove the import statements for `on_success` and `on_failure` to conserve space, but then the callback code wouldn't compile properly outside of Mage.

I implemented callbacks as a `CallbackBlock` block type, but they are treated as mostly just an extension of a block. I chose to make a subtype because the `Block` type has a lot of helper functions to help write and read from files.

# Tests

Screenshot 2023-02-22 at 10 59 43 PM

tested locally, will add unit tests

cc: @wangxiaoyou1993


mage_ai/data_preparation/models/block/__init__.py
```diff
@@ -1859,3 +1921,67 @@ def execute_block_function(
                 print('Sensor sleeping for 1 minute...')
                 time.sleep(60)
             return []
+
+
+class CallbackBlock(Block):
```
wangxiaoyou1993, line 1926: It'll be bettter to put this class in a separate file.

dy46: ok i can move it out

PR #2102, opened 2023-03-02
[dy] Allow block runs to fail

dy46:
# Summary

Add setting to allow block runs in a pipeline schedule to fail. The downstream blocks will be marked as `upstream_failed` and other blocks will continue running.

Screenshot 2023-03-01 at 4 17 09 PM

# Tests

tested locally, unit tests

cc: @wangxiaoyou1993


mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -317,9 +331,13 @@ def queued_block_runs(self) -> List[BlockRun]:
                 completed = block is not None and \
                     block.all_upstream_blocks_completed(completed_block_uuids)
 
+            failed_block_uuids = set(b.block_uuid for b in self.failed_block_runs)
             if completed:
                 block_run.update(status=BlockRun.BlockRunStatus.QUEUED)
                 queued_block_runs.append(block_run)
+            elif any(b in failed_block_uuids for b in block.upstream_block_uuids):
+                block_run.update(
```
wangxiaoyou1993, line 339: Why do we update status in this method?

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -109,7 +115,7 @@ def schedule(self, block_runs: List[BlockRun] = None) -> None:
         if PipelineType.STREAMING == self.pipeline.type:
             self.__schedule_pipeline()
         else:
-            if self.pipeline_run.all_blocks_completed():
+            if self.pipeline_run.all_blocks_completed(self.allow_blocks_to_fail):
```
wangxiaoyou1993, line 118: We should still update the pipeline run status to FAILED if there're failed blocks.

mage_ai/orchestration/pipeline_scheduler.py
dy46: because we were already iterating through the block runs and updating status, but yeah it's probably not the best place for this

PR #2117, opened 2023-03-03
[jk] Cancel all running pipeline runs

johnson-mage:
# Summary
- Add button to cancel all running pipeline runs for a pipeline.

# Tests
cancel running runs

### DIFF
--- a/mage_ai/api/resources/PipelineResource.py
+++ b/mage_ai/api/resources/PipelineResource.py
@@ -2,8 +2,9 @@ from mage_ai.api.resources.BaseResource import BaseResource
 from mage_ai.data_preparation.models.block.dbt.utils import add_blocks_upstream_from_refs
 from mage_ai.data_preparation.models.pipeline import Pipeline
 from mage_ai.data_preparation.repo_manager import get_repo_path
-from mage_ai.orchestration.db import db_connection, safe_db_query
+from mage_ai.orchestration.db import safe_db_query
 from mage_ai.orchestration.db.models import PipelineSchedule, PipelineRun
+from mage_ai.orchestration.pipeline_scheduler import PipelineScheduler
 from mage_ai.server.active_kernel import switch_active_kernel
 from mage_ai.server.kernels import PIPELINE_TO_KERNEL_NAME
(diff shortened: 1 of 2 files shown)


mage_ai/api/resources/PipelineResource.py
```diff
@@ -146,16 +147,33 @@ def update_schedule_status(status, pipeline_uuid):
...
+            pipeline_runs = (
+                PipelineRun.
+                query.
+                filter(PipelineRun.pipeline_uuid == pipeline_uuid).
+                filter(or_(
+                    PipelineRun.status == PipelineRun.PipelineRunStatus.INITIAL,
+                    PipelineRun.status == PipelineRun.PipelineRunStatus.RUNNING,
+                ))
+            )
+            for pipeline_run in pipeline_runs:
```
wangxiaoyou1993, line 161: * We can batch update the pipeline run status.
* We also need to cancel the block runs
* I think this method is only used for cancelling pipeline run. So you can just call it `cancel_pipeline_runs`

mage_ai/api/resources/PipelineResource.py
```diff
@@ -146,16 +147,33 @@ def update_schedule_status(status, pipeline_uuid):
             for schedule in schedules:
                 schedule.update(status=status)
 
+        @safe_db_query
+        def update_pipeline_run_status(status, pipeline_uuid):
+            pipeline_runs = (
+                PipelineRun.
+                query.
+                filter(PipelineRun.pipeline_uuid == pipeline_uuid).
+                filter(or_(
```
wangxiaoyou1993, line 156: you can do 
```
filter(PipelineRun.status.in_([PipelineRun.PipelineRunStatus.INITIAL, PipelineRun.PipelineRunStatus.RUNNING]))
```

mage_ai/api/resources/PipelineResource.py
johnson-mage: @wangxiaoyou1993 There was a PipelineScheduler method that canceled the pipeline and block runs. Does this work?

wangxiaoyou1993 (approved):

PR #2127, opened 2023-03-06
[dy] Support dbt-sqlserver

dy46:
# Summary

Add support for using sqlserver with dbt in Mage. Will add docs. Follows the schema here:
https://docs.getdbt.com/reference/warehouse-setups/mssql-setup

Also I used pyodbc instead of pymssql for the mssql io file because the dbt-sqlserver package uses pyodbc. Also, dbt-sqlserver doesn't technically support dbt-core 1.4.0 yet, but it seems to work based on my testing.

# Tests

tested locally.

cc: @wangxiaoyou1993


Dockerfile
```diff
@@ -3,6 +3,13 @@ LABEL description="Deploy Mage on ECS"
 ARG PIP=pip3
 USER root
 
+# Download ODBC headers for pyodbc
+RUN curl https://packages.microsoft.com/keys/microsoft.asc | apt-key add -
+RUN curl https://packages.microsoft.com/config/debian/11/prod.list > /etc/apt/sources.list.d/mssql-release.list
+RUN apt-get update
```
wangxiaoyou1993, line 9: there're duplicate code on `apt update`

mage_ai/io/mssql.py
```diff
@@ -0,0 +1,145 @@
...
+class MSSQL(BaseSQL):
+    def __init__(
+        self,
+        database: str,
+        host: str,
+        password: str,
+        user: str,
+        schema: str = None,
+        port: int = 1433,
+        charset='utf8',
```
wangxiaoyou1993, line 20: nit: type

mage_ai/io/mssql.py
```diff
@@ -0,0 +1,145 @@
...
+            schema=config[ConfigKey.MSSQL_SCHEMA],
+            driver=config[ConfigKey.MSSQL_DRIVER],
+            host=config[ConfigKey.MSSQL_HOST],
+            password=config[ConfigKey.MSSQL_PASSWORD],
+            port=config[ConfigKey.MSSQL_PORT],
+            user=config[ConfigKey.MSSQL_USER],
+        )
+
+    def open(self) -> None:
+        with self.printer.print_msg('Opening connection to MySQL database'):
```
wangxiaoyou1993, line 54: nit: MSSQL

mage_ai/io/mssql.py
```diff
@@ -0,0 +1,145 @@
...
+        sql = f'INSERT INTO {full_table_name} VALUES ({values_placeholder})'
+        cursor.executemany(sql, values)
+
+    def get_type(self, column: Series, dtype: str) -> str:
+        if dtype in (
+            PandasTypes.MIXED,
+            PandasTypes.UNKNOWN_ARRAY,
+            PandasTypes.COMPLEX,
+        ):
+            raise BadConversionError(
```
wangxiaoyou1993, line 100: Can we use TEXT for this case?

dy46: i can try

Dockerfile
dy46: i removed the other apt update

wangxiaoyou1993: Could you use `apt -y update` here

dy46: oops ok let me try that

Dockerfile
```diff
@@ -3,8 +3,15 @@ LABEL description="Deploy Mage on ECS"
...
+RUN ACCEPT_EULA=Y apt-get install -y msodbcsql18
+RUN apt install unixodbc-dev
```
wangxiaoyou1993, line 11: Add `-y` here as well

PR #2131, opened 2023-03-07
[dy] Add env var to disable edit access

dy46:
# Summary

Support disabling edit access for endpoints related to the notebook. I had to make some changes to the authentication middleware to support the new policy condition.

# Tests

tested locally. trying to perform an edit action will lead to an error pop up

Screenshot 2023-03-07 at 9 58 56 AM

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/api/utils.py
+++ b/mage_ai/api/utils.py
@@ -1,6 +1,6 @@
 from mage_ai.orchestration.db.models import Oauth2AccessToken
 from mage_ai.settings import (
-    DISABLE_EDIT_ACCESS,
+    DISABLE_NOTEBOOK_EDIT_ACCESS,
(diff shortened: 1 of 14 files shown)


mage_ai/api/utils.py
```diff
@@ -37,6 +40,10 @@ def has_at_least_editor_role(user) -> bool:
         (user.roles and user.roles & 2 != 0)
 
 
+def has_at_least_editor_role_and_edit_access(user) -> bool:
+    return not DISABLE_EDIT_ACCESS and has_at_least_editor_role()
```
wangxiaoyou1993, line 44: do you need to pass the user to `has_at_least_editor_role`?

PR #2136, opened 2023-03-07
[jk] Migrate additional resources when pipeline renamed

johnson-mage:
# Summary
- When renaming a pipeline, transfer existing triggers, backfills, pipeline runs, and block runs (blocks already transferred currently).
- Prevent users from renaming pipeline to a name already in use by another pipeline.

# Tests
- Added unit tests.
- Renaming a pipeline with an existing pipeline name:
pipeline name taken
- Renaming pipeline:
renaming pipeline

### DIFF
--- a/mage_ai/data_preparation/models/pipeline.py
+++ b/mage_ai/data_preparation/models/pipeline.py
@@ -507,6 +507,11 @@ class Pipeline:
             old_uuid = self.uuid
             new_name = data['name']
             new_uuid = clean_name(new_name)
+
+            all_pipelines = self.get_all_pipelines(self.repo_path)
+            if new_uuid in all_pipelines:
+                raise Exception(f'Pipeline {new_uuid} already exists. Choose a different name.')
+
             old_pipeline_path = self.dir_path
             self.name = new_name
             self.uuid = new_uuid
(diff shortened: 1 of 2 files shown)


mage_ai/data_preparation/models/pipeline.py
```diff
@@ -499,18 +500,39 @@ async def to_dict_async(
...
+            }, synchronize_session=False)
+            # Migrate pipeline runs (block runs have foreign key ref to PipelineRun id)
+            PipelineRun.query.filter(PipelineRun.pipeline_uuid == old_uuid).update({
+                PipelineRun.pipeline_uuid: new_uuid
+            }, synchronize_session=False)
+            # Migrate backfills
+            Backfill.query.filter(Backfill.pipeline_uuid == old_uuid).update({
+                Backfill.pipeline_uuid: new_uuid
+            }, synchronize_session=False)
+            db_connection.session.commit()
```
dy46, line 534: maybe create a separate method for these, and add the `@safe_db_query` decorator like here: https://github.com/mage-ai/mage-ai/blob/master/mage_ai/orchestration/db/models.py#L531

dy46 (approved):

PR #2161, opened 2023-03-09
[jk] Update API for cloning pipelines

johnson-mage:
# Summary
- Backend updates for cloning pipelines

# Tests
- Add unit tests.
- Clone standard pipeline:
copy standard pipeline
- Clone integration pipeline:
copy integration pipeline

### DIFF
--- a/mage_ai/data_preparation/models/pipeline.py
+++ b/mage_ai/data_preparation/models/pipeline.py
@@ -140,7 +140,6 @@ class Pipeline:
         self,
         source_pipeline: 'Pipeline',
         duplicate_pipeline_name: str,
-        included_blocks: List[str] = [],
     ):
         duplicate_pipeline = self.create(
             duplicate_pipeline_name,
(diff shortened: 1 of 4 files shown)


mage_ai/data_preparation/models/pipeline.py
```diff
@@ -133,45 +134,55 @@ def create(self, name, pipeline_type=PipelineType.PYTHON, repo_path=None):
         return pipeline
 
     @classmethod
-    def duplicate(cls, source_pipeline: 'Pipeline', duplicate_pipeline_name: str):
-        duplicate_pipeline = cls.create(
-            duplicate_pipeline_name,
+    def duplicate(
+        self,
```
dy46, line 138: classmethods are generally supposed to use `cls` instead of `self`

mage_ai/data_preparation/models/pipeline.py
```diff
@@ -133,45 +134,55 @@ def create(self, name, pipeline_type=PipelineType.PYTHON, repo_path=None):
...
+                import IntegrationPipeline
+
+            return IntegrationPipeline(
+                duplicate_pipeline_uuid,
+                repo_path=duplicate_pipeline.repo_path,
             )
-        duplicate_pipeline.save()
-        return duplicate_pipeline
+
+        return self(
```
dy46, line 182: can you use `Pipeline.get` here?

dy46 (commented):
what was the reason for changing the way we duplicate the pipelines?

johnson-mage:
> what was the reason for changing the way we duplicate the pipelines?

That method was done really early on by Sachchit, and it only copied over blocks and widgets but not other things like variables. Also, it didn't work properly for integration pipelines when testing it (it didn't copy over the configuration for the streams and source/destinations).

Seemed like copying over the yaml config file was an easier way to copy over the blocks/widgets/etc instead of looping through the source pipeline blocks multiple times as well, though not sure if there's a best practice here we should go with.

mage_ai/data_preparation/models/pipeline.py
johnson-mage: Will try that out

mage_ai/data_preparation/models/pipeline.py
johnson-mage: So i saw the python linting suggest that also, but I changed it to "self" to make it consistent with our other methods. I'll change it back to `cls` though.

mage_ai/data_preparation/models/pipeline.py
johnson-mage: Cool, it worked. I updated the code.

dy46 (approved):
ok this way does seem better than looping through everything

PR #2170, opened 2023-03-09
[dy] Update block run status logic

dy46:
# Summary

Update `allow_block_to_fail` logic for pipelines that have dynamic blocks. A block with upstream dynamic blocks will only not run if *all* of the upstream dynamic blocks failed.

# Tests

tested locally. added some dynamic block unit tests, but it's hard to mock everything that happens for the dynamic blocks :(

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/orchestration/pipeline_scheduler.py
+++ b/mage_ai/orchestration/pipeline_scheduler.py
@@ -333,6 +333,15 @@ class PipelineScheduler:
             )
 
             if dynamic_upstream_block_uuids:
+                if self.allow_blocks_to_fail:
+                    completed_block_uuids = [
+                        b.block_uuid for b in self.pipeline_run.block_runs
+                        if b.status in [
+                            BlockRun.BlockRunStatus.COMPLETED,
+                            BlockRun.BlockRunStatus.UPSTREAM_FAILED,
+                            BlockRun.BlockRunStatus.FAILED,
+                        ]
+                    ]
                 completed = all(uuid in completed_block_uuids
                                 for uuid in dynamic_upstream_block_uuids)
             else:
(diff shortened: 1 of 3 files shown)


mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -333,6 +333,15 @@ def executable_block_runs(self) -> List[BlockRun]:
             )
 
             if dynamic_upstream_block_uuids:
+                if self.allow_blocks_to_fail:
+                    completed_block_uuids = [
```
wangxiaoyou1993, line 337: we don't need to calculate it for each block_run. the `completed_block_uuids` above also doesn't need to be calculated for each block_run.

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -346,12 +355,41 @@ def executable_block_runs(self) -> List[BlockRun]:
         return executable_block_runs
 
     def __update_block_run_statuses(self) -> None:
-        failed_block_uuids = set(b.block_uuid for b in self.failed_block_runs)
+        failed_block_uuids = [
```
wangxiaoyou1993, line 358: it'll be better to use `set`

mage_ai/orchestration/pipeline_scheduler.py
wangxiaoyou1993: It's also better to use `set`

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -346,12 +355,41 @@ def executable_block_runs(self) -> List[BlockRun]:
...
+                    for b in block.upstream_block_uuids
+                ):
+                    block_run.update(
+                        status=BlockRun.BlockRunStatus.UPSTREAM_FAILED)
+                    updated_status = True
+
+        self.pipeline_run.refresh()
+        # keep iterating through block runs until no more updates can be made
+        if updated_status:
+            self.__update_block_run_statuses()
```
wangxiaoyou1993, line 392: Let's make sure there's no infinite loop.

wangxiaoyou1993 (approved):

PR #2187, opened 2023-03-11
[jk] Pipeline table filtering

johnson-mage:
# Summary
- Add filtering (by status and type) for pipelines.
- Add ToggleMenu component that could be used for filtering in other pages.

# Tests
pipeline filtering


mage_ai/api/resources/PipelineResource.py
```diff
@@ -20,6 +20,22 @@ async def collection(self, query, meta, user, **kwargs):
         if include_schedules:
             include_schedules = include_schedules[0]
 
+        pipeline_types = query.get('type[]', [None])
```
johnson-mage, line 23: @dy46 Does the logic and updates in this file (`mage_ai/api/resources/PipelineResource.py`) in particular seem okay? The rest of the PR is FE changes.

dy46: can you just set the default here to be an empty array? then you don't have to worry about there being a `None` element

mage_ai/api/resources/PipelineResource.py
```diff
@@ -56,13 +72,29 @@ def query_pipeline_schedules(pipeline_uuids):
...
             if mapping.get(pipeline.uuid):
                 schedules = mapping[pipeline.uuid]
             pipeline.schedules = schedules
 
+            if pipeline_statuses and (
+                ('active' in pipeline_statuses and
+                    any(s.status == 'active' for s in pipeline.schedules)) or
+                ('inactive' in pipeline_statuses and
+                    len(pipeline.schedules) > 0 and
+                    all(s.status == 'inactive' for s in pipeline.schedules)) or
```
dy46, line 90: i would use the enum value here instead of writing out the string

mage_ai/api/resources/PipelineResource.py
johnson-mage: I cleaned it up a bit in the latest commit: https://github.com/mage-ai/mage-ai/pull/2187/commits/31df5b3304fd016bb9c5501610fae278d6eefb05

PR #2219, opened 2023-03-17
[dy] Add sensor templates for sql databases

dy46:
# Summary

Add sensor templates for SQL databases

# Tests

tested locally

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/data_preparation/shared/secrets.py
+++ b/mage_ai/data_preparation/shared/secrets.py
@@ -66,6 +66,9 @@ def get_secret_value(name: str) -> str:
     from mage_ai.orchestration.db.models import Secret
     fernet = Fernet(get_encryption_key())
 
-    secret = Secret.query.filter(Secret.name == name).one_or_none()
-    if secret:
-        return fernet.decrypt(secret.value.encode('utf-8')).decode('utf-8')
+    try:
+        secret = Secret.query.filter(Secret.name == name).one_or_none()
+        if secret:
+            return fernet.decrypt(secret.value.encode('utf-8')).decode('utf-8')
+    except Exception:
+        print(f'WARNING: Could not find secret value for secret {name}')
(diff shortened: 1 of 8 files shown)


mage_ai/data_preparation/templates/sensors/bigquery.py
```diff
@@ -0,0 +1,33 @@
...
+    Return: True if the sensor should complete, False if it should
+    keep waiting
+    """
+
+    config_path = path.join(get_repo_path(), 'io_config.yaml')
+    config_profile = 'default'
+
+    query = 'Your BigQuery query'  # Specify your SQL query here
+
+    with BigQuery.with_config(
```
wangxiaoyou1993, line 25: BigQuery doesn't support `with`.
Just use `BigQuery.with_config(ConfigFileLoader(config_path, config_profile)).load(query)`
https://docs.mage.ai/integrations/databases/BigQuery#using-python-block

mage_ai/data_preparation/shared/secrets.py
```diff
@@ -66,6 +66,9 @@ def get_secret_value(name: str) -> str:
...
 
-    secret = Secret.query.filter(Secret.name == name).one_or_none()
-    if secret:
-        return fernet.decrypt(secret.value.encode('utf-8')).decode('utf-8')
+    try:
+        secret = Secret.query.filter(Secret.name == name).one_or_none()
+        if secret:
+            return fernet.decrypt(secret.value.encode('utf-8')).decode('utf-8')
+    except Exception:
+        print(f'WARNING: Could not find secret value for secret {name}')
```
wangxiaoyou1993, line 74: Is it to fix something?

dy46: i was getting a weird error when trying to fetch secrets, and it was just causing the entire block to fail even though the block didn't necessary need the secret to run

wangxiaoyou1993 (approved):

PR #2224, opened 2023-03-17
[dy] Update models file

dy46:
# Summary

Refactored the `mage_ai.orchestration.db.models` file into a couple of separate files. This should help resolve some circular dependency issues with Secrets.

@wangxiaoyou1993 do you know if this will affect anything that I'm not aware of? I was thinking it could potentially affect the migrations, but I didn't see any config that was using the old `models.py` file.

# Tests

tested locally, but might need some further testing

cc: @wangxiaoyou1993


wangxiaoyou1993:
Migration finds the models by setting `target_metadata = Base.metadata`. So it should be fine.

wangxiaoyou1993 (approved):

PR #2248, opened 2023-03-22
[tc] update images for doc

thomaschung408:
# Summary

# Tests

cc:


tommydangerous:
What about the python robots? There were 2 of them.

Issue #2233, opened 2023-03-21
Opening http://localhost:6789/ automatically


Ayumi08:
**Is your feature request related to a problem? Please describe.**
It is a little annoying when starting up a new project, `http://localhost:6789/pipelines` does not automatically open in a browser tab.

**Describe the solution you'd like**
Let console open `http://localhost:6789/pipelines` automatically.

**Describe alternatives you've considered**
Only one solution to this??

**Additional context**
N/A

bruno-uy:
I can take this 🙋🏼‍♂️

tommydangerous:
Wow, you’re the best @bruno-uy!!! <3

bruno-uy:
If you're using `pip` or `conda` to run mage it's pretty straightforward. It gets tricky inside a docker container (recommended usage) because it needs to interact with the host's default browser. I'm trying different options without luck. I'll keep you updated.

bruno-uy:
I'm thinking about using the standard `webbrowser` Python library to open the app URL after the server is up (right before [this print](https://github.com/mage-ai/mage-ai/blob/e6d2714430b27ce560bf523dd3a6283b0b60abbc/mage_ai/server/server.py#L271)), but as I mention the `webbrowser` library cannot access the default browser because docker runs in isolation from the host machine.

If you want to access the host file system, you need to mount a volume, expose those folders, and have control of the window system to open the browser (as far as I understood reading some docs).

*TL;DR:* opening the host browser from the docker container is not the best practice, but we can include that when using `pip` or `conda`.

@Ayumi08 @tommydangerous let me know if I'm missing something and what do you think about my proposal.

tommydangerous:
I think that's totally acceptable solution. Thank you so much!!!

wangxiaoyou1993:
Hi @Ayumi08 , this feature is available in version 0.8.29. Let us know whether it works for you.

Ayumi08:
Working very well! Thank you so much!

tommydangerous:
🙌🙌🙌

PR #2255, opened 2023-03-23
Automatically open app in a new tab using default browser

bruno-uy:
# Summary
When the app starts listening in the defined `host:port`, a new tab on the default browser is opened.

Solves #2233. It doesn't work with Docker, and the explanation is in the issue's comments.

# Tests
Tested locally with MacOS/Chrome, but since it's a standard library from Python, it shouldn't have any issues with other OS/browser.

### DIFF
--- a/mage_ai/server/server.py
+++ b/mage_ai/server/server.py
@@ -269,7 +269,8 @@ async def main(
         address=host if host != 'localhost' else None,
     )
 
-    webbrowser.open_new_tab(url := f'http://{host or "localhost"}:{port}')
+    url = f'http://{host or "localhost"}:{port}'
+    webbrowser.open_new_tab(url)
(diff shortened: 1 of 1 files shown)


tommydangerous:
Looking 👀

bruno-uy:
The Walrus operator (`:=`) is supported starting in Python 3.8, and I realized mage also supports Python 3.7. @tommydangerous I have another commit ready to change that to a standard variable assignment. Let me know if we want 3.7 support, and I can push it.

wangxiaoyou1993:
Hey @bruno-uy , we do want to support 3.7.

wangxiaoyou1993 (approved):

wangxiaoyou1993:
Thanks you @bruno-uy for contributing to Mage and solve that github issue!

PR #2329, opened 2023-04-04
[jk] Cancel pipeline runs without associated pipelines

johnson-mage:
# Summary
- If a pipeline is deleted while a pipeline run is still running, users weren't able to cancel those pipeline runs in the UI, so this PR allows users to cancel pipeline runs even if their related pipeline was deleted.

# Tests
Before (cannot cancel running pipeline run):
Before - cancel pipeline run without pipeline

After (can cancel running run, but not retry because the pipeline no longer exists):
After - cancel pipeline run without pipeline

### DIFF
--- a/mage_ai/api/resources/PipelineRunResource.py
+++ b/mage_ai/api/resources/PipelineRunResource.py
@@ -221,13 +221,10 @@ class PipelineRunResource(DatabaseResource):
 
             return super().update(dict(status=PipelineRun.PipelineRunStatus.RUNNING))
         elif PipelineRun.PipelineRunStatus.CANCELLED == payload.get('status'):
-            pipeline = None
-            if os.path.exists(os.path.join(
-                get_repo_path(),
-                PIPELINES_FOLDER,
+            pipeline = Pipeline.get(
                 self.model.pipeline_uuid,
-            )):
-                pipeline = Pipeline.get(self.model.pipeline_uuid)
+                check_if_exists=True,
+            )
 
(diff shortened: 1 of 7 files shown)


johnson-mage:
@wangxiaoyou1993 Can you please check the backend changes when you get a chance? There wasn't a lot.

mage_ai/api/resources/PipelineRunResource.py
```diff
@@ -219,8 +221,14 @@ def update(self, payload, **kwargs):
 
             return super().update(dict(status=PipelineRun.PipelineRunStatus.RUNNING))
         elif PipelineRun.PipelineRunStatus.CANCELLED == payload.get('status'):
-            from mage_ai.orchestration.pipeline_scheduler import PipelineScheduler
-
-            PipelineScheduler(self.model).stop()
+            pipeline = None
+            if os.path.exists(os.path.join(
```
wangxiaoyou1993, line 225: Let's put the logic of check the existence of a Pipeline in `pipeline.py`

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -44,6 +44,40 @@
 MEMORY_USAGE_MAXIMUM = 0.95
 
 
+def stop_pipeline_run(
```
wangxiaoyou1993, line 47: You can put the method below PipelineScheduler class since other methods are below this class.

wangxiaoyou1993 (approved):

PR #2354, opened 2023-04-07
[xy] Add GCS destination to data integration pipeline

chethanuk:
# Summary

Adds support for Google Cloud Storage as a destination for exporting data. It includes a new class, GoogleCloudStorage, which inherits from the Destination base class and implements the necessary methods for exporting data to GCS.

The class includes methods to export data to GCS and supports the following configuration options:

- bucket: Name of the GCS bucket to save data in. Example: user-generated content.
- file_type: The type of GCS files. Supported file type values: parquet, csv. Supported Types: parquet or csv.
- object_key_path: The path of the location where you have files. Don’t include the gs://, the bucket name, or the table name in this path value. Example: users/ds/20221225.
- region: Region where the bucket is located. Default value is us-west1.
- service_account_json: Path to service account JSON file. Example: /path/to/service_account.json.

## Changes Made
- Added GoogleCloudStorage class to destinations/gcs.py
- Implemented export_batch_data and test_connection methods for GCS
- Added configuration options for GCS to config.json file
- Added unit tests for GoogleCloudStorage class to test_gcs.py
- Updated README.md file to include documentation for GCS configuration options

cc:

### DIFF
--- a/mage_integrations/mage_integrations/destinations/gcs/__init__.py
+++ b/mage_integrations/mage_integrations/destinations/gcs/__init__.py
@@ -1,0 +1,3 @@
+from mage_integrations.destinations.gcs import GoogleCloudStorage
+
+__all__ = ["GoogleCloudStorage"]
--- a/mage_integrations/mage_integrations/destinations/gcs/gcs.py
+++ b/mage_integrations/mage_integrations/destinations/gcs/gcs.py
@@ -1,0 +1,168 @@
+import os
+import logging
+from typing import Dict, List
+from datetime import datetime, timezone
+from io import BytesIO
+
+import pandas as pd
+from google.cloud import storage
+
+from mage_integrations.destinations.base import Destination
+from mage_integrations.destinations.utils import update_record_with_internal_columns
+
+
+class GoogleCloudStorage(Destination):
+    """
+    Export data to Google Cloud Storage.
+
+    Attributes:
+        bucket (str): The GCS bucket name to export data to.
+        file_type (str): The file type to export the data in, default is 'csv'.
+        object_key_path (str): The object key path in the bucket to export data to.
+        region (str): The GCS region to export data to, default is 'us-west1'.
+        client (google.cloud.storage.Client): The GCS client object used to interact with the API.
+    """    
+    @property
+    def bucket(self) -> str:
+        """
+        str: The GCS bucket name to export data to.
+        """
+        return self.config['bucket']
+
+    @property
+    def file_type(self) -> str:
+        """
+        str: The file type to export the data in, default is 'csv'.
+        """
+        return self.config.get('file_type', 'csv')
+
+    @property
+    def object_key_path(self) -> str:
+        """
+        str: The object key path in the bucket to export data to.
+        """
+        return self.config['object_key_path']
+
+    @property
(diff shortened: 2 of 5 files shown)


tommydangerous:
🙌🙌🙌

tommydangerous:
👀👀👀

mage_integrations/mage_integrations/destinations/gcs/__init__.py
```diff
@@ -0,0 +1,3 @@
+from mage_integrations.destinations.gcs import GoogleCloudStorage
+
+__all__ = ["GoogleCloudStorage"]
```
wangxiaoyou1993, line 3: nit: add an empty line at the end of file

mage_integrations/mage_integrations/destinations/gcs/gcs.py
```diff
@@ -0,0 +1,168 @@
...
+
+    @property
+    def object_key_path(self) -> str:
+        """
+        str: The object key path in the bucket to export data to.
+        """
+        return self.config['object_key_path']
+
+    @property
+    def region(self) -> str:
```
wangxiaoyou1993, line 47: Is region needed for GCS?

mage_integrations/mage_integrations/destinations/gcs/templates/config.json
```diff
@@ -0,0 +1,10 @@
+{
+  "aws_access_key_id": null,
```
wangxiaoyou1993, line 2: Could you match the config json with the actual config needed?

mage_integrations/mage_integrations/destinations/gcs/gcs.py
chethanuk: GCS location or region is optional, will remove it

mage_integrations/mage_integrations/destinations/gcs/templates/config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "bucket": null,
+  "file_type": "parquet",
+  "object_key_path": null,
+  "table": null,
+  "date_partition_format": null,
+  "project": "your-gcp-project-id"
```
wangxiaoyou1993, line 7: I think you need `service_account_json` here.

chethanuk: In this case, I assumed [Workload identity will solve it](https://cloud.google.com/kubernetes-engine/docs/how-to/hardening-your-cluster#use_least_privilege_sa), but they may be cases where users won't be using workload identity and may be best to use service_account_json or should we make service_account_json optional?

wangxiaoyou1993: I see. Right not, the client is initiated with the `service_account_json`:
```
    def __init__(self, **kwargs):
        """
        Initialize a new GoogleCloudStorage object.
        Args:
            **kwargs: Additional arguments to pass to the superclass constructor.
        """        
        super().__init__(**kwargs)
        self.client = storage.Client.from_service_account_json(self.config['service_account_json'])
```
If we make `service_account_json` optional, let's make sure the code handles that.

PR #2368, opened 2023-04-10
[hw] Add return type hints for some methods

csharplus:
# Summary

When executing a list of queries, the results of execution-only query do not need to be added to the final returned query results.

For some methods, missing return type hints are added back.

# Tests

CI
cc:

@dy46 @wangxiaoyou1993 @tommydangerous

### DIFF
--- a/mage_ai/io/bigquery.py
+++ b/mage_ai/io/bigquery.py
@@ -321,8 +321,7 @@ WHERE table_id = '{table_name}'
             if fetch_query_at_indexes and idx < len(fetch_query_at_indexes) and \
                     fetch_query_at_indexes[idx]:
                 result = result.to_dataframe()
-
-            results.append(result)
+                results.append(result)
 
         return results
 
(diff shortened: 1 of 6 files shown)


mage_ai/io/bigquery.py
```diff
@@ -321,8 +321,7 @@ def execute_queries(
             if fetch_query_at_indexes and idx < len(fetch_query_at_indexes) and \
                     fetch_query_at_indexes[idx]:
                 result = result.to_dataframe()
-
-            results.append(result)
+                results.append(result)
```
wangxiaoyou1993, line 324: What does it fix?

csharplus: > What does it fix?

It was adding all the results (not DataFrame) from the following execution queries:
```
result = self.client.query(query, **variables)
```
to the returned list. With this code change, it only added the DataFrame results from queries.

csharplus: Actually, when `fetch_query_at_indexes` is None, all the results need to be returned. I'll make a code change to handle this case.

csharplus:
@wangxiaoyou1993 please take a look and see if this PR is safe to merge, thanks.

mage_ai/io/bigquery.py
wangxiaoyou1993: I actually don't have context what it's fixing.

PR #2377, opened 2023-04-10
[dy] Fix dbt file delete error

dy46:
# Summary

Use the last index of `parts` to determine the extension, and use the first n-1 elements of `parts` to determine the block uuid. For a block named `block.py.sql`, the extension is `.sql`, and the block uuid would be `block.py`.

# Tests

tested locally

cc: @johnson-mage


dy46:
@johnson-mage updated it

johnson-mage (approved):

PR #2400, opened 2023-04-13
[dy] Add limit to sql blocks

dy46:
# Summary

Add default limit and limit override to sql blocks. Similar to DBT blocks, but DBT blocks won't have a default limit.

# Tests

tested locally

Screenshot 2023-04-13 at 4 18 48 PM

cc:


tommydangerous:
How does this affect a sql block passing data to a python block when running from a trigger? Will the limit be removed?

dy46:
@tommydangerous this only applies when running from the notebook

tommydangerous:
Sweet

PR #2421, opened 2023-04-17
[hw] Automatically run pip install of requirements.txt file whenever user git pulls or merge

csharplus:
# Summary

When there are new libraries added to `requirements.txt`, the existing setup could fail after a Git Pull, as the new libraries would not be automatically installed.

# Tests

CI
cc:

@dy46 @wangxiaoyou1993


mage_ai/data_preparation/git/__init__.py
```diff
@@ -133,44 +134,50 @@ def wrapper(self, *args, **kwargs):
...
+    def __set_git_config(self) -> None:
         if self.git_config.username:
             self.repo.config_writer().set_value(
                 'user', 'name', self.git_config.username).release()
         if self.git_config.email:
             self.repo.config_writer().set_value(
                 'user', 'email', self.git_config.email).release()
+
+    def __pip_install(self) -> None:
+        cmd = f'pip3 install -r {self.repo.working_dir}/requirements.txt'
```
wangxiaoyou1993, line 181: Let's make sure we only install the requirements.txt if it exists

csharplus: > Let's make sure we only install the requirements.txt if it exists

Good catch: made the suggested change.

mage_ai/data_preparation/git/__init__.py
```diff
@@ -118,59 +123,67 @@ def wrapper(self, *args, **kwargs):
...
 
         return wrapper
 
     @_remote_command
-    def reset(self, branch: str = None):
+    def reset(self, branch: str = None) -> None:
         self.origin.fetch()
         if branch is None:
             branch = self.current_branch
         self.repo.git.reset('--hard', f'{self.origin.name}/{branch}')
```
dy46, line 144: it might be helpful to add it here as well.

csharplus: > it might be helpful to add it here as well.

Good catch: added the suggested change.

dy46:
were you able to test this change in the UI? i wonder how much it would slow down the git actions

csharplus:
> were you able to test this change in the UI? i wonder how much it would slow down the git actions

@dy46 I followed the instructions given at https://docs.mage.ai/production/data-sync/git, and configured the SSH setup. However, when I clicked the one-time Git sync button on the Git setup UI page, I got the following error, then the local development server is down. Please see if you encountered similar issues before. I'll try it again to see if it works.
Screenshot 2023-04-19 at 8 01 51 PM

csharplus:
> > were you able to test this change in the UI? i wonder how much it would slow down the git actions
> 
> @dy46 I followed the instructions given at https://docs.mage.ai/production/data-sync/git, and configured the SSH setup. However, when I clicked the `Synchronize code` yellow button to perform one-time `Synchronize code from remote repository` on the `Git settings` UI page, I got the following error, then the local development server is down. Please see if you encountered similar issues before. I'll try it again to see if it works. Screenshot 2023-04-19 at 8 01 51 PM

This error was not present when the following checkbox was unchecked:
```
Sync before each trigger run
```

dy46 (commented):
cool, what happens if the pip command fails?

csharplus:
> cool, what happens if the pip command fails?

@dy46 Good point: should we throw an exception when this happens or we can wrap this `pip` command in a try-catch block to let it pass silently even it fails, or we need let it re-tries for more times then throw the exception when all trials fail?

dy46:
I think it may be ok to have it silently fail for now with some backend logging. It could be helpful to show an alert in the frontend eventually though.

csharplus:
> I think it may be ok to have it silently fail for now with some backend logging. It could be helpful to show an alert in the frontend eventually though.

Sure, I've made the suggested change in the backend code. We may need to add a backlog item to update frontend to show an alert when pip installation failed.

I just noticed that there is an additional UI component showing the current Git branch on the Toolbar after setting up Git configurations. I tried running `Pull` action directly from this UI component:

Screenshot 2023-04-20 at 10 16 05 PM

It worked well and printed out the following messages, though the whole pip installation took around 13 seconds to completed even without any new Python libraries added to `requirements.txt`.

```
mage-ai-server-1  | WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
mage-ai-server-1  |  Running "pip3 install -r /home/src/requirements.txt"...Installing /home/src/requirements.txt completed successfully.
mage-ai-server-1  |  DONE
mage-ai-server-1  | INFO:mage_ai.api.logging:[2023-04-21T05:16:22.341728][api.views] Action: update git_branches csharplus_add_pip  
mage-ai-server-1  | INFO:mage_ai.api.logging:[2023-04-21T05:16:22.341895][api.views] Latency: 13.9548 seconds
```

@dy46 @wangxiaoyou1993 please review the updated code and see if any additional testing is needed for this change.

dy46:
@wangxiaoyou1993 do you know if that warning is ok?

wangxiaoyou1993:
> @wangxiaoyou1993 do you know if that warning is ok?

that warning is ok

PR #2432, opened 2023-04-17
[dy] Update logging

dy46:
# Summary

Update logging parsing logic as it was excluding some logs. I updated the logging in the block executor to be more consistent as well.

# Tests

tested locally

cc: @johnson-mage can you take a look? and see if it conflicts with your log changes?


johnson-mage (approved):
This shouldn't conflict with my changes. Thanks

PR #2455, opened 2023-04-19
[jk] Do not switch kernels when deleting pipeline

johnson-mage:
# Summary
- Error was occurring when trying to delete pyspark pipelines due to an attempt to switch kernels. For DELETE actions, we do not need to switch kernels.

# Tests
Before (error when trying to delete):
image

After (no issues deleting same pipeline):
delete spark pipeline

### DIFF
--- a/mage_ai/api/operations/base.py
+++ b/mage_ai/api/operations/base.py
@@ -171,7 +171,7 @@ class BaseOperation():
     async def __delete_show_or_update(self):
         updated_options = await self.__updated_options()
         res = await self.__resource_class().process_member(
-            self.pk, self.user, **updated_options)
+            self.pk, self.user, action=self.action, **updated_options)
 
         policy = self.__policy_class()(res, self.user, **updated_options)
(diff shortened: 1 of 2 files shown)


mage_ai/api/operations/base.py
```diff
@@ -171,7 +171,7 @@ async def __create_or_index(self):
     async def __delete_show_or_update(self):
         updated_options = await self.__updated_options()
         res = await self.__resource_class().process_member(
-            self.pk, self.user, **updated_options)
+            self.pk, self.user, action=self.action, **updated_options)
```
wangxiaoyou1993, line 174: Could you check whether you can use `api_operation_action` in the udpated_options? https://github.com/mage-ai/mage-ai/blob/master/mage_ai/api/operations/base.py#L298

johnson-mage: Yes, that kwarg does work. Updated.

wangxiaoyou1993 (approved):

PR #2481, opened 2023-04-21
create method for deleting secret keys

hjhdaniel:
# Summary
1) reason/use-case: There may be cases where there is a need to remove or update a secret value
2) example: Xero API has a rolling, single-use, offline refresh token that has to be updated whenever it is expended for retrieving access token. To keep a long-term automated pipeline connected to Xero API, the data loader has to consistently refresh token secret.
3) Notes:
    a) I exposed `delete_secret` instead of `update_secret` as it is reusable in cases where a secret simply needs to be deleted
    b) Not sure if there is any safety/security concern here that I am not considering, so do let me know if this feature is acceptable

# Tests
1) Made the changes and ran `./scripts/dev.sh [PROJECT_NAME]` for local testing
2) Created dummy secrets using the FE
3) Ran `delete_secret` function with a dummy loader block
4) Edge cases tested:
    a) When there are multiple secrets, it only deletes the desired one
    b) When there are no secrets with the key, no exception is thrown, only a soft warning is printed
    c) After the secret is deleted, restart the app to ensure that the deletion changes persists (checking actual physical deletion by using `get_secret_value`)
    d) Each secret can be deleted and recreated repeatedly without conflict


mage_ai/data_preparation/shared/secrets.py
```diff
@@ -1,10 +1,8 @@
...
-from mage_ai.data_preparation.repo_manager import (
-    get_data_dir,
-    get_repo_path
-)
-from typing import List
 import os
+from typing import List
+
+from cryptography.fernet import Fernet, InvalidToken
+from mage_ai.data_preparation.repo_manager import get_data_dir, get_repo_path
```
dy46, line 5: we organize the imports with the `from` imports first

mage_ai/data_preparation/shared/secrets.py
```diff
@@ -85,3 +83,15 @@ def get_secret_value(name: str) -> str:
...
+    from mage_ai.orchestration.db.models.secrets import Secret
+
+    secret = None
+    try:
+        secret = Secret.query.filter(Secret.name == name).one_or_none()
+    except Exception:
+        print(f'WARNING: Secret {name} does not exist')
+
+    if secret:
+        secret.delete()
```
dy46, line 97: can you add a new line as well?

mage_ai/data_preparation/shared/secrets.py
hjhdaniel: sorry accidentally committed code changes from auto-formatting. Have reverted the formatting

mage_ai/data_preparation/shared/secrets.py
hjhdaniel: new line added!

hjhdaniel:
> LGTM after fixing lint CI issues

hi @dy46, I am not sure why the lint test is failing but I have ensured to maintain the formatting consistent with the rest of the code in the file. Is there anything else I need to fix before this can be merged?

hjhdaniel:
> 

right, sorry i missed it! Have formatted it, thanks!

PR #2496, opened 2023-04-24
[jk] Bulk retry pipeline runs

johnson-mage:
# Summary
- Allow all pipeline runs on a page to be selected and retried.
- Bulk retry runs not enabled for streaming pipelines.

# Tests
Standard pipeline:
bulk retry standard pipeline runs

Integration pipeline:
bulk retry integration pipeline runs

cc:
@tommydangerous

### DIFF
--- a/mage_ai/orchestration/db/models/schedules.py
+++ b/mage_ai/orchestration/db/models/schedules.py
@@ -5,6 +5,7 @@ from mage_ai.data_preparation.models.block.utils import (
     get_all_ancestors,
     is_dynamic_block,
 )
+from mage_ai.data_preparation.models.constants import PipelineType
 from mage_ai.data_preparation.models.pipeline import Pipeline
 from mage_ai.data_preparation.models.triggers import (
     ScheduleInterval,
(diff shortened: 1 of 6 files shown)


dy46 (approved):

mage_ai/orchestration/db/models/schedules.py
```diff
@@ -308,6 +309,38 @@ def in_progress_runs(
             PipelineRun.passed_sla.is_(False),
         ).all()
 
+    @safe_db_query
+    def retry_pipeline_run(
```
wangxiaoyou1993, line 313: I would put this method in `pipeline_scheduler.py` instead of this file.

mage_ai/orchestration/db/models/schedules.py
```diff
@@ -308,6 +309,38 @@ def in_progress_runs(
...
+        from mage_ai.data_integrations.utils.scheduler import initialize_state_and_runs
+
+        is_integration = PipelineType.INTEGRATION == self.pipeline.type
+        schedule = PipelineSchedule.get(pipeline_run['pipeline_schedule_id'])
+        execution_date = datetime.fromisoformat(pipeline_run['execution_date'])
+        pipeline_run = self.create(
+            create_block_runs=False,
+            execution_date=execution_date,
+            pipeline_schedule_id=schedule.id,
+            pipeline_uuid=self.pipeline_uuid,
```
wangxiaoyou1993, line 327: I would check whether the pipeline_uuid is a valid pipeline.

mage_ai/orchestration/db/models/schedules.py
```diff
@@ -308,6 +309,38 @@ def in_progress_runs(
             PipelineRun.passed_sla.is_(False),
         ).all()
 
+    @safe_db_query
+    def retry_pipeline_run(
+        self,
+        pipeline_run: dict,
```
wangxiaoyou1993, line 315: Let's use `Dict` type here, which is imported by `from typing import Dict`
I remember `dict` type is only supported in python 3.9 and newer versions.

mage_ai/orchestration/db/models/schedules.py
wangxiaoyou1993: Also, this method has duplicate with this method: https://github.com/mage-ai/mage-ai/blob/master/mage_ai/api/resources/PipelineRunResource.py#L147-L183
We should try reducing the duplicate code.

johnson-mage: comments addressed in https://github.com/mage-ai/mage-ai/pull/2512

PR #2500, opened 2023-04-24
[dy] Remove useEffect

dy46:
# Summary

Remove useEffect and add to `useState` default.

# Tests

tested locally

cc: @tommydangerous

### DIFF
--- a/mage_ai/frontend/components/CodeBlock/index.tsx
+++ b/mage_ai/frontend/components/CodeBlock/index.tsx
@@ -240,6 +240,16 @@ function CodeBlock({
   const [content, setContent] = useState(defaultValue);
   const [currentTime, setCurrentTime] = useState<number>(null);
 
+  const isStreamingPipeline = PipelineTypeEnum.STREAMING === pipeline?.type;
+  const isDBT = BlockTypeEnum.DBT === block?.type;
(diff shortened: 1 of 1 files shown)


mage_ai/frontend/components/CodeBlock/index.tsx
```diff
@@ -240,6 +240,16 @@ function CodeBlock({
   const [content, setContent] = useState(defaultValue);
   const [currentTime, setCurrentTime] = useState<number>(null);
 
+  const isStreamingPipeline = PipelineTypeEnum.STREAMING === pipeline?.type;
+  const isDBT = BlockTypeEnum.DBT === block?.type;
+  const isSQLBlock = BlockLanguageEnum.SQL === block?.language;
+  const isRBlock = BlockLanguageEnum.R === block?.language;
+
+  let defaultLimitValue = blockConfiguration[CONFIG_KEY_LIMIT];
+  if (!isDBT && isSQLBlock && defaultLimitValue == null) {
```
tommydangerous, line 249: Shouldn’t we check for undefined?

dy46: it checks for both null and undefined, but changed it to just undefined

PR #2560, opened 2023-05-02
[dy] Use load jobs for bigquery

dy46:
# Summary

Add support for batch jobs in bigquery. We take the input records, convert them to a JSON format, and write them to a file. The batch load job reads from the file, and uploads the JSON records to BigQuery.

I added a config field in the bigquery config called `use_batch_load` that can be toggled to enable the batch load functionality.

# Tests

tested locally

cc: @wangxiaoyou1993


wangxiaoyou1993:
What data types did you test?
Could you make sure the following data types are tested at least?
* integer
* float
* string
* datetime
* object
* array
* array of object
* array of integer
* array of string

dy46:
@wangxiaoyou1993 tested those types

PR #2591, opened 2023-05-05
[jk] Retry block runs from specific block

johnson-mage:
# Summary
- For standard python pipelines, retry block runs from a selected block. The selected block and all downstream blocks will be re-ran after clicking the `Retry from selected block` button.
- This feature is not available for other types of pipelines (e.g. streaming or integration). Streaming pipelines don't have block runs, and users can still retry from failed block runs in integration pipelines.

# Tests
retry from selected block

### DIFF
--- a/mage_ai/api/resources/PipelineRunResource.py
+++ b/mage_ai/api/resources/PipelineRunResource.py
@@ -166,23 +166,33 @@ class PipelineRunResource(DatabaseResource):
 
     @safe_db_query
     def update(self, payload, **kwargs):
-        if 'retry_blocks' == payload.get('pipeline_run_action') and \
-                PipelineRun.PipelineRunStatus.COMPLETED != self.model.status:
+        if 'retry_blocks' == payload.get('pipeline_run_action'):
             self.model.refresh()
-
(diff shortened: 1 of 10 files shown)


mage_ai/api/resources/PipelineRunResource.py
```diff
@@ -166,23 +166,35 @@ def _create_callback(resource):
...
             pipeline = Pipeline.get(self.model.pipeline_uuid)
-
-            incomplete_block_runs = \
-                list(
-                    filter(
-                        lambda br: br.status != BlockRun.BlockRunStatus.COMPLETED,
-                        self.model.block_runs
+            block_runs_to_retry = []
+            from_block_uuid = payload.get('from_block_uuid')
+            if PipelineRun.PipelineRunStatus.COMPLETED != self.model.status:
```
johnson-mage, line 174: This `if block` coming first is intentional. If a block had failed and a user tries to re-run from a specific block downstream to the failed block, we need to rerun from the failed block.

johnson-mage:
@dy46 Can you review the BE changes? There's only a couple files.

dy46 (approved):

PR #2599, opened 2023-05-05
[xh] Add DynamoDB into source

matrixstone:
# Summary

# Tests

### Tested with Mage UI
(1) Add DynamoDB type
Dynamo DB Type

(2) It loads corresponding config
Shows config

(3) Test connection success
Test connection success

(4) Select streams to sync
select streams to sync

(4) Able to view and select streams:
View and select stream

(5) Able to load sample data:
Load sample data

(6) Run the pipeline with destination of S3 succeed:

Run trigger completes

Destination in S3

### Tested with command
```python3 mage_integrations/sources/dynamodb/__init__.py \
  --config mage_integrations/TEST_CONFIG1.json \
  --selected_streams '["feature_sets_development"]' \
  --load_sample_data
```

Data loaded

### Tested bookmark

After set the source pipeline as INCREMENTAL, there is 60 rows in the table, the latest sync only updates 36 rows:
success sync with incremental

UI shows bookmark property:
Manually edit bookmark

cc:
@wangxiaoyou1993


mage_integrations/mage_integrations/sources/dynamodb/README.md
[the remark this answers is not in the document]
matrixstone: Sounds good! Added content into README

wangxiaoyou1993:
@matrixstone did you test loading sample data via UI?

mage_integrations/mage_integrations/sources/dynamodb/__init__.py
```diff
@@ -0,0 +1,135 @@
...
+        for result in self.scan_table(table_name):
+            for item in result.get('Items', []):
+                record_message = common.row_to_singer_record(
+                        stream.to_dict(),
+                        item,
+                        None,
+                        None,
+                    )
+                data.append(record_message.record)
+        yield data
```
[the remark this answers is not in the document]
matrixstone: Sounds good. I updated it to yield every time complete record one result. The size of result is 1000 records controlled in the scan_table function.

mage_integrations/mage_integrations/sources/dynamodb/__init__.py
```diff
@@ -0,0 +1,135 @@
+from botocore.config import Config
+from botocore.exceptions import ClientError
+from datetime import datetime
+from mage_integrations.sources.base import Source, main
+from mage_integrations.sources.catalog import Catalog, CatalogEntry
+from mage_integrations.sources.utils import get_standard_metadata
+from typing import Dict, Generator, List
+import boto3
+import mage_integrations.sources.mongodb.tap_mongodb.sync_strategies.common as common
```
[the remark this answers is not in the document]
matrixstone: Make sense. I notice there are multiple scripts calling the common script. I leave a TODO here. I can refactor the common function to a util class in a different PR afterwords.

mage_integrations/mage_integrations/sources/dynamodb/__init__.py
```diff
@@ -0,0 +1,145 @@
...
+            # Check if streams available
+            for stream in streams:
+                catalog_entry = self.discover_stream(client, stream)
+                if catalog_entry is not None:
+                    outputs.append(catalog_entry)
+        else:
+            for table_config in self.table_configs:
+                catalog_entry = self.discover_stream(client, table_config['table_name'])
+                if catalog_entry is not None:
+                    outputs.append(catalog_entry)
```
dy46, line 73: i don't think we need this. The user will select the streams they want to fetch data from.

mage_integrations/mage_integrations/sources/dynamodb/__init__.py
```diff
@@ -0,0 +1,135 @@
...
+        return self.config.get('aws_region', 'us-west-2')
+    
+    @property
+    def table_configs(self):
+        """
+        Used for multiple streams. Each table config has the key of table_name.
+        """
+        configs = self.config.get('table_configs', [])
+        configs = [c for c in configs if all(k in c for k in REQUIRED_TABLE_CONFIG_KEYS)]
+        return configs
```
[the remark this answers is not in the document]
matrixstone: Done

mage_integrations/mage_integrations/sources/dynamodb/__init__.py
matrixstone: Done

mage_integrations/mage_integrations/sources/dynamodb/__init__.py
```diff
@@ -0,0 +1,182 @@
...
+        for column_schema in table_info.get(ATTRIBUTE_DEFINITIONS, []):
+            column_name = column_schema.get(ATTRIBUTE_NAME)
+            column_type = column_schema.get(ATTRIBUTE_TYPE)
+            column_types = []
+            if column_type == 'S':
+                column_types.append(COLUMN_TYPE_STRING)
+            elif column_type == 'N':
+                column_types.append(COLUMN_TYPE_NUMBER)
+            elif column_type == 'B':
+                column_types.append(COLUMN_TYPE_OBJECT)
```
dy46, line 75: are these the only data types in dynamodb?

mage_integrations/mage_integrations/sources/dynamodb/__init__.py
```diff
@@ -0,0 +1,182 @@
...
+
+            if result.get(LAST_EVALUATED_KEY):
+                scan_params[EXCLUSIVE_STAR_KEY] = result[LAST_EVALUATED_KEY]
+
+            has_more = result.get(LAST_EVALUATED_KEY, False)
+
+    def load_data(
+        self,
+        stream,
+        bookmarks: Dict = None,
```
dy46, line 161: we should probably take the bookmarks into account. you can look at other sources to see how bookmarks work

mage_integrations/mage_integrations/sources/dynamodb/__init__.py
matrixstone: One note, 'B' here actually stands for Binary. I created a new type constant.

The `AttributeType ` from describe_table only have 3 types: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_AttributeDefinition.html.

matrixstone: However when I look into what are the data types DynamoDB supports in general, it listed 10 types here: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html. The different between those two is weird.

mage_integrations/mage_integrations/sources/dynamodb/__init__.py
matrixstone: Added the logic with bookmark.

So if `bookmarks` contains the `LAST_EVALUATED_KEY`, we will scan the table beginning with `LAST_EVALUATED_KEY`. Meanwhile, anytime we get result from scan_table, we record the LAST_EVALUATED_KEY as the beginning for next round.

dy46: @matrixstone the bookmarks here will look something like based on the `state` of the source.
```
bookmarks: {
  bookmark_property_1: xxx,
  bookmark_property_2: xxx,
}
```

also, users can select what their bookmark properties are, so we'll want to filter by that as well.

overall, the bookmarking logic makes sense, but we just need to fix how we read from the bookmarks parameter, and add some logic for the bookmark properties. here's the sql one for reference: https://github.com/mage-ai/mage-ai/blob/master/mage_integrations/mage_integrations/sources/sql/base.py#L328

matrixstone: Updated the PR to include two functions `_get_bookmark_properties_for_stream` and `__get_bookmarks_for_stream` in `dynamodb/__init__.py`.  Also when reading from bookmark in `load_data`, we will check if it is in `bookmark_properties`

matrixstone: Updated the bookmark logic.

Once I set the `valid_replication_keys` with `key properties`, the `bookmarks` dict contains those keys and its values. At the same time, the format of bookmarks is same as the `last evaluated key` from dynamodb, so we can use it to query incremental data directly.

I tested it with Mage UI. Attached the screenshot in the description in this PR.

matrixstone:
@wangxiaoyou1993 @dy46 Updated the PR with bookmark handling and tested it with Mage UI. I attached the screenshot in description section. PTAL!

dy46 (approved):

PR #2614, opened 2023-05-09
[jk] Retry from selected block in integration pipeline

johnson-mage:
# Summary
- If a block is selected in an integration pipeline to retry block runs, only the block runs for the selected block's stream will be ran (e.g. if there are multiple streams).

# Tests
retry from block - integration pipeline

### DIFF
--- a/mage_ai/data_preparation/models/pipelines/integration_pipeline.py
+++ b/mage_ai/data_preparation/models/pipelines/integration_pipeline.py
@@ -40,10 +40,6 @@ class IntegrationPipeline(Pipeline):
         return find(lambda x: BlockType.DATA_EXPORTER == x.type, self.blocks)
 
     @property
-    def transformer(self) -> Block:
-        return find(lambda x: BlockType.TRANSFORMER == x.type, self.blocks)
-
-    @property
     def destination_config(self) -> Dict:
         if self.data_exporter and self.data_exporter.content:
             return yaml.safe_load(self.data_exporter.content)
--- a/mage_ai/data_preparation/models/pipelines/integration_pipeline.py
+++ b/mage_ai/data_preparation/models/pipelines/integration_pipeline.py
@@ -122,6 +122,12 @@ class IntegrationPipeline(Pipeline):
         transformer_file = importlib.import_module('mage_integrations.transformers.base')
         return os.path.abspath(transformer_file.__file__)
 
+    def block_from_block_uuid_with_stream(
+        self,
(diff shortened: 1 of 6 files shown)


mage_ai/data_preparation/models/pipelines/integration_pipeline.py
```diff
@@ -39,6 +39,10 @@ def data_loader(self) -> Block:
     def data_exporter(self) -> Block:
         return find(lambda x: BlockType.DATA_EXPORTER == x.type, self.blocks)
 
+    @property
+    def transformer(self) -> Block:
+        return find(lambda x: BlockType.TRANSFORMER == x.type, self.blocks)
+
```
dy46, line 45: i remember xiaoyou mentioning that some users wanted to have multiple transformers in a data integration pipeline. not sure if that's being prioritized but it might break this in the future

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -573,6 +574,26 @@ def run_integration_pipeline(
...
+            and one would indicate starting from the data exporter block.
+            """
+            if len(block_runs_for_stream) == 2 and \
+                transformer_block and \
+                find(
+                    lambda br: transformer_block.uuid in br.block_uuid,
+                    block_runs_for_stream):
+                starting_block = transformer_block
+            elif len(block_runs_for_stream) == 1 and \
+                    data_exporter_block.uuid in block_runs_for_stream[0].block_uuid:
```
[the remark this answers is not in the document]
johnson-mage: This was intentionally `in` cuz the data exporter block run `block_uuid` has the format f'{data_exporter_block.uuid}:{tap_stream_id}:{idx}' so it's just checking for the first part without the stream and idx. It is assuming if there's only 1 executable block id, it's a data exporter block (further up the pipeline would have to have more), so this is just another check to make sure.

mage_ai/shared/array.py
```diff
@@ -43,3 +44,8 @@ def unique_by(arr1, key):
         arr2.append(item)
         mapping[k] = True
     return arr2
+
+
+def is_equal(li1, li2):
```
dy46, line 49: can we remove this?

dy46 (approved):

mage_ai/shared/array.py
johnson-mage: yes, will remove

PR #2355, opened 2023-04-07
Adding Pre-commit configuration for Black and Flake8

chethanuk:
# Pre-commit, Black, and Flake8 Updates

## Summary

Adding pre-commit hooks for Black and Flake8, as well as updating the project configuration files for Flake8 to ensure consistent code style and catch common errors before committing.

## Changes

- Added pre-commit hooks for Black and Flake8 to the Maze project.
- Updated the Flake8 configuration file to ignore certain directories and files.
- Fixed various Flake8 errors and warnings found during the update process.

# Tests

The updated pre-commit hooks were tested by running `pre-commit run --all-files`
The Flake8 configuration changes were also tested by running flake8 on the repository both before and after the updates were applied.

cc:

PR #2616, opened 2023-05-09
Add pre-commit and improve Contributor Friendliness

### LINKED ISSUE #2256: Add pre-commit to the project to prevent failures in lint test
As a mage contributor, I would like a more straightforward way to check my commits against the "standard". Simple things like `flake8`, but other things like "we use single quotes for stings". When I created my first PR the `Build and test / Lint (pull_request)` test failed because I had spaces in an empty line. It was an easy fix, but I had to take time to correct it when having something checking for that before pushing would be better.

I would recommend adding [pre-commit](https://pre-commit.com/) and setting the rules we're currently trying to follow.

You can configure that on your own IDE, but it's not "mandatory". You see the failures when the commit is already pushed.

jcorrado76:
This PR closes #2256.

There is already the PR https://github.com/mage-ai/mage-ai/pull/2355.

I wanted to start the pre-commit configuration with the simplest linting first, and gradually add more. Using Black and flake8 are both rather invasive, and would be difficult to validate in addition to the other pre-commit hooks. 

For this PR, only some vanilla hooks from pre-commit, and isort are included. 

Since there is already a `.flake8` configuration file, we can probably add that one in as well without too many issues. 

I wanted to create this draft PR to get feedback on some of the developer friendliness changes included. 

In particular:
* added a Makefile in the root of the repo to help with dev environment setup - right now it just copies a single git hook over into the `.git/hooks` directory to facilitate including a reference to the GitHub issue you're working on
* added dev dependencies into the `pyproject.toml` so that contributors can use poetry locally to create a development environment (the build-backend was left as is as setuptools)
* added a developer-specific README to include documentation on how to setup the development virtual environment
* add isort configuration that sorts the imports in all Python modules, and labels sections according to first party, third party and standard library imports


jcorrado76:
@wangxiaoyou1993, if you or someone on the team could take a preliminary look and give me some feedback, that'd be awesome! Let me know if there are changes I should remove, rename, or move somewhere else. Thanks!

wangxiaoyou1993:
@jcorrado76 Thanks for putting up this PR! The PR looks good overall.

Here are some feedbacks:
* I imagine there'll be check failures from `isort` plugin. But I think it's good to have that.
* Besides conda virtual environment, we use the `virtualenv` more.
* It'll be helpful is to link the `README_dev.md` in this doc: https://github.com/mage-ai/mage-ai/blob/master/docs/community/contributing.mdx?plain=1#L38

.pre-commit-config.yaml
```diff
@@ -0,0 +1,46 @@
...
+    name: isort
+    entry: isort
+    # assumes isort is installed in local virtual environment as dev dependency
+    language: system
+    types: [python]
+  - id: flake8
+    name: flake8
+    entry: flake8
+    # assumes flake8 is installed in local virtual environment as dev dependency
+    language: system
```
wangxiaoyou1993, line 27: Let's add `args: ["--profile", "black"]` here as well

pyproject.toml
```diff
@@ -1,3 +1,38 @@
...
+isort = "*"
+commitizen = "*"
+ipykernel = "^6.17.1"
+pandas-stubs = "*"
+flake8-bugbear = "*"
+sourcery = "*"
+
+[tool.isort]
+profile = "black"
+import_heading_stdlib = "Python Standard Library Imports"
```
wangxiaoyou1993, line 28: we don't need the heading.

.pre-commit-config.yaml
jcorrado76: Hi, I have:
```toml
[tool.isort]
profile = "black"
```

In the `pyproject.toml`, so `isort` picks this up. I think I would prefer to keep the configuration for the pre-commit/linting tooling in one place as much as possible (outside of flake8, since it doesn't support `pyproject.toml` yet), and so think we should avoid setting the profile in both the `pyproject.toml`, and the `.pre-commit-config.yaml` file explicitly. 

Do you still want me to add the `args` for isort as described in your comment? 
Thanks

pyproject.toml
jcorrado76: Okay, I'll remove the `isort` headings for imports.

jcorrado76:
Hi @wangxiaoyou1993,

* there will definitely be check failures from `isort`. In my experience, it's kind of annoying to just add isort into the CI/CD all at once as it'll always fail, and try to fix all the python files in the repo if you let it. Do you have any opinions about how to roll this out? I was thinking we could at first just leave it as a pre-commit hook, and actually add it into CI/CD as an actual check at a later date. In other words, we start with it being optional, and then as contributors add new changes, they will run pre-commit to make sure new contributions are isort'd. Then, at some point, we add it into the formal CI/CD pipeline, and run it on the full codebase (I'm not sure how you feel about having massive commits editing 500+ files just to isort every python file)
* I will add documentation about how to use virtualenv as well
* I will link the README_dev.md in the https://github.com/mage-ai/mage-ai/blob/master/docs/community/contributing.mdx?plain=1#L38

wangxiaoyou1993:
> Hi @wangxiaoyou1993,
> 
> * there will definitely be check failures from `isort`. In my experience, it's kind of annoying to just add isort into the CI/CD all at once as it'll always fail, and try to fix all the python files in the repo if you let it. Do you have any opinions about how to roll this out? I was thinking we could at first just leave it as a pre-commit hook, and actually add it into CI/CD as an actual check at a later date. In other words, we start with it being optional, and then as contributors add new changes, they will run pre-commit to make sure new contributions are isort'd. Then, at some point, we add it into the formal CI/CD pipeline, and run it on the full codebase (I'm not sure how you feel about having massive commits editing 500+ files just to isort every python file)
> * I will add documentation about how to use virtualenv as well
> * I will link the README_dev.md in the https://github.com/mage-ai/mage-ai/blob/master/docs/community/contributing.mdx?plain=1#L38

Yeah, i agree. We can make isort optional and applied before commit for now.

.pre-commit-config.yaml
wangxiaoyou1993: ok. no need to add the `args` then.

wangxiaoyou1993 (approved):

wangxiaoyou1993:
Can we skip certain files in pre-commit? "mage_ai/data_preparation/templates, mage_ai/orchestration/db/migrations/versions"

jcorrado76:
Can you provide more information about how you're running pre-commit when you want to skip files? 
By default, when you let pre-commit run automatically as a hook, it only runs on files that have diffs. 
In addition, for each of the hooks in the `.pre-commit-config.yaml` file, I've added the `types: [python]` attribute, which means that it'll only check `.py` files. 

Which of the hooks was running on the files in `mage_ai/data_preparation/templates`? 
And were you running it using the `--files` flag on the root of the repo? 

Thanks

wangxiaoyou1993:
> Hi @wangxiaoyou1993 Can you provide more information about how you're running pre-commit when you want to skip files? By default, when you let pre-commit run automatically as a hook, it only runs on files that have diffs. In addition, for each of the hooks in the `.pre-commit-config.yaml` file, I've added the `types: [python]` attribute, which means that it'll only check `.py` files.
> 
> Which of the hooks was running on the files in `mage_ai/data_preparation/templates`? And were you running it using the `--files` flag on the root of the repo?
> 
> Thanks

We do have some files in `mage_ai/data_preparation/templates` that use `.py` as file extension https://github.com/mage-ai/mage-ai/tree/master/mage_ai/data_preparation/templates/data_loaders. But they have jinja template syntax in it, so if we make change to the file, it won't pass the some checks.

jcorrado76:
Hi @wangxiaoyou1993 

I've added a global exclude in the `.pre-commit-config.yaml` file. It now excludes the two directories you've mentioned. 
So even if you run:
```bash
pre-commit run --all-files
```

It won't process files in the `mage_ai/data_preparation/templates` and `mage_ai/orchestration/db/migrations/versions` directories. 

Let me know if this PR is good to merge in, or if there are other things to fix. 
Thanks

wangxiaoyou1993:
> Hi @wangxiaoyou1993
> 
> I've added a global exclude in the `.pre-commit-config.yaml` file. It now excludes the two directories you've mentioned. So even if you run:
> 
> ```shell
> pre-commit run --all-files
> ```
> 
> It won't process files in the `mage_ai/data_preparation/templates` and `mage_ai/orchestration/db/migrations/versions` directories.
> 
> Let me know if this PR is good to merge in, or if there are other things to fix. Thanks

I think your PR looks good now. I'll do some testing with it. If there're no issues, I'll merge it.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,52 @@
...
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+    -   id: trailing-whitespace
+        types: [python]
+    -   id: end-of-file-fixer
+        types: [python]
+    -   id: check-yaml
+        types: [python]
```
wangxiaoyou1993, line 13: check-yaml should be applied to yaml, right?

jcorrado76: I switched it from python to YAML.

wangxiaoyou1993 (approved):

wangxiaoyou1993 (approved):

PR #2637, opened 2023-05-13
[jk] Display error popup with link to docs for too many open files err

johnson-mage:
# Summary
- If `Too many open files` occurs while fetching pipelines, error popup with link to docs on how to resolve will appear on the Pipelines Dashboard.

# Tests
too many open files err

### DIFF
--- a/mage_ai/api/resources/PipelineResource.py
+++ b/mage_ai/api/resources/PipelineResource.py
@@ -46,8 +46,7 @@ class PipelineResource(BaseResource):
             try:
                 return await Pipeline.get_async(uuid)
             except Exception as err:
-                print(f'Error loading pipeline {uuid}: {err}.')
-                return None
+                raise Exception(f'Error loading pipeline {uuid}: {err}.')
 
         pipelines = await asyncio.gather(
             *[get_pipeline(uuid) for uuid in pipeline_uuids]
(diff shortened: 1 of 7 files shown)


mage_ai/api/resources/PipelineResource.py
```diff
@@ -46,8 +46,7 @@ async def get_pipeline(uuid):
             try:
                 return await Pipeline.get_async(uuid)
             except Exception as err:
-                print(f'Error loading pipeline {uuid}: {err}.')
-                return None
+                raise Exception(f'Error loading pipeline {uuid}: {err}.')
```
johnson-mage, line 49: @wangxiaoyou1993 This BE change seemed necessary cuz if we don't explicitly raise the exception, a portion of the pipelines could still be fetched and return a successful response (which wouldn't cause the error popup to appear in the UI).

wangxiaoyou1993: Oh, I don't want this exception to be raised. There could be other exceptions, like invalid pipeline yaml file. In that case, we should still be able to show other valid pipelines.

johnson-mage: @wangxiaoyou1993 Does [this](https://github.com/mage-ai/mage-ai/pull/2637/files#diff-eafd460a9a56974cc78079ca50178202b942f455f8977d22d78a7a32f3fa7cddR49-R54) work? I added a condition to check specifically for the `Too many open files` error and raise an exception for that, but not other errors.

wangxiaoyou1993: Yes, I think it works

wangxiaoyou1993 (approved):

PR #2654, opened 2023-05-17
[dy] Update user authentication

dy46:
# Summary

Update user authentication to user updated roles, permissions, and entity access. Design doc here: https://www.notion.so/mageai/Multi-Workspace-117673bab31f43ac90f49510b2939aba.

I'm still testing the PR, and may make some additional changes, but wanted to get review of the general logic first. For each Policy class, I added an `entity` property that will be used to determine what access the user has for that Resource. For example, the `PipelinePolicy` class will have entity = (PIPELINE, model.id), if the user is trying to access a pipeline, they will need to at least have access to that pipeline. 

This PR won't add support Pipeline level permissions yet, it's just to add the new models. Most of the new backend user authentcation logic is in the files `api/utils.py`, `models/oauth.py`, and `server/server.py`.

# Tests

still testing...

cc: @wangxiaoyou1993


mage_ai/orchestration/db/models/oauth.py
```diff
@@ -31,6 +35,7 @@ class User(BaseModel):
 
     oauth2_applications = relationship('Oauth2Application', back_populates='user')
     oauth2_access_tokens = relationship('Oauth2AccessToken', back_populates='user')
+    roles_new = relationship('Role', secondary='user_role', back_populates='users')
```
wangxiaoyou1993, line 38: Could you add comment for the roles and roles_new? Would user_roles be a better name here?

mage_ai/orchestration/db/models/oauth.py
```diff
@@ -73,14 +100,151 @@ def roles_display(self) -> str:
             elif self.roles & 4 != 0:
                 return 'Viewer'
 
+    @property
+    def owner(self) -> bool:
+        access = self.project_access if self.roles_new else 0
+        return self._owner or access & 1 != 0
```
wangxiaoyou1993, line 106: Can we create enum for the access 0, 1, 2, 4, 8 to make it more readable?

mage_ai/orchestration/db/models/oauth.py
```diff
@@ -73,14 +100,151 @@ def roles_display(self) -> str:
...
+                    Permission.query.filter(
+                        Permission.entity == Permission.Entity.GLOBAL,
+                        Permission.access == 8,
+                    ).first()
+                ]
+            )
+
+    def get_access(
+        self,
+        entity: Union['Permission.Entity', None],
```
wangxiaoyou1993, line 175: nit: If entity is optional, you can add a default value of None

mage_ai/server/server.py
```diff
@@ -215,12 +220,13 @@ async def main(
         # We need to sleep for a few seconds after creating all the tables or else there
         # may be an error trying to create users.
         sleep(3)
-        user = User.query.filter(User.owner == True).first()  # noqa: E712
+        Role.create_default_roles()
+        user = User.query.filter(User._owner == True).first()  # noqa: E712
```
wangxiaoyou1993, line 224: Do you need to update the code for getting the owner?

mage_ai/server/server.py
```diff
@@ -215,12 +220,13 @@ async def main(
...
         if not user:
             print('User with owner permission doesn’t exist, creating owner user.')
             if AUTHENTICATION_MODE.lower() == 'ldap':
                 user = User.create(
-                    owner=True,
+                    roles_new=[Role.query.filter(Role.name == 'owner').first()],
```
wangxiaoyou1993, line 229: I would suggest you adding a method to the Role class to get the predefined roles and wrap the method with `@safe_db_query` decorator.

mage_ai/server/server.py
```diff
@@ -229,9 +235,21 @@ async def main(
                     email='admin@admin.com',
                     password_hash=create_bcrypt_hash('admin', password_salt),
                     password_salt=password_salt,
-                    owner=True,
+                    roles_new=[Role.query.filter(Role.name == 'owner').first()],
                     username='admin',
                 )
+        else:
+            if not user.roles_new:
```
wangxiaoyou1993, line 242: Is this a one-off migration from the old roles to the new roles? If so, I would suggest do a batch update here and commit at the end. Otherwise, there could be issues if the update fails in the middle.

mage_ai/server/server.py
```diff
@@ -229,9 +235,21 @@ async def main(
...
+                all_users = User.query.all()
+                for user in all_users:
```
wangxiaoyou1993, line 244: In the "if branch", I would suggest using a different variable name than "user". Or you can rename the user above to "owner_user".

mage_ai/orchestration/db/migrations/versions/83a74eca6ea1_create_user_roles_roles_and_permissions.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = '83a74eca6ea1'
+down_revision = '66e67039b8a2'
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
```
wangxiaoyou1993, line 19: Could you test the migration script in both sqlite and postgres db?

mage_ai/api/utils.py
```diff
@@ -21,38 +27,64 @@ def authenticate_client_and_token(client_id: str, token: str) -> Tuple[Oauth2Acc
     return oauth_token, valid
 
 
-def is_owner(user) -> bool:
+def is_owner(user: User, entity=None, entity_id=None) -> bool:
     return (user and user.owner) or \
-        (not REQUIRE_USER_AUTHENTICATION and not is_test())
+        (not REQUIRE_USER_AUTHENTICATION and not is_test()) or \
+        (user and user.get_access(entity, entity_id) & 1 != 0)
```
wangxiaoyou1993, line 33: Same, let's use the enum for access instead of the number in this file.

dy46:
@wangxiaoyou1993 updated

mage_ai/api/utils.py
```diff
@@ -21,38 +27,64 @@ def authenticate_client_and_token(client_id: str, token: str) -> Tuple[Oauth2Acc
     return oauth_token, valid
 
 
-def is_owner(user) -> bool:
+def is_owner(user: User, entity=None, entity_id=None) -> bool:
```
wangxiaoyou1993, line 30: Let's add type for entity (Permission.Entity) and entity_id(str) here.

mage_ai/orchestration/db/models/oauth.py
```diff
@@ -77,8 +103,16 @@ def roles_display(self) -> str:
...
+    @hybrid_property
+    def owner(self) -> bool:
+        access = self.project_access if self.roles_new else 0
+        return self._owner or access & Permission.Access.OWNER != 0
+
     @property
     def is_admin(self) -> bool:
+        if self.roles_new:
+            access = self.project_access
+            return access & 1 == 0 and access & 2 != 0
```
wangxiaoyou1993, line 115: nit: use enum

mage_ai/api/resources/UserResource.py
```diff
@@ -114,7 +140,8 @@ def update(self, payload, **kwargs):
                 error.update(
                     {'message': 'Admins cannot update users who are Admins.'})
                 raise ApiError(error)
-            elif payload.get('roles') and int(payload.get('roles')) & 1 != 0:
+            elif payload.get('roles') and int(payload.get('roles')) & 1 != 0 or \
+                    access & 2 != 0:
```
wangxiaoyou1993, line 144: nit: use enum

mage_ai/shared/hash.py
```diff
@@ -68,9 +68,17 @@ def _build(obj, key):
     return reduce(_build, keys, {})
 
 
+def extract_including_blank_values(d, keys):
```
wangxiaoyou1993, line 71: Can you just add an optional boolean kwarg `including_blank_values` to the `extract` method if the only difference is None check?

mage_ai/server/server.py
```diff
@@ -215,12 +215,15 @@ async def main(
         # We need to sleep for a few seconds after creating all the tables or else there
         # may be an error trying to create users.
         sleep(3)
+
+        # Create new roles on existing users. This should only need to be run once.
+        Role.create_default_roles()
         user = User.query.filter(User.owner == True).first()  # noqa: E712
```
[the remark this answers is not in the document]
dy46: @wangxiaoyou1993 I looked into it more and I don't think the `hybrid_property` decorator would work for this case... I think I just need to fetch all the users and do a filter.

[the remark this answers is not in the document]
dy46: I updated the query to not fetch as many users. However, I needed to make a distinction between the "legacy" owner, and an owner in the new user authentication system which is why I make two separate queries to fetch owners.

mage_ai/orchestration/db/models/oauth.py
```diff
@@ -89,6 +123,140 @@ def git_settings(self) -> Union[Dict, None]:
...
+            if entity_permissions:
+                permissions.extend(entity_permissions)
+
+        access = 0
+        if permissions:
+            for permission in permissions:
+                access = access | permission.access
+            return access
+        else:
+            return self.get_parent_access(entity)
```
wangxiaoyou1993, line 201: Actually, i feel there're some corner cases here, which is not an issue for now but might be an issue for later (custom role).
If a role has 2 permissions (view pipeline 1, edit project), then when we get_access for pipeline 1, it'll only return view permission. But it actually also has edit access for the pipeline.

You can add a TODO here to handle role of permissions with different entity types.

mage_ai/api/policies/ClusterPolicy.py
```diff
@@ -5,7 +5,9 @@
 
 
 class ClusterPolicy(BasePolicy):
-    pass
+    @property
+    def entity(self):
+        return None, None
```
wangxiaoyou1993, line 10: Why do we override the entity with None? Should we just use the project level permission by default? Does using None entity mean a user with only permission to one pipeline can access the resource in any projects?
Same question in other resources.

mage_ai/api/resources/UserResource.py
```diff
@@ -40,6 +45,17 @@ async def create(self, payload, user, **kwargs):
...
         error = ApiError.RESOURCE_INVALID.copy()
 
+        roles_new = payload.get('roles_new', [])
+        roles_new = list(map(
+            lambda role_id: Role.query.get(int(role_id)),
+            roles_new,
+        ))
+        if any(role is None for role in roles_new):
+            error.update(
+                {'message': 'Role does not exist'})
```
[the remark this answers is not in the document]
wangxiaoyou1993: Are you going to address this comment?

mage_ai/api/resources/UserResource.py
```diff
@@ -105,20 +125,44 @@ async def _create_callback(resource):
...
-                    {'message': 'Admins cannot update users who are Admins.'})
-                raise ApiError(error)
-            elif payload.get('roles') and int(payload.get('roles')) & 1 != 0:
+        if 'roles_new' in payload:
+            roles_new = payload.get('roles_new', [])
+            roles_new = list(map(
+                lambda role_id: Role.query.get(int(role_id)),
+                roles_new,
+            ))
+            payload['roles_new'] = roles_new
```
wangxiaoyou1993, line 134: Do you need to throw error on None here?

mage_ai/orchestration/db/models/oauth.py
dy46: Hmm yeah I guess I was thinking if they have a permission on a more specific entity, that would override the "parent" entity permission. I could definitely see other corner cases though, I'll add a TODO

mage_ai/api/policies/ClusterPolicy.py
dy46: Yeah, the reason I added the `None` entity is for resources that a user could access it with any permission to the project. There are some resources that don't really belong to any entity such as the `AutocompleteItemResource`, so for those I added the `None` entity type.

mage_ai/api/resources/UserResource.py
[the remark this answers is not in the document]
dy46: oops, I forgot to do this. The main issue is the role_id is just a number, so it's not really that useful for the user. I was going to try to display the name of the role instead, but forgot.

mage_ai/api/resources/UserResource.py
dy46: ah yeah i'll do that

wangxiaoyou1993 (approved):

PR #2659, opened 2023-05-18
[jk] Limit block print output

johnson-mage:
# Summary
- Limit the code block output from print statements to avoid sending excessively large payload request bodies when saving the pipeline.

# Tests
Block print output limited to 100 lines:
print output limited to 100 lines


mage_ai/data_preparation/models/constants.py
```diff
@@ -19,6 +19,7 @@
 DATAFRAME_SAMPLE_COUNT_PREVIEW = 10
 DATAFRAME_SAMPLE_COUNT = 1000
 DATAFRAME_SAMPLE_MAX_COLUMNS = 1000
+PRINT_OUTPUT_MAX_LINES = 100
```
johnson-mage, line 22: If `100` lines for the print output doesn't seem like enough, we can increase it.

mage_ai/server/kernel_output_parser.py
```diff
@@ -44,6 +46,9 @@ def parse_output_message(message: dict) -> dict:
         text_stdout = content.get('text')
         data_content = text_stdout.split('\n')
         data_type = DataType.TEXT_PLAIN
+        if content.get('name') == 'stdout' and len(data_content) > PRINT_OUTPUT_MAX_LINES:
```
wangxiaoyou1993, line 49: The websocket server could send multiple messages to UI continuously. Limiting the number of rows in one message doesn't solve the issue completely.

johnson-mage: @wangxiaoyou1993 Added a [commit](https://github.com/mage-ai/mage-ai/pull/2659/commits/ab1511d0fd833edaf46b5fcee589a24ff4c79d04) that limits the output that would be saved to the pipeline (that should solve the issue of an excessively large req body due to a large print output) even if multiple messages with print output is sent from BE.

For example in the attached gif, I temporarily changed the `MAX_PRINT_OUTPUT_LINES` constant to `100` for the demo test. The print output sent from the BE is 101 lines, so the print output is not included in the req body when saving the pipeline (but the sample data for the table is still saved and visible after refresh, while the print output is not).
limited print output saved

mage_ai/data_preparation/models/constants.py
wangxiaoyou1993: Can we set it as an environment variable? It's only needed by certain envrionments.

johnson-mage: @wangxiaoyou1993 Does a [default](https://github.com/mage-ai/mage-ai/pull/2659/files#diff-ca01b0d0977f659f1fb140c7f3cfa840dea282ba4bceec1122326ff04cf54286R23) of `1000` seem reasonable or something higher?

mage_ai/frontend/components/PipelineDetail/constants.ts
```diff
@@ -1,4 +1,5 @@
 export const PAGE_NAME_EDIT = 'edit';
+export const MAX_REQ_PRINT_OUTPUT_LINES = 150;
```
wangxiaoyou1993, line 2: Is this still hardcoded? Can we get the env var from backend?

mage_ai/data_preparation/models/constants.py
wangxiaoyou1993: 1000 sounds reasonable

mage_ai/frontend/components/PipelineDetail/constants.ts
johnson-mage: Updated. We needed to include it in the `statuses` api response, similar to what we do for `DISABLE_NOTEBOOK_EDIT_ACCESS` and `REQUIRE_USER_AUTHENTICATION`.

wangxiaoyou1993 (approved):

PR #2675, opened 2023-05-23
[hw] Add pipeline-level configuration settings for the Spark session

csharplus:
# Summary

Enable custom settings for the Spark session used by PySpark. The behavior for Spark session is summarized below:
```
A custom SparkSession is created based on pipeline-level settings in `metadata.yaml`, and will be re-used for all blocks in the same Mage pipeline.
```
The following behavior is not supported with the current implementation:
```
1. Inside each block, the user creates and destroys a new custom SparkSession instance
2. User may replace the global kwargs['spark'] with a new SparkSession instance any time with a custom created new instance, and keep it till another new instance was created
```

# Tests

Manually tested in a local development environment, with the following steps:
1. Initiate a new Mage pipeline with the following command in a terminal:
```
% ./scripts/init.sh demo_project
```
2. Check the `metadata.yaml` generated inside the `demo_project`:
```
% cat demo_project/metadata.yaml
```
3. Make necessary adjustments to the `spark_config` settings in `metadata.yaml`:
```
spark_config:
  # Application name
  app_name: 'my spark app'
  # Master URL to connect to
  # e.g., spark_master: 'spark://host:port', or spark_master: 'yarn'
  spark_master: 'local'
  # Executor environment variables
  # e.g., executor_env: {'PYTHONPATH': '/home/path'}
  executor_env: {}
  # Jar files to be uploaded to the cluster and added to the classpath
  # e.g., spark_jars: ['/home/path/example1.jar']
  spark_jars: []
  # Path where Spark is installed on worker nodes,
  # e.g. spark_home: '/usr/lib/spark'
  spark_home: null
  # List of key-value pairs to be set in SparkConf
  # e.g., others: {'spark.executor.memory': '4g', 'spark.executor.cores': '2'}
  others: {}
``` 
4. Start the Mage application with the following command:
```
% ./scripts/dev.sh demo_project
```
5. Click the Terminal icon on the right side of the Mage UI, and install the required packages and libraries for pyspark:
```
root@488dc9529cf3:/home/src#  apt-get update && \
apt-get install -y --no-install-recommends \
        openjdk-11-jre

root@488dc9529cf3:/home/src#  pip install pyspark
```
6. Stop and restart the Mage application to pick up the newly installed pyspark libraries.
7. Create a new `Standard (batch)` pipeline, and add a new `Data loader` -> `Python` -> `Generic (no template)` block to it.
8. Add the following code to the `load_data` function:
```
@data_loader
def load_data(*args, **kwargs):
    """
    Template code for loading data from any source.

    Returns:
        Anything (e.g. data frame, dictionary, array, int, str, etc.)
    """
    # Specify your data loading logic here

    spark = kwargs.get('spark')
    spark_conf = spark.sparkContext.getConf()
    print('#' * 40)
    print(spark_conf.toDebugString())
    print('#' * 40)
    return {}
```
9. Run the `Data loader` by clicking the Play button on its toolbar. It would generate results similar to the following, which reflects the settings given in the `metadata.yaml` file listed above. 
Screenshot 2023-05-22 at 11 57 20 PM

cc:

@dy46 @wangxiaoyou1993

### DIFF
--- a/mage_ai/data_preparation/models/block/__init__.py
+++ b/mage_ai/data_preparation/models/block/__init__.py
@@ -1786,11 +1786,11 @@ df = get_variable('{self.pipeline.uuid}', '{block_uuid}', 'df')
         if self.spark_init:
             return self.spark
         try:
-            repo_config = RepoConfig(repo_path=self.repo_path)
-            spark_config = repo_config.get('spark_config')
-
             from pyspark.conf import SparkConf
             from pyspark.sql import SparkSession
+
+            repo_config = RepoConfig(repo_path=self.repo_path)
+            spark_config = repo_config.spark_config
 
             if spark_config:
                 conf = SparkConf()
--- a/mage_ai/data_preparation/models/block/__init__.py
+++ b/mage_ai/data_preparation/models/block/__init__.py
@@ -1801,7 +1801,7 @@ df = get_variable('{self.pipeline.uuid}', '{block_uuid}', 'df')
                 if spark_config.get('spark_home'):
                     conf.setSparkHome(spark_config.get('spark_home'))
                 if spark_config.get('spark_jars'):
-                    conf.setJars(spark_config.get('spark_jars'))
+                    conf.set('spark.jars', ','.join(spark_config.get('spark_jars')))
                 if spark_config.get('others'):
                     conf.setAll(spark_config.get('others'))
 
--- a/mage_ai/data_preparation/models/block/__init__.py
+++ b/mage_ai/data_preparation/models/block/__init__.py
@@ -47,7 +47,7 @@ from mage_ai.data_preparation.models.constants import (
 )
 from mage_ai.data_preparation.models.file import File
 from mage_ai.data_preparation.models.variable import VariableType
-from mage_ai.data_preparation.repo_manager import get_repo_path
+from mage_ai.data_preparation.repo_manager import RepoConfig, get_repo_path
 from mage_ai.data_preparation.shared.stream import StreamToLogger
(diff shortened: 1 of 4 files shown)


mage_ai/data_preparation/models/block/__init__.py
```diff
@@ -1784,11 +1786,33 @@ def __get_spark_session(self):
...
+            if spark_config:
+                conf = SparkConf()
+                if spark_config.get('app_name'):
+                    conf.setAppName(spark_config.get('app_name'))
+                if spark_config.get('spark_master'):
+                    conf.setMaster(spark_config.get('spark_master'))
+                if spark_config.get('spark_home'):
+                    conf.setSparkHome(spark_config.get('spark_home'))
+                if spark_config.get('spark_jars'):
+                    conf.setJars(spark_config.get('spark_jars'))
```
wangxiaoyou1993, line 1804: Did you test setJars? It throws error `AttributeError: 'SparkConf' object has no attribute 'setJars'`

csharplus: Good catch: this method is available in Java but not in PySpark, as shown in the Python source code for [SparkConf](https://github.com/apache/spark/blob/master/python/pyspark/conf.py#L171). I found a work-around in [Python SparkConf.setExecutorEnv Examples](https://python.hotexamples.com/examples/pyspark/SparkConf/setExecutorEnv/python-sparkconf-setexecutorenv-method-examples.html), and made a fix. Please review this change. I'll test it out tomorrow as well.

wangxiaoyou1993 (approved):

PR #2690, opened 2023-05-24
[hw] Update the Spark user guide with a Custom Spark Session section

csharplus:
# Summary

Add document related to the newly added Spark custom session feature.

# Tests

Manually tested with a local development environment.
cc:

@dy46 @wangxiaoyou1993

### DIFF
--- a/mage_ai/data_preparation/models/block/utils.py
+++ b/mage_ai/data_preparation/models/block/utils.py
@@ -1,8 +1,10 @@
+from mage_ai.data_preparation.repo_manager import RepoConfig, get_repo_path
 from mage_ai.shared.array import find, unique_by
 from mage_ai.shared.hash import merge_dict
 from mage_ai.shared.utils import clean_name as clean_name_orig
 from typing import Any, Dict, List, Tuple
 import json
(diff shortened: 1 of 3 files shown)


mage_ai/data_preparation/models/block/utils.py
```diff
@@ -536,3 +538,34 @@ def fetch_input_variables(
                 input_vars[idx] = final_value
 
     return input_vars, kwargs_vars, upstream_block_uuids_final
+
+
+def get_spark_session():
```
wangxiaoyou1993, line 543: I would suggest creating a `spark` folder in https://github.com/mage-ai/mage-ai/tree/master/mage_ai/services
then put spark related methods in the file in that folder.

csharplus: > I would suggest creating a `spark` folder in https://github.com/mage-ai/mage-ai/tree/master/mage_ai/services then put spark related methods in the file in that folder.

Sure, made the suggested change.

wangxiaoyou1993 (approved):

PR #2718, opened 2023-05-26
[jk] Snowflake sql block defaults

johnson-mage:
# Summary
- Allow more flexibility in Snowflake sql block queries. Users only need to include the database and schema config in one of these 3 places, in order of priority (so if all 3 are included, `#1` takes priority, then `#2`, then `#3`):
1. Users can include the db and schema directly in the query (e.g. `select * from demo_db.mage_dev2.jk_demo_table;`)
2. Users can include the db and schema in the SQL code block header inputs. For example:
image
3. Users can include the default db and schema in the `io_config.yaml using these fields:
```
SNOWFLAKE_DEFAULT_DB: DEMO_DB
SNOWFLAKE_DEFAULT_SCHEMA: MAGE_DEV2
``` 
# Tests
snowflake query
- Also confirmed snowflake queries were successful with the priorities in the scenarios above.

### DIFF
--- a/mage_ai/data_preparation/models/block/sql/utils/shared.py
+++ b/mage_ai/data_preparation/models/block/sql/utils/shared.py
@@ -1,6 +1,6 @@
 import re
 from os import path
-from typing import Callable, Dict, List, Tuple
+from typing import Callable, Dict, List, Tuple, Union
 
 from jinja2 import Template
 from pandas import DataFrame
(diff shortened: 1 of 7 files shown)


mage_ai/data_preparation/models/block/sql/utils/shared.py
```diff
@@ -257,6 +257,17 @@ def table_name_parts(
     return database, schema, table
 
 
+def table_name_parts_from_query(
+    query: str,
+) -> Tuple[str, str, str]:
+    match = re.search(r'[^.+]from[\s](.+)\.(.+)\.(.+)', query, re.IGNORECASE)
```
wangxiaoyou1993, line 263: why is there a `from` string in the regex pattern?
also, could you add a unit test for this method?

wangxiaoyou1993: could you also add comment to this method?

wangxiaoyou1993 (approved):

wangxiaoyou1993 (approved):

mage_ai/data_preparation/models/block/sql/utils/shared.py
wangxiaoyou1993: If the method can also return None, you should set the return type to `Union[Tuple[str, str, str], None]`

johnson-mage: @wangxiaoyou1993 Added a comment. Also updated the regex to be more flexible (e.g. allow for a WHERE clause). I also added a condition so that we are only parsing the table name parts if the user is NOT using a raw sql query. There might be some unexpected issues with the raw sql query I'm not handling, since this was moreso for users not customizing their own queries and making it more convenient to use SELECT queries specifically (hence the `from` in the regex).

I will add a unit test.

PR #2761, opened 2023-06-05
[jk] Export block output to csv file

johnson-mage:
# Summary
- Add `Save` button to code block footer in Pipeline Editor and also the Block Runs table (`/pipelines/[pipeline_uuid]/runs/[pipeline_run_id]`) that opens save dialogue for downloading the block output as a CSV file.
- Steps: 
1. Pipeline Editor: Run a code block and then click the `Save` icon in the block output footer (should work even with `DISABLE_NOTEBOOK_EDIT_ACCESS` enabled).
2. Block Runs table: Click on the `Save` button under the `Output` column of the table. The block run must belong to a standard pipeline and have `completed` status.

# Tests
- Tested on Editor role with `DISABLE_NOTEBOOK_EDIT_ACCESS` env var enabled. Once the download is in progress, the download request is non-blocking (user can still use the app and make other api requests that can complete even while downloading the block output).

In Pipeline Editor in block output footer:
code block output
block output 1M rows

In Block Runs Table:
image


mage_ai/server/api/downloads.py
```diff
@@ -0,0 +1,42 @@
+from mage_ai.data_preparation.models.pipeline import Pipeline
+from mage_ai.data_preparation.models.variable import VariableType
+from mage_ai.orchestration.db.models.schedules import PipelineRun
+from mage_ai.server.api.base import BaseHandler
+from tornado import gen, iostream
+
+
+class ApiDownloadHandler(BaseHandler):
+    async def get(self, pipeline_uuid, block_uuid, **kwargs):
```
wangxiaoyou1993, line 9: Could you add authentication to this api?

johnson-mage: added

johnson-mage: tested the API request with/without the token and confirmed that an exception is raised if no valid token was included

mage_ai/server/api/downloads.py
```diff
@@ -0,0 +1,61 @@
...
+
+        tables = block.get_outputs(
+            execution_partition=execution_partition,
+            include_print_outputs=False,
+            csv_lines_only=True,
+            sample=False,
+            variable_type=VariableType.DATAFRAME,
+        )
+        for table in tables:
+            line_count = len(table)
```
wangxiaoyou1993, line 52: i think the table is just a string. why is the length the line_count?

mage_ai/data_preparation/models/block/__init__.py
```diff
@@ -1190,16 +1192,21 @@ def get_outputs(
...
-                    sample_data=dict(
-                        columns=columns_to_display,
-                        rows=json.loads(data[columns_to_display].to_json(orient='split'))['data']
-                    ),
-                    shape=[row_count, column_count],
-                    type=DataType.TABLE,
-                    variable_uuid=v,
-                )
+                if csv_lines_only:
+                    data = data.to_csv(header=True, index=False).strip('\n')
```
wangxiaoyou1993, line 1196: I think this changes the return value type of this the method. Could you either return the value in the format or update the method return value type?
It would be better to add method comments as well.

mage_ai/data_preparation/models/block/__init__.py
```diff
@@ -1190,16 +1192,21 @@ def get_outputs(
...
-                data = dict(
-                    sample_data=dict(
-                        columns=columns_to_display,
-                        rows=json.loads(data[columns_to_display].to_json(orient='split'))['data']
-                    ),
-                    shape=[row_count, column_count],
-                    type=DataType.TABLE,
-                    variable_uuid=v,
-                )
+                if csv_lines_only:
```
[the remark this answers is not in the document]
johnson-mage: updated

mage_ai/server/api/downloads.py
johnson-mage: thanks for catching that. I thought it was a list of strings, but I read the output wrong. It was writing each character instead of line, so ive updated it and the download is a lot faster now.

mage_ai/data_preparation/models/block/__init__.py
johnson-mage: Got it, somewhat related to the other comment, but didn't catch that it was a giant string. Updated the return value to be consistent with the return type. It returns a dictionary with a value that is a list of strings (pertaining to each row in the table).

mage_ai/server/api/downloads.py
```diff
@@ -0,0 +1,64 @@
...
+            table = data.get('table', [])
+            line_count = len(table)
+            for line in range(0, line_count):
+                is_last_line = line == line_count - 1
+                try:
+                    csv_line = table[line] if is_last_line else table[line] + '\n'
+                    self.write(csv_line.encode('UTF-8'))
+                except iostream.StreamClosedError:
+                    break
+                if (line == 0 or (line % 5000 == 0 or is_last_line)):
```
wangxiaoyou1993, line 61: nit: you can simplify it to be `if (line % 5000 == 0 or is_last_line)`

wangxiaoyou1993 (approved):

mage_ai/server/api/downloads.py
johnson-mage: howd i miss that 😭

PR #2770, opened 2023-06-06
[dy] Workspace management p3

dy46:
Attempt 2 at mage-ai/mage-ai#2760.

Updates how we do project level permissions. Instead of using the `repo_path` as the identifier for a project, we use the `project_uuid` instead.

One change I made is to update how we store the subproject files. I was having a lot trouble getting the volumes to work when the files for the subproject were stored in a subdirectory of the main project. I changed the main project to only store the project_uuid of the subproject so that the project level permissions could be created, but the rest of the subproject files would be created when the subproject was initialized. The main downside to doing this is that we can't reuse metadata from the main project to the subproject, but we can add that in later.

Tested locally and on GKE.

cc: @wangxiaoyou1993


dy46:
@wangxiaoyou1993 can you take a look when you have time?

wangxiaoyou1993:
> @wangxiaoyou1993 can you take a look when you have time?

Yes, will review it today

dy46:
i'm going to try to move some of the changes out to a separate PR. there are some changes in here that are not directly related to workspaces

mage_ai/cli/main.py
```diff
@@ -27,6 +27,10 @@ def list_commands(self, ctx: Context):
 INIT_PROJECT_PATH_DEFAULT = typer.Argument(..., help='path of the Mage project to be created.')
 INIT_PROJECT_TYPE_DEFAULT = typer.Option(
     'standalone', help='type of project to create, options are main, sub, or standalone')
+INIT_CLUSTER_TYPE_DEFAULT = typer.Option(
+    'k8s',
```
wangxiaoyou1993, line 31: Can we have a enum for cluster type?

mage_ai/cli/main.py
```diff
@@ -36,6 +40,18 @@ def list_commands(self, ctx: Context):
 START_DBT_DOCS_INSTANCE_DEFAULT = typer.Option('0', help='')
 START_INSTANCE_TYPE_DEFAULT = typer.Option(
     InstanceType.SERVER_AND_SCHEDULER.value, help='specify the instance type.')
+START_PROJECT_TYPE_DEFAULT = typer.Option(
+    'standalone',
```
wangxiaoyou1993, line 44: Same here. Can we have a enum for project type?

mage_ai/cli/main.py
```diff
@@ -36,6 +40,18 @@ def list_commands(self, ctx: Context):
...
+    help='create project of this type if does not exist, options are main, sub, or standalone',
+)
+START_CLUSTER_TYPE_DEFAULT = typer.Option(
+    'k8s',
+    help='type of instance to create for workspace management',
+)
+START_PROJECT_UUID_DEFAULT = typer.Option(
+    None,
+    help='type of instance to create for workspace management',
```
wangxiaoyou1993, line 53: Is this description correct?

mage_ai/cli/main.py
```diff
@@ -36,6 +40,18 @@ def list_commands(self, ctx: Context):
 START_DBT_DOCS_INSTANCE_DEFAULT = typer.Option('0', help='')
 START_INSTANCE_TYPE_DEFAULT = typer.Option(
     InstanceType.SERVER_AND_SCHEDULER.value, help='specify the instance type.')
+START_PROJECT_TYPE_DEFAULT = typer.Option(
+    'standalone',
+    help='create project of this type if does not exist, options are main, sub, or standalone',
+)
+START_CLUSTER_TYPE_DEFAULT = typer.Option(
```
wangxiaoyou1993, line 47: Why is the default cluster type "k8s"? Should it be None by default?

mage_ai/data_preparation/repo_manager.py
```diff
@@ -155,28 +154,34 @@ def save(self, **kwargs) -> None:
             yml.dump(data, f)
 
 
-def init_repo(repo_path: str, project_type: str = ProjectType.STANDALONE) -> None:
+def init_repo(
+    repo_path: str,
+    project_type: str = ProjectType.STANDALONE,
+    cluster_type: str = 'k8s',
```
wangxiaoyou1993, line 160: Same here. Why is the default value "k8s"? Should it be None by default?

mage_ai/data_preparation/repo_manager.py
```diff
@@ -185,6 +190,11 @@ def init_repo(repo_path: str, project_type: str = ProjectType.STANDALONE) -> Non
         )
         copy_template_directory('repo', repo_path)
 
+    if project_uuid is None:
```
wangxiaoyou1993, line 193: also check empty string

mage_ai/server/server.py
```diff
@@ -290,19 +293,28 @@ def start_server(
     manage: bool = False,
     dbt_docs: bool = False,
     instance_type: InstanceType = InstanceType.SERVER_AND_SCHEDULER,
+    project_type: ProjectType = ProjectType.STANDALONE,
+    cluster_type: ClusterType = ClusterType.K8S,
```
wangxiaoyou1993, line 297: Same here about the default value for cluster type.

mage_ai/api/resources/UserResource.py
```diff
@@ -126,18 +120,13 @@ def update(self, payload, **kwargs):
...
-            if len(missing_values) >= 1:
-                error.update(
-                    {'message': 'Missing required values: {}.'.format(', '.join(missing_values))})
-                raise ApiError(error)
-
-            access = get_access_for_roles(roles_new, Permission.Entity.PROJECT, get_repo_path())
+            access = get_access_for_roles(
+                roles_new,
+                Permission.Entity.PROJECT,
+                get_project_uuid(),
```
wangxiaoyou1993, line 128: Does it work for project that has no uuid?

mage_ai/api/resources/PipelineResource.py
```diff
@@ -79,7 +79,10 @@ def query_pipeline_schedules(pipeline_uuids):
                     a.status,
                     a.updated_at,
                 ]).
-                filter(a.pipeline_uuid.in_(pipeline_uuids))
+                filter(
+                    a.pipeline_uuid.in_(pipeline_uuids),
+                    a.repo_path == get_repo_path(),
```
wangxiaoyou1993, line 84: should we use project_uuid here as well? will there be duplicate repo_path in subprojects?

mage_ai/cluster_manager/kubernetes/workload_manager.py
```diff
@@ -146,55 +151,64 @@ def create_deployment(
...
                     }
                 },
+                'volumeClaimTemplates': [
+                    {
+                        'metadata': {
+                            'name': 'mage-data'
+                        },
+                        'spec': {
+                            'accessModes': [
+                                'ReadWriteOnce'
```
wangxiaoyou1993, line 195: this could cause problem. we usually use ReadWriteMany. but we can make it configurable.
1G below might not always enough as well.

mage_ai/cli/main.py
dy46: yeah you're right

mage_ai/api/resources/UserResource.py
dy46: every project should have a uuid after this PR. also there shouldn't be any project level permissions created before this unless a user manually created one.

mage_ai/api/resources/PipelineResource.py
dy46: i already made the change to save the repo path in previous PR, so if i change it to project_uuid here, it may affect existing projects.

the repo_paths for the subprojects should be different because each subproject should have a different project name.

mage_ai/cluster_manager/kubernetes/workload_manager.py
dy46: ok i can make this configurable

dy46:

mage_ai/settings/__init__.py
```diff
@@ -47,3 +47,23 @@ def is_disable_pipeline_edit_access():
 
 DEFAULT_LOCALHOST_URL = 'http://localhost:6789'
 MAGE_PUBLIC_HOST = os.getenv('MAGE_PUBLIC_HOST') or DEFAULT_LOCALHOST_URL
+
+
+MAGE_SETTINGS_ENVIRONMENT_VARIABLES = [
```
wangxiaoyou1993, line 52: Could you add comment for this list so that other developers know what environment variable should be in this list?

mage_ai/server/server.py
```diff
@@ -290,19 +293,28 @@ def start_server(
...
 
     if not os.path.exists(project):
-        init_repo(project)
+        init_repo(
+            project,
+            project_type=project_type,
+            cluster_type=cluster_type,
+            project_uuid=project_uuid,
+        )
+    update_project_uuid()
```
wangxiaoyou1993, line 317: I think you'll need to move it below `set_repo_path`. Since `get_repo_config` relies on the repo path in the environment variable.

mage_ai/orchestration/db/models/oauth.py
```diff
@@ -192,7 +192,7 @@ def get_access(
         entity_id: Union[str, None] = None,
     ) -> int:
         permissions = []
-        if entity is None:
+        if entity == Permission.Entity.ANY:
```
wangxiaoyou1993, line 195: Will entity be None? What do we return for that case?

mage_ai/cluster_manager/kubernetes/workload_manager.py
```diff
@@ -220,16 +243,41 @@ def create_deployment(
...
-    def __populate_env_vars(self, container_config) -> List:
-        env_vars = []
+    def delete_workload(self, name: str):
+        self.apps_client.delete_namespaced_stateful_set(name, self.namespace)
+        self.core_client.delete_namespaced_service(f'{name}-service', self.namespace)
+
+    def __populate_env_vars(
+        self,
+        name,
+        project_type: str = 'standalone',
```
wangxiaoyou1993, line 261: Let's use enum value here.

mage_ai/cluster_manager/kubernetes/workload_manager.py
```diff
@@ -68,25 +71,38 @@ def list_services(self):
...
+        if container_config_yaml:
+            container_config = yaml.full_load(container_config_yaml)
 
-        env_vars = self.__populate_env_vars(container_config)
+        service_account_name = kwargs.get(
+            'service_account_name',
+            os.getenv(KUBE_SERVICE_ACCOUNT_NAME),
+        )
+        storage_class_name = kwargs.get('storage_class_name', os.getenv(KUBE_STORAGE_CLASS_NAME))
+        access_mode = kwargs.get('access_mode', 'ReadWriteOnce')
```
wangxiaoyou1993, line 91: nit: storage_access_mode

mage_ai/cli/main.py
```diff
@@ -26,7 +27,17 @@ def list_commands(self, ctx: Context):
 
 INIT_PROJECT_PATH_DEFAULT = typer.Argument(..., help='path of the Mage project to be created.')
 INIT_PROJECT_TYPE_DEFAULT = typer.Option(
-    'standalone', help='type of project to create, options are main, sub, or standalone')
+    ProjectType.STANDALONE,
```
wangxiaoyou1993, line 30: I think you'll need to use `ProjectType.STANDALONE.value` here. Otherwise, the cli help message will show "ProjectType.STANDALONE" as the default value, which looks a bit weird.

mage_ai/cli/main.py
```diff
@@ -36,6 +47,18 @@ def list_commands(self, ctx: Context):
 START_DBT_DOCS_INSTANCE_DEFAULT = typer.Option('0', help='')
 START_INSTANCE_TYPE_DEFAULT = typer.Option(
     InstanceType.SERVER_AND_SCHEDULER.value, help='specify the instance type.')
+START_PROJECT_TYPE_DEFAULT = typer.Option(
+    ProjectType.STANDALONE,
```
wangxiaoyou1993, line 51: Same here. `ProjectType.STANDALONE.value`

wangxiaoyou1993 (approved):

mage_ai/orchestration/db/models/oauth.py
dy46: It shouldn't be None, but I added a check for it just in case.

PR #2775, opened 2023-06-09
Gjvv fix mssql

gjvanvuuren:
# Summary
The command CREATE SCHEMA IF NOT EXISTS is not supported by MSSQL. Provided a default command in BaseSQL -> build_create_schema_command, and an overridden implementation in MSSQL -> build_create_schema_command containing compatible syntax.

# Tests
Called `build_create_schema_command('my_schema')` in REPL and executed returned string on Azure Sql database.

cc:


wangxiaoyou1993 (approved):
Thanks for fixing the schema creation for mssql!

PR #2777, opened 2023-06-10
2776 Monaco Editor Fix - CDN Change

halidrauf:
# Summary
Change CDN from JSDeliver to CDNJS by Cloudflare

# Tests
Compiled and ran in a Docker Container

cc:
@johnson-mage


johnson-mage:
@halidrauf Thanks for putting up this PR, taking a look.

johnson-mage:
@halidrauf Thanks for looking up the alternative CDN. Unfortunately, I don't think this would permanently change the CDN used by the Monaco Editor, since once we create a new frontend production build, it would overwrite this change. 

Refer to my PR https://github.com/mage-ai/mage-ai/pull/2821 for how we could update the CDN. I referred to this (https://github.com/suren-atoyan/monaco-react#loader-config) for reference. Going to close this PR since we will not be merging it.

PR #2839, opened 2023-06-22
[jk]  Update error messaging for testing connection and loading sample data

johnson-mage:
# Summary
Improve visibility into non-functioning "test connection" and "load sample data" features for integration pipelines:
- Show unsupported error is "Test connection" is not implemented for an integration source.
- Update error messaging for "Load sample data" to let user know that it may not be supported for the currently selected integration source.

# Tests
Testing connection when `test_connection` method not implemented (tested on all sources where the `test_connection` method was removed):
image

Load sample data error (a bit verbose but if the `load_data` method is not implemented, a clear exception message is not provided otherwise):
image

cc:

### DIFF
--- a/mage_ai/data_preparation/models/pipelines/integration_pipeline.py
+++ b/mage_ai/data_preparation/models/pipelines/integration_pipeline.py
@@ -286,9 +286,9 @@ class IntegrationPipeline(Pipeline):
                     except Exception:
                         error = line
             if not error:
-                raise Exception('The sample data was not able to be loaded. Please check \
-                                if the stream still exists. If it does not, click the "View and \
-                                select streams" button and confirm the valid streams.')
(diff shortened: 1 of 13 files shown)


mage_ai/data_preparation/models/pipelines/integration_pipeline.py
```diff
@@ -212,6 +212,8 @@ def test_connection(self, block_type: BlockType, config: str = None):
                         error = dig(json_object, 'tags.error')
                     except Exception:
                         error = line
+                elif not error and line.startswith('CRITICAL'):
```
johnson-mage, line 215: Included this fallback cuz there may be instances where user clicks "Test connection" but the error is blank. For example, this scenario occurred when testing the "Test connection" for Commercetools and Stripe, but the fallback  error message has some details the user can act upon:

image

image

PR #2858, opened 2023-06-24
Adding Salesforce source to Streaming Pipelines

Luishfs:
# Summary
Added Salesforce source to Streaming Pipelines using aiosfstream
Added docs to Salesforce Source

### Python 3.10 Support
To allow aiosfstream Python 3.10 support, changes were made on a [forked aiocometd library](https://github.com/Luishfs/aiocometd)
This forked library was added on all Dockerfiles 

### Async Streaming Pipelines
In [Streaming executor](https://github.com/mage-ai/mage-ai/blob/master/mage_ai/data_preparation/executors/streaming_pipeline_executor.py#L179C1-L179C44) file, async.get_event_loop() was being used, and since no event loop was being created, a raise would happen.
Changes were made to catch that error, and if no event loop is currently running, loop will be set to None and code runs as default

# Tests
Tests were made using a Developer Salesforce Account, and by creating events and triggers on Account creation.

Salesforce Source Trigger logs:
Screenshot from 2023-06-24 17-07-34

cc:
@wangxiaoyou1993


myrulezzz:
@Luishfs It would be nice if you can add also Salesforce CDC as a second option to connect to Salesforce.Thanks

PR #2899, opened 2023-07-03
[hw] Make a fix for wrong parameter passing

csharplus:
# Summary

The passing of `tries` was fixed to use `tag` instead of `kwargs`.

# Tests

CI
cc:

@dy46 @wangxiaoyou1993

### DIFF
--- a/mage_integrations/mage_integrations/destinations/bigquery/__init__.py
+++ b/mage_integrations/mage_integrations/destinations/bigquery/__init__.py
@@ -22,6 +22,7 @@ from mage_integrations.destinations.bigquery.constants import (
 from mage_integrations.destinations.bigquery.utils import (
     convert_array,
     convert_array_for_batch_load,
+    convert_column_if_json,
     convert_column_type,
     convert_converted_type_to_parameter_type,
     convert_datetime,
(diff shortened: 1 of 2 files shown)


csharplus:
@wangxiaoyou1993 please take a look at this minor fix, and see if it is Ok to merge.

mage_integrations/mage_integrations/destinations/bigquery/__init__.py
```diff
@@ -265,7 +259,7 @@ def process_queries(
                         record_data=record_data,
                         stream=stream,
                         tags=tags,
-                        tries=tries,
```
wangxiaoyou1993, line 268: I think the parameter passing also works here. `tries` is passed via kwargs.

csharplus: @wangxiaoyou1993 the signature of the same function changed when calling recursively changed if passing `tries` this way, i.e.,
**Old signature**
```
    def process_queries(
        self,
        query_strings: List[str],
        record_data: List[Dict],
        stream: str,
        tags: Dict = None,
        **kwargs,
    ) 
```
**New signature**
```
    return self.process_queries(
                  query_strings=query_strings,
                  record_data=record_data,
                  stream=stream,
                  tags=tags,
                  tries=tries,
```
`kwargs` may need to be updated to contain the latest `tries` parameter if we indeed want to pass `tries` through `kwargs`.

csharplus:
@wangxiaoyou1993 please take a look at this change and see if it is ok to merge.

PR #2906, opened 2023-07-05
[jk] Pipeline monitoring dashboard

johnson-mage:
# Summary
- Add "Overview" page to dashboard providing summary of pipeline run failures.

# Tests
overview_dashboard2


mage_ai/orchestration/monitor/monitor_stats.py
```diff
@@ -72,10 +73,18 @@ def get_pipeline_run_count(
             data = stats_by_schedule_id[p.pipeline_schedule_id]['data']
             if created_at_formatted not in data:
                 data[created_at_formatted] = dict()
-            if p.status not in data[created_at_formatted]:
-                data[created_at_formatted][p.status] = 1
+            if group_by_pipeline_type:
```
johnson-mage, line 76: @wangxiaoyou1993 So the API endpoint that is taking a while to load is this `/api/monitor_stats/pipeline_run_count?group_by_pipeline_type=1&start_time=2023-06-28T00:00:00.000Z`, and specifically, this block of code for grouping the stats by pipeline type (adding in the `group_by_pipeline_type` query param) seems to slow the request down a lot.

wangxiaoyou1993: `p.pipeline_type` is very inefficient because it reads the file. we don't want to read file for each pipeline run.

johnson-mage: @wangxiaoyou1993 thanks for the tips, loading the "Last 30 days" tab only takes about a second to load the stats now

PR #2915, opened 2023-07-06
[dy] Update serialization for mssql

dy46:
# Summary

I updated the serialization of mssql objects to be more like postgres. Objects will only be converted to json if the data type is a dictionary or list.

# Tests

tested locally

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/data_preparation/models/block/sql/__init__.py
+++ b/mage_ai/data_preparation/models/block/sql/__init__.py
@@ -26,7 +26,6 @@ from mage_ai.io.base import QUERY_ROW_LIMIT, DataSource, ExportWritePolicy
 from mage_ai.io.config import ConfigFileLoader
 from mage_ai.settings.repo import get_repo_path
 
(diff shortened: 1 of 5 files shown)


mage_ai/data_preparation/models/block/sql/__init__.py
```diff
@@ -544,46 +544,6 @@ def execute_sql_code(
...
-def split_query_string(query_string: str) -> List[str]:
-    text_parts = []
-
-    matches = re.finditer(r"'(.*?)'|\"(.*?)\"", query_string, re.IGNORECASE)
-
-    previous_idx = 0
-
-    for _, match in enumerate(matches):
-        matched_string = match.group()
-        updated_string = re.sub(r';', MAGE_SEMI_COLON, matched_string)
```
wangxiaoyou1993, line 556: You can remove the MAGE_SEMI_COLON definition in this file as well.

wangxiaoyou1993 (approved):

PR #2936, opened 2023-07-09
[xh]Generate documentation for block and pipeline

matrixstone:
# Summary
Generate documentation for block and pipeline.

This PR creates a `LLMPipelineWizard` class under `ai/` folder.
Two major functions it exposes are `generate_pipeline_documentation` and `generate_block_documentation`

# Commands to run:
`python mage_ai/ai/gen_doc_for_pipeline.py generate-pipeline-documentation project_name pipeline_name --print-block-doc`

# Sample of documentation

## For MySQL data loader block:
image

**Documentation generated**
This code is used to load data from a MySQL database. It uses the `get_repo_path()` function from the `mage_ai.data_preparation.repo_manager` module to get the path of the repository. The configuration settings are specified in the `io_config.yaml` file. The `load_data_from_mysql()` function is used to load the data from the MySQL database using the specified query. The `test_output()` function is used to test the output of the block. It checks if the output is not None.

## For Filter Transformer Block
image

**Documentation generated**
This python code implements a custom transformer action to filter data from a given DataFrame. The execute_transformer_action function takes a DataFrame as an argument and returns a filtered DataFrame based on the action_code specified. The action_code is a boolean expression that is used to filter the rows of the DataFrame. The action_code can contain any valid boolean expression that is supported by the BaseAction class. The output of the execute_transformer_action function is then tested using the test_output function. The test_output function checks if the output is not None and raises an assertion error if it is.

## For S3 explorter
image

**Documentation generated**
This python code is used to export data to a S3 bucket. The configuration settings are specified in the 'io_config.yaml' file. The bucket name and object key need to be specified in order to export the data. The S3 class is used to export the data with the help of the ConfigFileLoader class. The data_exporter decorator is used to export the data. The documentation for this code can be found at https://docs.mage.ai/design/data-loading#s3.

## Pipeline
image

**Documentation generated**
This data pipeline is used to load data from a MySQL database, an API, and a DataFrame. It then uses custom transformer actions to filter, impute missing values, count distinct values, and remove outliers from the data. Finally, it exports the data to a S3 bucket. The code uses the `get_repo_path()` function from the `mage_ai.data_preparation.repo_manager` module to get the path of the repository, the requests library to make a request to the API, the pandas library to read the response into a DataFrame, the `load_data_from_mysql()` function to load the data from the MySQL database, the `execute_transformer_action` function to filter, impute missing values, count distinct values, and remove outliers from the data, and the S3 class to export the data. The `test_output()` function is used to test the output of each block.

## For SQL block and pipeline:

**Documentation generated**
This SQL script is used to select all data from the users table. It will return all columns and rows from the users table.

The execute_transformer_action function is a custom SQL script used to transform data from one format into a different format. It takes in two parameters, the source table and the target table, and performs the transformation on the data.

The script begins by creating a temporary table to store the transformed data. It then uses a SELECT statement to retrieve the data from the source table and insert it into the temporary table. The SELECT statement includes a CASE statement to transform the data into the desired format.

Once the data has been transformed, the script uses an INSERT statement to insert the data into the target table. Finally, the script drops the temporary table.

This script is useful for transforming data from one format to another quickly and easily. It can be used to transform data from one database to another, or to transform data from one format to another within the same database.

This SQL script creates a table called 'mage.users2' if it does not already exist. The table contains two columns, 'id' and 'username', with 'id' being a BIGINT and 'username' being a VARCHAR of length 255. The script then inserts data from the dataframe 'df_1' into the 'mage.users2' table.

cc:
@wangxiaoyou1993


requirements.txt
```diff
@@ -21,9 +21,11 @@ joblib>=1.1.0
 jupyter-server-proxy==3.2.1
 jupyter-server==1.23.5
 jupyter_client==7.3.4
+langchain>=0.0.222
```
wangxiaoyou1993, line 24: Let's make add these to the extra packages.
* Move it below the `# extras` comment
* Add a section here (e.g. ai) https://github.com/mage-ai/mage-ai/blob/master/setup.py#L42
* Add the two packages to the ai section and all section

wangxiaoyou1993 (approved):

wangxiaoyou1993:
Could you fix CI? Is `langchain` supported in python 3.7?

requirements.txt
matrixstone: Updated

requirements.txt
```diff
@@ -57,7 +57,9 @@ google-cloud-pubsub==2.16.0
 google-cloud-run==0.5.0
 google-cloud-storage==2.5.0
 kubernetes==25.3.0
+langchain>=0.0.222
```
wangxiaoyou1993, line 60: You can update it to `langchain>=0.0.222; python_version >= '3.8'`

matrixstone: Updated

wangxiaoyou1993 (approved):

PR #2990, opened 2023-07-14
[dy] Remove data integration config from logs

dy46:
# Summary

I added filtering to data integration logging to hide string source/destination config values. 

At some point we should make the data integration logging more organized since we log things in a lot of places.

# Tests

tested locally

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/shared/security.py
+++ b/mage_ai/shared/security.py
@@ -23,9 +23,17 @@ def filter_out_env_var_values(value: str) -> str:
 def filter_out_config_values(log: str, config: Dict) -> str:
     """
     Used for integration pipeline logging security. Filters out values from
-    the passed in config.
+    the passed in config if the config value is a string and is over
+    MIN_SECRET_LENGTH characters long.
+
+    Args:
+        log (str): The log to filter.
+        config (Dict[str, any]): The config to get filter values from.
+
+    Returns:
+        str: Log with config values filtered out.
     """
-    if config is None:
+    if not log or not config:
         return log
     values = config.values()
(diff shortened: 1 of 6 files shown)


mage_ai/shared/security.py
```diff
@@ -10,15 +11,33 @@
...
-    value_clean = value
-    for env_var_value in env_var_values:
-        replace_value = '*' * len(env_var_value)
-        value_clean = value_clean.replace(env_var_value, replace_value)
-    return value_clean
+    return filter_out_values(value, env_var_values)
+
+
+def filter_out_config_values(log: str, config: Dict) -> str:
+    """
```
wangxiaoyou1993, line 24: For the method comment, let's use the standard format of the docstring.
Example: https://github.com/mage-ai/mage-ai/blob/master/mage_ai/io/base.py#L56-L67

mage_ai/shared/security.py
```diff
@@ -10,15 +11,33 @@
...
+    Used for integration pipeline logging security. Filters out values from
+    the passed in config.
+    """
+    if config is None:
```
wangxiaoyou1993, line 28: `if not log or not config`

mage_ai/shared/security.py
```diff
@@ -10,15 +11,33 @@
...
+        return log
+    values = config.values()
+    secret_values = [v for v in values if isinstance(v, str) and len(v) >= MIN_SECRET_LENGTH]
+    return filter_out_values(log, secret_values)
+
+
+def filter_out_values(log: str, values: List[str]) -> str:
+    if not values:
```
wangxiaoyou1993, line 36: `if not log or not values`

wangxiaoyou1993 (approved):

PR #2991, opened 2023-07-14
[xy] Fix dynamic block on k8s executor.

wangxiaoyou1993:
# Summary

Fix dynamic block on k8s executor. Not update the block status to complete without creating dynamic child block runs.

# Tests

tested locally

cc:


wangxiaoyou1993:
Fixing in another PR: https://github.com/mage-ai/mage-ai/pull/3035

PR #3000, opened 2023-07-17
[dy] Update secrets management backend

dy46:
# Summary

Update secrets management backend to handle multiple environments by adding a `key_uuid` to the Secret db model. The `key_uuid` field will be unique to each dev environment because it will be stored alongside the encryption key in the Mage data directory. Also, to support pipeline/project level secrets in the future, I added the concept of `Entity` to secrets. For pipeline level secrets, the encryption key and key_uuid will be stored in its own pipeline specific directory.

# Tests

tested locally with sqlite and postgres

cc: @wangxiaoyou1993


mage_ai/orchestration/db/models/__init__.py
```diff
@@ -0,0 +1,10 @@
+import enum
+
+
+class Entity(str, enum.Enum):
```
wangxiaoyou1993, line 4: why is there a duplicate definition of Entity?

mage_ai/orchestration/db/models/oauth.py
```diff
@@ -227,10 +228,10 @@ def get_parent_access(self, entity) -> int:
         This method is used when a role does not have a permission for a specified entity. Then,
         we will go up the entity chain to see if there are permissions for parent entities.
         '''
-        if entity == Permission.Entity.PIPELINE:
-            return self.get_access(Permission.Entity.PROJECT, get_project_uuid())
-        elif entity == Permission.Entity.PROJECT:
```
wangxiaoyou1993, line 232: Do you also want to remove the Entity definition in Permission class?

mage_ai/data_preparation/shared/secrets.py
```diff
@@ -1,38 +1,70 @@
...
+from mage_ai.orchestration.constants import Entity
 from mage_ai.settings.repo import get_data_dir, get_repo_path
 
 DEFAULT_MAGE_SECRETS_DIR = 'secrets'
 
 
-def get_secrets_dir():
-    return os.path.join(
-        get_data_dir(), DEFAULT_MAGE_SECRETS_DIR)
+def get_secrets_dir(
```
wangxiaoyou1993, line 13: Let's add comment to this method: The directory for storing the secret key.

mage_ai/data_preparation/shared/secrets.py
```diff
@@ -1,38 +1,70 @@
...
+    project_uuid: str = None,
+    pipeline_uuid: str = None,
+) -> str:
+    secrets_dir = os.path.abspath(os.path.join(get_data_dir(), DEFAULT_MAGE_SECRETS_DIR))
+
+    if entity == Entity.GLOBAL:
+        return secrets_dir
+    elif entity == Entity.PROJECT:
+        return os.path.join(secrets_dir, 'projects', project_uuid)
+    elif entity == Entity.PIPELINE:
```
wangxiaoyou1993, line 24: Would pipeline renaming break the secrets?

mage_ai/data_preparation/shared/secrets.py
```diff
@@ -75,38 +102,128 @@ def get_valid_secrets() -> List:
...
+                Secret.key_uuid == key_uuid,
+            ).one_or_none()
 
     if secret:
         secret.delete()
+    else:
+        print(f'WARNING: Could not find secret {name}')
+
+
+def _create_key_files(secrets_dir: str) -> Tuple[str, str]:
```
wangxiaoyou1993, line 178: Let's add comments to these methods

mage_ai/orchestration/db/migrations/versions/e3593cc2191e_add_key_uuid_to_secrets.py
```diff
@@ -0,0 +1,35 @@
...
+revision = 'e3593cc2191e'
+down_revision = 'dfe49d040487'
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('secret', schema=None) as batch_op:
+        batch_op.alter_column('name', existing_type=sa.String(length=255), unique=False, nullable=False)
```
wangxiaoyou1993, line 22: Do you check whether the unique constraint is dropped after migration (in sqlite and postgres)?
The unique constraint is created separately https://github.com/mage-ai/mage-ai/blob/master/mage_ai/orchestration/db/migrations/versions/1f9353eddbc6_add_secrets_table.py#L29

mage_ai/data_preparation/shared/secrets.py
```diff
@@ -41,22 +73,17 @@ def create_secret(name: str, value: str):
...
-
-def get_valid_secrets() -> List:
+def get_valid_secrets(
+    entity: Entity = Entity.GLOBAL,
+    pipeline_uuid: str = None,
+    project_uuid: str = None,
+) -> List:
     from mage_ai.orchestration.db.models.secrets import Secret
-    key = get_encryption_key()
+    key, _ = _get_encryption_key(
```
wangxiaoyou1993, line 82: Does this method still get secrets with the repo path?

mage_ai/orchestration/db/migrations/versions/e3593cc2191e_add_key_uuid_to_secrets.py
dy46: I did, it seemed to work this way. I tried to manually drop the constraint, but I always got an error doing it that way

mage_ai/data_preparation/shared/secrets.py
dy46: good point

mage_ai/orchestration/db/models/oauth.py
dy46: yes, let me remove that

mage_ai/data_preparation/shared/secrets.py
dy46: yes, I left it like this for now so that it still returns the same secrets that it did before this change. In the future, this method will need to be updated. I can add a comment

mage_ai/data_preparation/shared/secrets.py
wangxiaoyou1993: I don't have a good solution now. I think we can mention this issue in docs for now and ask user to manually rename the folder in prod.

wangxiaoyou1993 (approved):

wangxiaoyou1993 (approved):

Issue #3022, opened 2023-07-20
BigQuery Fails to Load Tables Created from Drive


mattppal:
**Describe the bug**
When reading from Google BigQuery, tables that are [created from Google Drive](https://www.optimizesmart.com/connect-and-transfer-data-from-google-sheets-to-bigquery/) will fail to read with the following error:

![image](https://github.com/mage-ai/mage-ai/assets/59450879/f7195872-f3af-4e46-8f8b-0c797777fb19)

This occurs even after adding the service account to the requisite sheet

![image](https://github.com/mage-ai/mage-ai/assets/59450879/db9a2271-b99b-40ac-99a0-499066b7039a)

**To Reproduce**
1. Load table into GBQ from Drive
2. Create service account and grant access
3. Attempt to query table from mage

mattppal:
Referenced [here](https://docs.getdbt.com/faqs/troubleshooting/access-gdrive-credential), the issue is that, when using the `oauth` method, additional scopes must be passed to the `service_account.Credentials` object. 

The dbt article uses the following scopes:

```
["https://www.googleapis.com/auth/userinfo.email",
            "https://www.googleapis.com/auth/cloud-platform",
            "https://www.googleapis.com/auth/sqlservice.login",
            "https://www.googleapis.com/auth/drive"]
```

When I supply these to `service_account.Credentials.from_service_account_info()` and `service_account.Credentials.from_service_account_file()`, the issue is resolved and the table loads

mattppal:
![image](https://github.com/mage-ai/mage-ai/assets/59450879/35b93dc6-dfaa-41bb-8561-7f6d1a679804)

PR #3023, opened 2023-07-20
[mp] add google oauth2 scopes to bigquery client

mattppal:
# Summary

Specify oauth2 scopes in Google BigQuery client to enable pulling from tables imported via Google Drive.

- [Scopes reference](https://developers.google.com/identity/protocols/oauth2/scopes)
- [dbt article](https://docs.getdbt.com/faqs/troubleshooting/access-gdrive-credential) with solution
- [Similar solution](https://stackoverflow.com/questions/53320537/creating-google-api-credentials-from-service-account-with-scope-and-delegated-a)

This closes https://github.com/mage-ai/mage-ai/issues/3022

image

image

# Tests
Locally with Mage Client.

### DIFF
--- a/mage_ai/io/bigquery.py
+++ b/mage_ai/io/bigquery.py
@@ -1,14 +1,16 @@
+from typing import Dict, List, Mapping, Union
+
 from google.cloud.bigquery import Client, LoadJobConfig, WriteDisposition
 from google.oauth2 import service_account
-from mage_ai.io.base import BaseSQLDatabase, ExportWritePolicy, QUERY_ROW_LIMIT
+from pandas import DataFrame
+
+from mage_ai.io.base import QUERY_ROW_LIMIT, BaseSQLDatabase, ExportWritePolicy
 from mage_ai.io.config import BaseConfigLoader, ConfigKey
 from mage_ai.io.export_utils import infer_dtypes
 from mage_ai.shared.utils import (
     convert_pandas_dtype_to_python_type,
     convert_python_type_to_bigquery_type,
 )
-from pandas import DataFrame
-from typing import Dict, List, Mapping, Union
 
 
 class BigQuery(BaseSQLDatabase):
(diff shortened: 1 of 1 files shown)


mage_ai/io/bigquery.py
```diff
@@ -37,36 +39,49 @@ def __init__(self, **kwargs) -> None:
...
-
-        credentials = kwargs.get('credentials')
+        if kwargs.get("verbose") is not None:
+            kwargs.pop("verbose")
+        super().__init__(verbose=kwargs.get("verbose", True))
+
+        CLIENT_SCOPES = [
+            "https://www.googleapis.com/auth/userinfo.email",
+            "https://www.googleapis.com/auth/cloud-platform",
+            "https://www.googleapis.com/auth/sqlservice.login",
```
wangxiaoyou1993: Is this scope needed?

mage_ai/io/bigquery.py
```diff
@@ -41,16 +43,28 @@ def __init__(self, **kwargs) -> None:
             kwargs.pop('verbose')
         super().__init__(verbose=kwargs.get('verbose', True))
 
+        CLIENT_SCOPES = [
+            'https://www.googleapis.com/auth/userinfo.email',
+            'https://www.googleapis.com/auth/cloud-platform',
+            'https://www.googleapis.com/auth/sqlservice.login',
+            'https://www.googleapis.com/auth/drive'
+
```
wangxiaoyou1993, line 51: nit: remove empty line

mage_ai/io/bigquery.py
mattppal: https://cloud.google.com/bigquery/docs/external-data-drive#python

Looks like we might only need drive & bq. I borrowed these from dbt. I'll test

mattppal: Confirmed this works with the following scopes, which match googles docs

```python
CLIENT_SCOPES = [
            "https://www.googleapis.com/auth/drive",
            "https://www.googleapis.com/auth/bigquery",
        ]
```

mattppal: I would say I'm marginally concerned this could break something if a google drive api isn't enabled

wangxiaoyou1993 (approved):

PR #3032, opened 2023-07-20
[mp] Add service argument to OracleDB data loader

mattppal:
# Summary

Our OracleDB [data integration](https://github.com/mage-ai/mage-ai/blob/master/mage_integrations/mage_integrations/sources/oracledb/README.md) accepts 5 arguments, which are necessary for connection:

- `host`
- `port`
- `service`
- `user`
- `password`

Our `io` config currently only accepts four. 

image

Given that we're using the `python-oracledb` package, we need [five](https://python-oracledb.readthedocs.io/en/latest/user_guide/connection_handling.html#creating-a-standalone-connection) to connect.

```python
import os

userpwd = os.environ.get("PYTHON_PASSWORD")

connection = oracledb.connect(user="hr", password=userpwd,
                              host="localhost", port=1521, service_name="orclpdb")
```

This pulls through the fifth argument, `service_name`, and allows the user to input via `io_config.yaml`

# Tests

cc:


mage_ai/io/oracledb.py
```diff
@@ -24,7 +32,8 @@ def with_config(cls, config: BaseConfigLoader) -> 'OracleDB':
             user=config[ConfigKey.ORACLEDB_USER],
             password=config[ConfigKey.ORACLEDB_PASSWORD],
             host=config[ConfigKey.ORACLEDB_HOST],
-            port=config[ConfigKey.ORACLEDB_PORT]
+            port=config[ConfigKey.ORACLEDB_PORT],
+            service=config[ConfigKey.ORACLEDB_SERVICE]
```
wangxiaoyou1993, line 36: Did you test the case when the service is None? Just want to make sure both cases work.

mattppal: I have not tested yet, that's a todo

mage_ai/io/oracledb.py
```diff
@@ -10,11 +10,19 @@
 
 
 class OracleDB(BaseSQL):
-    def __init__(self, user, password, host, port, verbose: bool = False, **kwargs) -> None:
+    def __init__(self,
+                 user,
+                 password,
+                 host,
+                 port,
+                 service,
```
Luishfs, line 18: Could you change service to service_name? That's how [python-oracledb](https://python-oracledb.readthedocs.io/en/latest/user_guide/connection_handling.html) expects this parameter name

Screenshot from 2023-07-24 17-47-47

Luishfs, line 18: Could you change service to service_name? That's how [python-oracledb](https://python-oracledb.readthedocs.io/en/latest/user_guide/connection_handling.html) expects this parameter name

Screenshot from 2023-07-24 17-47-47

mage_ai/io/oracledb.py
Luishfs: Just tested it, by adding SERVICE_NAME as null ( or just not declaring it), code run as usual
Screenshot from 2023-07-24 17-50-10
Screenshot from 2023-07-24 17-52-43

mage_ai/io/oracledb.py
mattppal: nice catch, thanks @Luishfs !

PR #3039, opened 2023-07-20
[td] Limit number of files returned from Git Custom Branch API

tommydangerous:
# Summary
There is an edge case where 17K file names are being returned from the API response.

This causes too much data to be transferred over the network and slows down the entire app.

### DIFF
--- a/mage_ai/api/presenters/GitCustomBranchPresenter.py
+++ b/mage_ai/api/presenters/GitCustomBranchPresenter.py
@@ -21,6 +21,15 @@ class GitCustomBranchPresenter(GitBranchPresenter):
 
         data_to_display = self.model
 
+        for key in [
+            'modified_files',
+            'staged_files',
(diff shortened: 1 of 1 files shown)


mage_ai/api/presenters/GitCustomBranchPresenter.py
```diff
@@ -21,6 +21,15 @@ def present(self, **kwargs):
...
         data_to_display = self.model
 
+        for key in [
+            'modified_files',
+            'staged_files',
+            'untracked_files',
+        ]:
+            arr = data_to_display.get(key)
+            if arr and type(arr) is list and len(arr) >= 1:
+                data_to_display[key] = arr[:100]
```
wangxiaoyou1993, line 31: This might not only be an issue in presenter.
I wonder whether the resource takes a long time to get the file list. If so, we'll limit the time in the resource.

PR #3045, opened 2023-07-21
fix: add missing rules for k8s Role

gabriel-milan:
# Summary
When working with [Workspace management](https://docs.mage.ai/developing-in-the-cloud/workspaces/overview), the Helm chart generated role misses a few rules, which led to this:

image

Adding these rules grants enough authorization for the service account to manage the resources it needs.

# Tests
I've tested in my current toy deployment.

### DIFF
--- a/kube/app.yaml
+++ b/kube/app.yaml
@@ -13,8 +13,14 @@ rules:
 - apiGroups: [""] # "" indicates the core API group
   resources: ["pods"]
   verbs: ["get", "list", "watch"]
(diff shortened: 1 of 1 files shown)


wangxiaoyou1993:
cc: @dy46

dy46 (commented):
can we create a new role and role binding for these workspace permissions?

gabriel-milan:
> can we create a new role and role binding for these workspace permissions?

maybe that's a better option, yes

PR #3051, opened 2023-07-21
[dy] Add timeout for fetching files for git

dy46:
# Summary

Added a 10 second timeout for fetching untracked and staged files. The modified files call is a bit trickier to add a timeout for, but I think the main issue is the untracked files anyway.

# Tests

tested locally

cc:


wangxiaoyou1993:
Did you test creating 10k folders locally and checking the latency of the request?

dy46:
@wangxiaoyou1993 I didn't try with 10000 folders, but I did test that it times out correctly after 10 seconds.

wangxiaoyou1993:
Can you try creating 10k folders locally and see if Mage app can function correctly with it?

dy46:
@wangxiaoyou1993 I created the folders, and everything seemed ok. The latency of the request is still around 10 seconds

wangxiaoyou1993 (approved):

PR #3052, opened 2023-07-21
[jk] Default error log to Errors tab when opening log detail

johnson-mage:
# Summary
- When user clicks on a log row from the Logs table, default to opening the Errors tab in the Log Detail instead of the Details tab if the log is an error log.

# Tests
open error log

### DIFF
--- a/mage_ai/frontend/components/DependencyGraph/BlockNode/index.tsx
+++ b/mage_ai/frontend/components/DependencyGraph/BlockNode/index.tsx
@@ -205,7 +205,7 @@ function BlockNode({
             </Flex>
           </Flex>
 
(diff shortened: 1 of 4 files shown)


mage_ai/frontend/components/DependencyGraph/BlockNode/index.tsx
```diff
@@ -205,7 +205,7 @@ function BlockNode({
             </Flex>
           </Flex>
 
-          <Spacing mr={HEADER_SPACING_HORIZONTAL_UNITS} />
+          <Spacing mr="15px" />
```
johnson-mage, line 208: Decreased to 15px (instead of 16px) to fix styling that cuts off text when screen is zoomed out (e.g. to 90% on certain screen displays).

PR #3075, opened 2023-07-25
Use Playwright for web UI test automation

erictse:
# Summary
This modifies the GitHub Action CI so that instead of just building the frontend, it uses the Playwright web test framework to run integration tests on the production UI served statically from the backend.

It currently runs on each flavor of Python which may not be necessary. It also only runs on Chromium, but we could easily adopt WebKit and Firefox as desired.

I really wanted to package actually useful tests, but grokking the React component library to instrument the UI was delaying this PR too long, when someone else with more familiarity might be able to run with it.

# Tests
I pushed my branch to my forked repo on GitHub and a simple home page test passed.

cc: @wangxiaoyou1993 @tommydangerous


wangxiaoyou1993:
Thanks for putting up the PR and sharing this UI testing framework.
@johnson-mage Could you take a look at it?

johnson-mage:
@erictse Thank you for setting this up! I think we can use this as a framework for adding tests on the UI. As you alluded to in the summary, we can just run the playwright tests one time, instead of on every Python version.

Can you add another job (e.g. `test_backend_3.10_and_frontend`) that runs the backend tests on Python 3.10 and runs the Playwright tests there with the additional yml steps as needed? Or split up the jobs another way as you see fit to only set up and run the Playwright tests once. In this separate job, I believe this step (`- uses: actions/checkout@v3`) only needs to be executed once.

johnson-mage:
@erictse feel free to tag me or re-request review when you'd like me to take another look at this PR

erictse:
@johnson-mage Not ready yet, but I'll let you know.

erictse:
@johnson-mage Ready to go!

I did not opt to set up a separate job for frontend tests since I wanted to avoid code duplication where we run into trouble with changes not occurring in both jobs. Instead, I just have the Python 3.10 variant executing the web UI tests.

When the tests are run locally (such as by the VS Code Playwright plugin) or using `yarn playwright test`, Playwright will spin up a dev server. The GitHub Action will solely rely on the Python server to serve static frontend code, since it is closest to production.

I wrote a couple Playwright tests to verify that Mage pipeline creation worked. At first, they were mysteriously failing, but using Playwright's tracing ability, I was able to narrow it down to the usage statistics modal dialog, which I decided would be most cleanly dealt with when project is created in the CI.

.github/workflows/build_and_test.yml
```diff
@@ -46,16 +46,33 @@ jobs:
...
+          yarn install &&
+          yarn export_prod &&
+          yarn playwright install chromium &&
+          yarn playwright test -c playwright.config.ci.ts
+        working-directory: mage_ai/frontend/
+        env:
+          PYTHONPATH: ${{ env.pythonLocation }}:. # Playwright tests on CI depend on the Python web server
+        if: matrix.python-version == '3.10'
+      - uses: actions/upload-artifact@v3
+        if: always()
```
johnson-mage, line 74: Does this step need to be executed for every python version? Seems like we can just run it for 3.10 since running it for the other versions doesn't do anything (this was the output: `No files were found with the provided path: mage_ai/frontend/playwright-report/. No artifacts will be uploaded.`).

johnson-mage: Everything else looks good, thanks Eric!

erictse: No, it's true that it should be for 3.10, but this approach uses `always()` which forces this step to be run, even if previous steps failed and the job is cut short. Do you recommend something else?

johnson-mage (approved):

.github/workflows/build_and_test.yml
johnson-mage: @erictse That seems fine then since running that step is innocuous.

PR #3096, opened 2023-07-26
[dy] Cancel block runs when pipeline run fails

dy46:
# Summary

For pipeline runs that don't have `allow blocks to fail` enabled, we will cancel the in progress block runs if the pipeline run fails.

# Tests

tested locally

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/orchestration/pipeline_scheduler.py
+++ b/mage_ai/orchestration/pipeline_scheduler.py
@@ -1164,7 +1164,10 @@ def stop_pipeline_run(
     cancel_block_runs_and_jobs(pipeline_run, pipeline)
 
 
-def cancel_block_runs_and_jobs(pipeline_run, pipeline):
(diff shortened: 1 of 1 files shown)


mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -1159,6 +1161,24 @@ def stop_pipeline_run(
     pipeline_run.update(status=PipelineRun.PipelineRunStatus.CANCELLED)
 
     # Cancel all the block runs
+    cancel_block_runs_and_jobs(pipeline_run, pipeline)
+
+
+def cancel_block_runs_and_jobs(pipeline_run, pipeline):
```
wangxiaoyou1993, line 1167: nit: Let's add types to the arguments

wangxiaoyou1993 (approved):

PR #3099, opened 2023-07-26
[jk] Clean up code in VariableResource method

johnson-mage:
# Summary
- See title.

# Tests
- N/A


mage_ai/api/resources/VariableResource.py
```diff
@@ -66,6 +63,9 @@ def collection(self, query, meta, user, **kwargs):
         ]
         variables = global_variables_arr
         if not global_only:
+            # Get global variables from project's path
```
wangxiaoyou1993, line 66: no need to move the comment here

PR #3102, opened 2023-07-26
[dy] Batch git sync in pipeline_scheduler

dy46:
# Summary

Currently, git syncs are run every time a pipeline run is started. However, it is possible for many pipeline runs to get created at the same time which could start a large number of git syncs which can cause problems. I changed the logic in the `pipeline_scheduler` to only perform 1 git sync for all active pipeline schedules.

This can be an issue for event and api triggers as well, but I'll fix that in the future since it's a more difficult problem to solve.

# Tests

tested locally

cc: @wangxiaoyou1993


mage_ai/orchestration/triggers/api.py
```diff
@@ -28,6 +30,13 @@ def create_and_start_pipeline_run(
         payload,
     )
 
+    # TODO: make sure git syncs are not run concurrently if there are
```
wangxiaoyou1993, line 33: This could actually be an issue. Let's not run git sync in the API call.

dy46: ok i'll remove it

PR #3103, opened 2023-07-26
[xh]Generate pipeline(multiple blocks) template from description

matrixstone:
# Summary
It firstly identifies how many Blocks are required and the function of each Block.
Then it generate block template asynchronizly for each block.
Eventually it returns a dictionary of Block with Block Id telling the dependency.

This PR also prefix class name before enum value in TEMPLATE_CLASSIFICATION_FUNCTION definition. Because I found issue sometime LLM does not return one of the value in enum list. This fix helps increase the probability of enum value.

# Tests
Test samples recorded in the spreadsheet: https://docs.google.com/spreadsheets/d/1vp0UB6EqM3vnY4f0hYV0nB36X7ngiRIVZZU4FFcybH0/edit#gid=1933961353

cc:
@wangxiaoyou1993


mage_ai/ai/llm_pipeline_wizard.py
```diff
@@ -154,16 +196,56 @@ async def async_generate_block_with_description(self, block_description: str) ->
...
                 function_args)
-            return fetch_template_source(
+            block_template = fetch_template_source(
                 block_type=block_type,
                 config=config,
                 language=block_language,
                 pipeline_type=pipeline_type,
             )
+            block_uuid = uuid.uuid4().hex
+            return Block(name=block_uuid,
```
wangxiaoyou1993, line 206: You don't need to create the block object here. The block object can't be returned as API response. Just return a dictionary of the config.

matrixstone:
Thank you for the review Xiaoyou! Rebased master so we are returning multiple block configurations in this PR. PTAL @wangxiaoyou1993

wangxiaoyou1993 (approved):

PR #3109, opened 2023-07-27
[td] Pipeline landing times

tommydangerous:
# Summary
- Schedule pipeline runs and start it based on landing times

### DIFF
--- a/mage_ai/orchestration/db/models/schedules.py
+++ b/mage_ai/orchestration/db/models/schedules.py
@@ -5,6 +5,7 @@ import traceback
 import uuid
 from datetime import datetime, timedelta, timezone
 from dateutil.relativedelta import relativedelta
+from statistics import stdev
 from typing import Dict, List
 
 import pytz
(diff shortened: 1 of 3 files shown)


mage_ai/orchestration/db/models/schedules.py
```diff
@@ -212,16 +251,100 @@ def should_schedule(self) -> bool:
...
                 ) == 0,
                 self.pipeline_runs
             ):
-                return True
+                if self.landing_time_enabled():
+                    if len(previous_runtimes) == 0:
+                        return True
+                    else:
+                        runtime = ceil(sum(previous_runtimes) / len(previous_runtimes))
+                        sd = ceil(stdev(previous_runtimes))
```
wangxiaoyou1993, line 268: I think it'll throw error when there's only one item in previous_runtimes.

mage_ai/orchestration/db/models/schedules.py
```diff
@@ -165,36 +167,73 @@ def create_or_update(self, trigger_config: Trigger):
...
-            return now.replace(second=0, microsecond=0, minute=0, hour=0, day=1)
+            current_execution_date = now.replace(second=0, microsecond=0, minute=0, hour=0, day=1)
         else:
             cron_itr = croniter(self.schedule_interval, now)
-            return cron_itr.get_prev(datetime)
+            current_execution_date = cron_itr.get_prev(datetime)
+
+        if self.landing_time_enabled() and self.start_time:
+            if self.schedule_interval == '@hourly':
+                current_execution_date = current_execution_date.replace(
```
wangxiaoyou1993, line 192: I'm confused by the logic here

mage_ai/orchestration/db/models/schedules.py
```diff
@@ -212,16 +251,100 @@ def should_schedule(self) -> bool:
...
                 self.pipeline_runs
             ):
-                return True
+                if self.landing_time_enabled():
+                    if len(previous_runtimes) == 0:
+                        return True
+                    else:
+                        runtime = ceil(sum(previous_runtimes) / len(previous_runtimes))
+                        sd = ceil(stdev(previous_runtimes))
+                        if datetime.now(timezone.utc) >= current_execution_date - timedelta(
```
wangxiaoyou1993, line 269: I think it could cause duplicate pipeline runs being scheduled.

mage_ai/orchestration/db/models/schedules.py
tommydangerous: fixed

mage_ai/orchestration/db/models/schedules.py
tommydangerous: I’ll add comments.

mage_ai/orchestration/db/models/schedules.py
tommydangerous: I don’t think so unless we have 2 schedulers.

PR #3110, opened 2023-07-27
[dy] Import block function from file for test execution

dy46:
# Summary

Updated how python blocks are executed when run from the notebook. Instead of running the block function directly from the file, we will import the block function and then execute it so that the error logging is more readable. 

I also changed the `test_execution` parameter to `from_notebook` to be consistent across the Block functions.

# Tests

tested locally with different blocks

examples:

Screenshot 2023-07-28 at 11 57 58 AM
Screenshot 2023-07-28 at 11 57 47 AM

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/server/websocket_server.py
+++ b/mage_ai/server/websocket_server.py
@@ -325,7 +325,7 @@ class WebSocketServer(tornado.websocket.WebSocketHandler):
 
         error = message.get('error')
         if error:
-            message['data'] = cls.format_error(error)
+            message['data'] = cls.format_error(error, block_uuid=block_uuid)
 
         output_dict = dict(
             block_type=block_type,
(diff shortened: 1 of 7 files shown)


mage_ai/server/websocket_server.py
```diff
@@ -540,25 +544,33 @@ async def check_for_messages():
             task = asyncio.create_task(check_for_messages())
             set_current_message_task(task)
 
-    def __format_error(error: List[str]) -> List[str]:
+    @classmethod
+    def format_error(cls, error: List[str], block_uuid: str = None) -> List[str]:
```
wangxiaoyou1993, line 548: Can you add the comment to this method? It's a bit hard to understand the logic here.

wangxiaoyou1993 (approved):

mage_ai/server/websocket_server.py
```diff
@@ -546,16 +546,51 @@ async def check_for_messages():
...
+        code should be executed within the `{block_uuid}.py` file. Otherwise, we will default to
+        using `execute_block_function` which is a method called when a block is executed.
+
+        The `custom_block_end_regex` pattern is used as a fallback if the `end_regex` pattern cannot
+        be found. If a block is not a standard Python block, this fallback regex will be used.
+
+        All the lines in the error between the initial_idx and end_idx (or custom_block_end_idx)
+        will be filtered out in the return value.
+
+        :param error: The list of strings representing the error message, where each element
```
wangxiaoyou1993, line 570: @dy46 Can you follow our docstring format?

PR #3115, opened 2023-07-28
{orchestration] check_status method bug fix to look at pipelines ran between specific time period

sumanshusamarora:
# Summary
There was a bug with check_status method in orchestration while compared exact execution_datetime with pipeline run datetime which is not the purpose of the method. Rather it should check if pipeline was ran within a specific time period. Hence included another parameter hours. Now the method checks if pipeline had a successful run between execution_date - hours and execution_date

# Tests
Tested locally

cc:


wangxiaoyou1993 (approved):

wangxiaoyou1993:
Thanks for fixing this issue!

PR #3142, opened 2023-07-31
[dy] Only try to interpolate variables if they are in the query

dy46:
# Summary

For sql blocks, we try to interpolate the query even if there are no interpolated variables in the block. I added a regex to check if there is the `{{ df_X }}` syntax in the query before trying to interpolate variables.

I also removed the git sync from api triggers because it can potentially cause issues.

# Tests

tested locally

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/server/server.py
+++ b/mage_ai/server/server.py
@@ -256,7 +256,7 @@ async def main(
         print('User authentication is enabled.')
         # We need to sleep for a few seconds after creating all the tables or else there
         # may be an error trying to create users.
-        sleep(3)
+        sleep(5)
 
         # Create new roles on existing users. This should only need to be run once.
         Role.create_default_roles()
--- a/mage_ai/data_preparation/models/block/sql/utils/shared.py
+++ b/mage_ai/data_preparation/models/block/sql/utils/shared.py
@@ -110,11 +110,11 @@ def interpolate_input(
 
         return '.'.join(list(filter(lambda x: x, [db, schema, tn])))
 
-    if not has_interpolated_inputs(query):
-        return query
-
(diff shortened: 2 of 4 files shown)


wangxiaoyou1993:
What bug does it fix?

dy46:
@wangxiaoyou1993 this one: https://mageai.slack.com/archives/C03HTTWFEKE/p1690478813984429

> I have a desire to run a custom Raw SQL block for one data source (Postgres) after another custom Raw SQL block with a datasource (trino) where there is no dependency on the outputs of the previous block. Mage is telling me that it can not run my second block due to being unable interpolate one output into the other...even though im not actually using the output of the previous block anywhere. Any idea on how to make this happy?

dy46:
actually i think it would be better to check each variable since that could cover more cases.

wangxiaoyou1993:
> actually i think it would be better to check each variable since that could cover more cases.

Yes, could you check in the for loop?

dy46:
@wangxiaoyou1993 moved into the for loop

wangxiaoyou1993 (approved):

mage_ai/server/server.py
```diff
@@ -256,7 +256,7 @@ async def main(
         print('User authentication is enabled.')
         # We need to sleep for a few seconds after creating all the tables or else there
         # may be an error trying to create users.
-        sleep(3)
+        sleep(5)
```
wangxiaoyou1993, line 259: Is this an intentional change?

dy46: yeah i wanted to give it a little longer before creating the default roles and users

PR #3161, opened 2023-08-02
[xh]Generate comment for code

matrixstone:
# Summary
Generate comment for each function inside the code.
The newly generated comment will be added as a doc string beneath function definition.
Also it can merge into existing doc string if there is error.

Current issues are (1) the output code is not well formatted (2) # comment misses

# Tests
Command to run the code:
`python mage_ai/ai/generator_cmds.py generate-comment-for-block code_file`

code_file is the file contains the code to generate comment for.

Reviewer:
@wangxiaoyou1993

### DIFF
--- a/mage_ai/ai/llm_pipeline_wizard.py
+++ b/mage_ai/ai/llm_pipeline_wizard.py
@@ -1,3 +1,4 @@
+import ast
 import asyncio
 import json
 import os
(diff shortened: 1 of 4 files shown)


mage_ai/ai/llm_pipeline_wizard.py
```diff
@@ -247,6 +257,42 @@ async def async_generate_pipeline_with_description(self, pipeline_description: s
...
+        prompt_template = PromptTemplate(
+            input_variables=[
+                'block_content',
+            ],
+            template=PROMPT_FOR_FUNCTION_COMMENT,
+        )
+        chain = LLMChain(llm=self.llm, prompt=prompt_template)
+        function_comments = await chain.arun(block_content=block_content)
+        function_comments_json = json.loads(function_comments)
+        return self.__insert_comments_in_functions(block_content, function_comments_json)
```
dy46, line 294: i think the variable names here should be flipped. function_comments here is the json string, and function_comments_json is the object

mage_ai/ai/llm_pipeline_wizard.py
```diff
@@ -247,6 +257,42 @@ async def async_generate_pipeline_with_description(self, pipeline_description: s
         await asyncio.gather(*block_tasks)
         return blocks
 
+    def __insert_comments_in_functions(self, code: str, function_comments_json: dict):
```
dy46, line 260: let's use `Dict` instead of `dict` here

matrixstone:
Addressed the comments. Thank you for the review David! @dy46

dy46 (approved):
LGTM. are we going to try to improve the comments later?

PR #3168, opened 2023-08-02
[dy] Fix circular dependency and repo config issues

dy46:
# Summary

Fixes https://mageai.slack.com/archives/C03HTTWFEKE/p1690993959608629

This PR does a couple of things:
1. Removes `pipeline.py` dependency on `db_connection`.
2. Initializes all the `RepoConfig` variables so that if there is an issue rendering the yaml file, it doesn't cause the entire app to crash.
3. Populate project_uuid directly from the `metadata.yaml` file instead of through the RepoConfig.

# Tests

tested locally

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/data_preparation/repo_manager.py
+++ b/mage_ai/data_preparation/repo_manager.py
@@ -27,8 +27,16 @@ class ProjectType(str, Enum):
     STANDALONE = 'standalone'
 
 
+def get_metadata_path():
+    return os.path.join(get_repo_path_new(), 'metadata.yaml')
+
+
 class RepoConfig:
-    def __init__(self, repo_path: str = None, config_dict: Dict = None):
+    def __init__(
+        self,
+        repo_path: str = None,
+        config_dict: Dict = None,
+    ):
         self.repo_path = repo_path or get_repo_path_new()
         self.repo_name = os.path.basename(self.repo_path)
         self.project_uuid = None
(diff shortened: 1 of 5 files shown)


mage_ai/data_preparation/repo_manager.py
```diff
@@ -212,20 +235,20 @@ def get_variables_dir(repo_path: str = None) -> str:
     return get_repo_config(repo_path=repo_path).variables_dir
 
 
-config = get_repo_config()
 project_uuid = None
 try:
-    project_uuid = config.project_uuid
+    with get_repo_config().metadata_path as f:
```
wangxiaoyou1993, line 240: It's a bit confusing here. Why can't we use `config.project_uuid`?

dy46: @wangxiaoyou1993 the problem is that the `RepoConfig` is read before the server is finished starting. If a user tries to use a secret variable in the project metadata, getting the `RepoConfig` will fail and `config.project_uuid` will return `None` because the db connection session may not be initialized yet.

dy46: The `RepoConfig` being able to use the template vars causes a lot of these dependency issues.

dy46: It doesn't directly resolve that circular dependency, and I didn't see any errors for that dependency. It would probalby be a good idea to fix that though, so I'll look into it

dy46:
@wangxiaoyou1993 I updated the PR to remove that dependency. Basically, I have to assume that the user will not use a `mage_secrets_var` for the `variables_dir` config because the db_connection needs the `variables_dir` to initialize. I changed the `get_variables_dir` method to not render the template with any functions that require a db connection.

mage_ai/data_preparation/repo_manager.py
```diff
@@ -41,30 +76,15 @@ def __init__(self, repo_path: str = None, config_dict: Dict = None):
...
-            # Priority:
-            # 1. 'variables_dir' from config_dict
-            # 1. os.getenv(MAGE_DATA_DIR_ENV_VAR)
-            # 2. 'variables_dir' from project's metadata.yaml file
-            # 3. DEFAULT_MAGE_DATA_DIR
-            if config_dict is not None and config_dict.get('variables_dir'):
-                self.variables_dir = config_dict.get('variables_dir')
-            elif os.getenv(MAGE_DATA_DIR_ENV_VAR):
-                self.variables_dir = os.getenv(MAGE_DATA_DIR_ENV_VAR)
+            if config_dict and config_dict.get('variables_dir'):
```
wangxiaoyou1993, line 79: why don't you put this inside the `get_variables_dir` method?

dy46: because this seems like a different flow (where the config_dict is directly passed into the `RepoConfig` object). It seems separate from fetching the repo config from the project `metadata.yaml` file.

wangxiaoyou1993: but there's duplicate logic in the `get_variables_dir`
```
    elif repo_config is not None:
        variables_dir = repo_config.get('variables_dir')
```

dy46: oh that's supposed to be different. The `repo_config` in `get_variables_dir` is the config from the project `metadata.yaml`. 

I only added the `repo_config` parameter so that if the caller method already fetched the project `metadata.yaml`, it could just pass it in to `get_variables_dir` so that it doesn't try to read from the `metadata.yaml` again.

wangxiaoyou1993 (approved):

PR #3174, opened 2023-08-03
jventura/disableTerminal

juancaven1988:
# Summary

Create a new ENV variable called "DISABLE_TERMINAL" and a function that checks if that env is equal or greater than 1
the when a solicited to access terminal is received send an unauthorized message if this is true.

# Tests

cc:

@dy46


mage_ai/settings/__init__.py
```diff
@@ -19,10 +19,18 @@
...
 
+
+
 def is_disable_pipeline_edit_access():
     return DISABLE_NOTEBOOK_EDIT_ACCESS >= 1
 
 
+#------------------------- DISABLE TERMINAL ----------------------#'
+
+DISABLE_TERMINAL = 1 if os.environ.get("DISABLE_TERMINAL") == "1" else 0
```
dy46, line 30: you can use `os.getenv('DISABLE_TERMINAL')`

dy46: You also don't need the `is_terminal_access_disable()` method for this variable. You can just fetch the value of the variable.

juancaven1988: I change os.environt to getenv(). 
Remove the function, but in the condition still ask if the variable is 1 so it only applies with that value and ignore in other cases.

dy46: I would change the definition to how it is for the `REQUIRE_USER_AUTHENTICATION`, so that you can just check if the variable is true or false.

juancaven1988: I made the suggested changes

dy46 (approved):
looks good!

PR #3189, opened 2023-08-04
[dy] Escape password in db_connection_url

dy46:
# Summary

We currently don't escape characters in the db_connection_url, but the sqlalchemy and alembic docs do suggest doing so. I also ran into an error when I had a special character in the password which caused the alembic migrations to fail.

Docs:
https://docs.sqlalchemy.org/en/14/core/engines.html#database-urls
https://alembic.sqlalchemy.org/en/latest/api/config.html#alembic.config.Config.set_main_option

# Tests

tested locally with postgres and sqlite

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/orchestration/db/__init__.py
+++ b/mage_ai/orchestration/db/__init__.py
@@ -59,6 +59,7 @@ try:
     )
     engine.connect()
 except SQLAlchemyError:
+    engine.dispose()
     url_parsed = urlparse(db_connection_url)
     if url_parsed.password:
         db_connection_url = db_connection_url.replace(
--- a/mage_ai/orchestration/db/__init__.py
+++ b/mage_ai/orchestration/db/__init__.py
@@ -1,6 +1,6 @@
 import logging
 import os
-from urllib.parse import parse_qs, urlparse
+from urllib.parse import parse_qs, quote_plus, urlparse
 
 import sqlalchemy
 from sqlalchemy import create_engine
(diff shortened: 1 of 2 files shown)


wangxiaoyou1993:
Could you give an example what db_connection_url has the issue?

dy46:
soemthing like this
postgresql+psycopg2://postgres:abc%3def$123@xxx.us-west-2.rds.amazonaws.com:5432/postgres

wangxiaoyou1993:
> soemthing like this postgresql+psycopg2://postgres:abc%3def$[123@xxx.us-west-2.rds.amazonaws.com](mailto:123@xxx.us-west-2.rds.amazonaws.com):5432/postgres

Thanks. Just wondering, could it break things if user already provides an escaped url in the db_connection_url?

dy46:
that is a good point

dy46:
@wangxiaoyou1993 updated the PR to escape the strings only when certain exceptions are raised

mage_ai/orchestration/db/__init__.py
```diff
@@ -51,10 +52,25 @@
...
+    engine.connect()
+except SQLAlchemyError:
+    engine.dispose()
+    url_parsed = urlparse(db_connection_url)
+    if url_parsed.password:
+        db_connection_url = db_connection_url.replace(
+            url_parsed.password,
+            quote_plus(url_parsed.password),
+        )
+    engine = create_engine(
```
wangxiaoyou1993, line 69: If there's no url_parsed.password, we can directly throw exception, right?

wangxiaoyou1993 (approved):

mage_ai/orchestration/db/__init__.py
dy46: oops i didn't see this, i'll add it as a follow up

PR #3193, opened 2023-08-05
[hw] Create the destination database if it does not exist when exporting

csharplus:
# Summary

Exception was thrown when the destination database does not exist when using ClickHouse as a data exporter.

# Tests

Manually tested with a local development environment.
cc:

@dy46 @wangxiaoyou1993


csharplus:
@wangxiaoyou1993 Please review this PR and see if it is safe to merge.

Issue #2742, opened 2023-06-01
Renaming block breaks for the block shared in multiple pieplines


wangxiaoyou1993:
Original thread:

> Hello Mage Team :wave: :mage:

`mage v0.8.86`

I've been using mage extensively for weeks now, still love it !
I enjoy reusing blocks around from different pipelines, however the experience could probably be improved
• bug: if I reuse a block from pipeline_1 in pipeline_2, rename it in pipeline_2 -&gt; pipeline_1 breaks :broken_heart: 
    ◦ either rename both
    ◦ or rename only the second one but maybe duplicate the file ?
    ◦ or have a prompt for either of those options
• improvement: maybe indicate on the block that it is shared ? I tend to forget it is used across several pipelines, change something and then boom unexpected results in other pipelines
    ◦  display `shared` just like you display more info for `replicated` blocks ? 
    ◦ prevent other pipelines from changing the code in the original block ? Just like what is done for `replicated` blocks ? 
    ◦ or when changing a block, ask if the changes should be made everywhere or the file duplicated + renamed ? 
Thanks in advance !

[Slack Message](https://mageai.slack.com/archives/C03HTTWFEKE/p1685635573544499?thread_ts=1685635573.544499&cid=C03HTTWFEKE)

PR #3199, opened 2023-08-07
[jk] Renaming blocks with shared pipelines

johnson-mage:
# Summary
- Prevent users from creating a new block with the same name as an existing block if the existing block is used in a pipeline.
- Update UI to more visibly indicate blocks shared by pipelines and warn users before updating a shared block name.
- Misc UI bufgixes

Related to https://github.com/mage-ai/mage-ai/issues/2742

# Tests
Renaming blocks (users can rename blocks from 1.  block header or 2. block settings in sidekick):
block renaming
image

Error thrown if user tries to create new block with existing uuid:
creating new block with same name

### TODOs for separate PRs
- Add block "Detach" feature, which lets user duplicate a block and remove any associations with other pipelines (would require to choose a new, unique block name).
- Add setting for user to update the `metadata.yaml` configs for all shared pipelines when renaming a block, so instead of getting a "Block file no longer exists" message in the shared pipelines, those pipelines would automatically get updated with the renamed block.


mage_ai/api/resources/BlockResource.py
```diff
@@ -36,6 +36,15 @@ async def create(self, payload, user, **kwargs):
         language = payload.get('language')
         name = payload.get('name')
         block_name = name or payload.get('uuid')
+        require_unique_name = payload.get('require_unique_name')
+
+        cache = await BlockCache.initialize_cache()
+        if require_unique_name:
```
wangxiaoyou1993, line 42: If a block exists, we already prevent creating the block: https://github.com/mage-ai/mage-ai/blob/master/mage_ai/data_preparation/models/block/__init__.py#L529-L535

johnson-mage: @wangxiaoyou1993 I did see that code, but there were at least 3 scenarios of adding a block i was testing for when there is already an existing block file with a block in a pipeline:
1) Different pipeline does not have the block and user tries to create new block with existing block name. [not covered by existing code, which is why I added the `require_unique_name` if block]
2) Different pipeline has the shared block already, but user tries to add same block by dragging from file browser. [covered by existing code]
3) Different pipeline does not have the shared block, and user adds it by dragging from the file browser.

With this addition block payload prop (`require_unique_name`), I updated the UI so that users are no longer allowed to  create a new block with an existing block name if the block exists in the cache (i.e. is used by another pipeline). If it is NOT used by another pipeline (e.g. if the block file exists), users can still create a new block with the existing block name and it would use the contents of the existing block file.

Is there a better way to handle all those scenarios above?

wangxiaoyou1993: Got it. I think we should always not allow creating a block with an existing name. So you can just simply check whether the file path exists.

johnson-mage: @wangxiaoyou1993 I was testing with that scenario (where we always prevent creating new block with existing name for example by checking if the block file path exists), but it felt a bit restrictive since it prevented users from being able to drag an existing block from the file browser (or would require more BE changes) since that would technically create a new block with existing name even tho it might be the user's intention.

For example, someone might have sent a user a block file so that block may not be in any pipelines already, but we don't want to prevent them from being able to create a new block with that block file (either by drag and drop or by creating a new block with the name of the block file) when that block name isn't used in any pipelines yet.

johnson-mage: The `require_unique_name` prop gave the FE flexibility to handle all the diff scenarios for what I thought made sense.

johnson-mage: @wangxiaoyou1993 After discussing with Kai, he suggested a better flow that also simplifies the changes on the BE. If a user tries to create a new block with an existing block name and type, the UI will throw an error (even if the existing block does not belong to any pipelines) but include a link to open the existing block file editor, which allows the user to then add that block to the pipeline.

See gif for example. The `damp_frost.py` data exporter block file already exists. User  creates new block with same name `damp frost`. Error appears but has link to open existing block file. Then user adds block to pipeline.
final create block flow existing name

wangxiaoyou1993 (approved):

PR #3203, opened 2023-08-07
[dy] Fix variables dir construction

dy46:
# Summary

Call `os.expanduser` on the variables dir when it comes from the project metadata or the default path.

# Tests

tested locally

cc: @wangxiaoyou1993


wangxiaoyou1993 (approved):
Could you add a unit test for the DEFAULT_MAGE_DATA_DIR case?

dy46:
i do actually have a unit test for the `DEFAULT_MAGE_DATA_DIR` case, but in the test environment it uses the `./` directory by default, so it didn't catch that issue

PR #3215, opened 2023-08-08
[dy] Add openssl dockerfile

dy46:
# Summary

Add dockerfile for openssl 1.1.1. I'm not sure how downgrading to openssl 1.1.1 will affect packages in the future, so I thought it could be good to create a separate dockerfile to install openssl 1.1.1. This way if a user doesn't need openssl 1.1.1, they can just use the standard docker images.

I also removed the temporary pymssql code because the previous issue was resolved in version 2.2.8

# Tests

tested locally

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/orchestration/db/__init__.py
+++ b/mage_ai/orchestration/db/__init__.py
@@ -39,8 +39,7 @@ elif not db_connection_url:
             db_connection_url = f'sqlite:///{TEST_DB}'
         elif os.path.exists('mage_ai/orchestration/db/'):
             # For local dev environment
-            path = os.path.abspath(os.path.join('mage_ai/orchestration/db', 'mage-ai.db'))
-            db_connection_url = f'sqlite:///{path}'
+            db_connection_url = 'sqlite:///mage_ai/orchestration/db/mage-ai.db'
         elif os.path.exists('mage-ai.db'):
             # For backward compatiblility
             db_connection_url = f'sqlite:///{get_variables_dir()}/mage-ai.db'
--- a/mage_ai/orchestration/db/__init__.py
+++ b/mage_ai/orchestration/db/__init__.py
@@ -39,7 +39,8 @@ elif not db_connection_url:
             db_connection_url = f'sqlite:///{TEST_DB}'
         elif os.path.exists('mage_ai/orchestration/db/'):
             # For local dev environment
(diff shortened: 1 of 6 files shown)


openssl1_1_1.Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+
+# Install NFS dependencies, and pymssql dependencies
+RUN apt-get -y install nfs-common freetds-dev freetds-bin
+
+### Install openssl 1.1.1 instead of the default openssl 3.0.9
+# (Install compiling library Make)
+RUN apt-get -y install make 
+
+# (single command that will download latest binaries, extract them, cd into the directory, compile configuration and then install the files)  
+RUN wget https://www.openssl.org/source/openssl-1.1.1g.tar.gz
```
wangxiaoyou1993, line 21: Can it be installed based on base Mage docker image?

dy46: hmm i could do that, but doesn't that mean we would have to wait until the base docker image is done building before creating this image?

wangxiaoyou1993: I would expect user to build the docker image themselves with this custom Dockerfile. It's similar with how we provide the Dockerfile for Mage with spark: https://github.com/mage-ai/mage-ai/tree/master/integrations/spark

openssl1_1_1.Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM mageai/mageai:latest
+LABEL description="Deploy Mage on ECS"
```
wangxiaoyou1993, line 2: No need to specify LABEL, USER here if they're same with the base docker image.

mage_ai/orchestration/db/__init__.py
```diff
@@ -39,7 +39,8 @@
             db_connection_url = f'sqlite:///{TEST_DB}'
         elif os.path.exists('mage_ai/orchestration/db/'):
             # For local dev environment
-            db_connection_url = 'sqlite:///mage_ai/orchestration/db/mage-ai.db'
+            path = os.path.abspath(os.path.join('mage_ai/orchestration/db', 'mage-ai.db'))
```
wangxiaoyou1993, line 42: What's this change for?

dy46: it's something to fix sqlite on dev, but i'll move it to a separate PR

PR #3224, opened 2023-08-09
[dy] Update where the `frontend_dist_base_path` is created

dy46:
# Description

I updated where the `frontend_dist_base_path` folder is created because it's possible that Mage won't have permission to edit the folder where the Mage package is installed. I chose to use the `get_variables_dir()` directory since I wasn't really sure where else to put it.

I also changed the `print` statements in the `server.py` file to `logger.info` since the print statements don't show up in server logs outside of the dev environment.

# How Has This Been Tested?

- [x] Tested locally with `MAGE_BASE_PATH` environment variable set
- [X] Existing unit tests pass

# Checklist
- [] I have made corresponding changes to the documentation

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/server/server.py
+++ b/mage_ai/server/server.py
@@ -76,7 +76,7 @@ from mage_ai.settings import (
     SHELL_COMMAND,
     USE_UNIQUE_TERMINAL,
 )
-from mage_ai.settings.repo import set_repo_path
+from mage_ai.settings.repo import DEFAULT_MAGE_DATA_DIR, get_repo_name, set_repo_path
 from mage_ai.shared.constants import InstanceType
 from mage_ai.shared.logger import LoggingLevel
 from mage_ai.shared.utils import is_port_in_use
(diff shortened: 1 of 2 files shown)


mage_ai/server/server.py
```diff
@@ -113,7 +117,7 @@ def post(self, action_type):
         self.write(dict(scheduler=dict(status=scheduler_manager.get_status())))
 
 
-def replace_base_path(base_path: str) -> None:
+def replace_base_path(base_path: str) -> str:
```
wangxiaoyou1993, line 120: Let's add return value to the docstring

mage_ai/server/server.py
```diff
@@ -122,7 +126,7 @@ def replace_base_path(base_path: str) -> None:
         base_path (str): The base path to replace the placeholder with.
     """
     src = os.path.join(os.path.dirname(__file__), BASE_PATH_TEMPLATE_EXPORTS_FOLDER)
-    dst = os.path.join(os.path.dirname(__file__), BASE_PATH_EXPORTS_FOLDER)
+    dst = os.path.join(get_variables_dir(), BASE_PATH_EXPORTS_FOLDER)
```
wangxiaoyou1993, line 129: This is a corner case but I want to make sure we catch this.
The variables_dir could be a s3 path.

wangxiaoyou1993 (commented):
Could you list the detailed test cases in the Test section?

dy46:
@wangxiaoyou1993 I added some logic to handle the s3 case.

Issue #3122, opened 2023-07-28
Setting time zone doesn't change completion time


mattppal:
https://mageai.slack.com/archives/C03HTTWFEKE/p1690550078083589

It appears that setting the timezone doesn't change the `completed` time.

PR #3229, opened 2023-08-09
[jk] Block run and pipeline run start/completion times in UTC

johnson-mage:
# Description
- If the timezone setting on a VM or Docker container running Mage was changed from UTC (which Mage uses across the app), the block and pipeline run completion times would be set in that timezone, although the start times were still in UTC. This caused confusion as a run could seem like it took a lot longer than it actually did. For example, if the timezone was set to UTC+3 (3 hours ahead of UTC) and a pipeline was run, its completion time appeared 3 hours after the start time even though the run did not take 3 hours.
- This PR makes the completion time for block and pipeline runs in UTC even if the VM or Docker container is set to a different timezone. This also updates the backfill start and completion time as those were also set to the system's timezone setting.
- Resolves https://github.com/mage-ai/mage-ai/issues/3122

# How Has This Been Tested?
- [x] Changed timezone of local Docker container to Egypt time (UTC+3). Ran a pipeline and confirmed its start time (`execution_date` since the run was created using "Run pipeline now" button) and creation time were in UTC, while the completion time was in Egypt time. After adding fix, re-ran same pipeline and confirmed its completion time was now in UTC while the system timezone setting was still Egypt time. In screenshot below, the top row was in UTC (after fix) and the bottom row was in UTC+3 time (before fix).
image

- [x] Performed same test as above but with backfills. Previously, the backfill start and end times were in the system timezone setting, or Egypt/UTC+3 on my Docker container, but after adding the fix, the backfill start and end times were in UTC. In screenshot below, the top row was in UTC (after fix) and the bottom row was in UTC+3 time (before fix).
image

# Checklist

### DIFF
--- a/mage_ai/api/resources/BackfillResource.py
+++ b/mage_ai/api/resources/BackfillResource.py
@@ -1,11 +1,13 @@
 from datetime import datetime
+
+import pytz
+from sqlalchemy import desc
+
 from mage_ai.api.resources.DatabaseResource import DatabaseResource
-from mage_ai.orchestration.backfills.service import start_backfill, cancel_backfill
+from mage_ai.orchestration.backfills.service import cancel_backfill, start_backfill
 from mage_ai.orchestration.db import safe_db_query
 from mage_ai.orchestration.db.models.schedules import Backfill
 from mage_ai.shared.hash import extract, merge_dict
-from sqlalchemy import desc
-import pytz
 
 ALLOWED_PAYLOAD_KEYS = [
     'block_uuid',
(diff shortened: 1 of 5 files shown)


johnson-mage:
@wangxiaoyou1993 Wasn't sure if adding unit tests was warranted for this PR

mage_ai/api/resources/BackfillResource.py
```diff
@@ -5,6 +5,7 @@
 from mage_ai.orchestration.db.models.schedules import Backfill
 from mage_ai.shared.hash import extract, merge_dict
 from sqlalchemy import desc
+import pytz
```
wangxiaoyou1993, line 8: Have you set up precommit hook? It'll fix the import statement order for you.

johnson-mage: I used to but removed it cuz i wasnt able to use it with lazygit, but I just added a vscode setting that lets me sort python imports quickly with a shortcut (shift + cmd + o). I didn't realize that feature was there but will use it moving forward.

wangxiaoyou1993 (approved):

PR #3231, opened 2023-08-09
[dy] Fix setting project uuid error

dy46:
# Description

Fix issue setting project uuid on project start.

# How Has This Been Tested?

- [x] Tested locally on start up
- [x] Tested with unit tests

# Checklist

cc: @wangxiaoyou1993


mage_ai/data_preparation/repo_manager.py
```diff
@@ -293,23 +293,28 @@ def get_variables_dir(
...
-except Exception:
-    pass
+def set_project_uuid_from_metadata():
+    global project_uuid
+    try:
+        with open(get_metadata_path(), 'r', encoding='utf-8') as f:
+            config = yml.load(f) or {}
+            project_uuid = config.get('project_uuid')
+    except Exception:
+        pass
```
wangxiaoyou1993, line 303: I wonder whether we should catch the exception. If the metadata file can't be loaded by yaml, the service shouldn't be started, right?

wangxiaoyou1993: Can you test the following cases?
* The metadata file doesn't exist
* The metadata file has in valid json
* It fails to open the metadata file randomly

mage_ai/data_preparation/repo_manager.py
```diff
@@ -293,23 +293,28 @@ def get_variables_dir(
...
 
 
-def update_project_uuid():
+def init_project_uuid():
```
wangxiaoyou1993, line 306: Can we add another check here? If `get_repo_config().project_uuid` exists, we don't need to set the new uuid, right?

dy46: yeah the idea is that the `project_uuid` global variable will contain the value from `get_repo_config().project_uuid`

mage_ai/data_preparation/repo_manager.py
dy46: * i added a check for this, but the metadata file will be created in the `init_project_uuid` call
* this will cause the server to not start because the `init_project_uuid` call will fail
* not sure why this would happen, but it could cause problems... should i add a retry or something?

mage_ai/data_preparation/repo_manager.py
wangxiaoyou1993: I don't know whether there could be race conditions if multiple servers are started since we allow user to start multiple servers.

wangxiaoyou1993: oh, just saw you added the check here. so it should be good.

dy46: I added another check

dy46: sorry, i was looking at it again, and it made sense to add the 2nd check

wangxiaoyou1993 (approved):

wangxiaoyou1993:
Also, please tag the PR with the proper labels.

PR #3234, opened 2023-08-10
[dy] Update error logging for block runs

dy46:
# Description

Looking into the issue for [this slack thread](https://mageai.slack.com/archives/C03HTTWFEKE/p1689936582602419?thread_ts=1689928310.536919&cid=C03HTTWFEKE). It seems like the logs from the block runs are not getting logged for this user which I think is a separate issue or maybe is a result of the user's setup.

For now, I changed the logging to print one error log in the block run and one error log in the pipeline run. I also removed the error logging from the `retry` method because it doesn't really add any information and can be confusing.

# How Has This Been Tested?

- [x] Tested with a local integration pipeline run
- [x] Tested with a local batch pipeline run

Screenshot 2023-08-16 at 2 42 32 PM

# Checklist

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/data_preparation/executors/block_executor.py
+++ b/mage_ai/data_preparation/executors/block_executor.py
@@ -253,12 +253,12 @@ class BlockExecutor:
 
                 result = __execute_with_retry()
             except Exception as error:
-                # self.logger.exception(
-                #     f'Failed to execute block {self.block.uuid}',
-                #     **merge_dict(tags, dict(
-                #         error=error,
-                #     )),
-                # )
+                self.logger.exception(
+                    f'Failed to execute block {self.block.uuid}',
+                    **merge_dict(tags, dict(
+                        error=error,
+                    )),
+                )
                 if on_failure is not None:
                     on_failure(
                         self.block_uuid,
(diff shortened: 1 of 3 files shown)


mage_ai/data_preparation/executors/block_executor.py
```diff
@@ -228,8 +228,6 @@ def execute(
                     delay=retry_config.delay,
                     max_delay=retry_config.max_delay,
                     exponential_backoff=retry_config.exponential_backoff,
-                    logger=self.logger,
```
wangxiaoyou1993, line 231: Did user mention the retry logging is confusing?

dy46: no, but it just has a strange message in the log: `Exception thrown when attempting to run <function BlockExecutor.execute.<locals>.test.<locals>.__execute_with_retry at 0xffff8cde8280>, attempt 1 of 1`.

dy46: I see... I think the main issue I'm trying to avoid is having multiple error logs for the same block since if I add the pipeline run error message, it will add another error log

dy46: @wangxiaoyou1993 I changed the retry log to be a warning and added it back

wangxiaoyou1993: ok

wangxiaoyou1993 (approved):

PR #3262, opened 2023-08-12
add-repo-owner-write-permissions

### LINKED ISSUE #3260: Set permissions when using shutil explicitly in copy_template_directory
If we want to enable non-root usage, we need to make sure that when copying files/dirs with `shutil` we are setting the correct permissions. shutil will copy the source permissions without changing them. Setting mage to rx will make the default_repo also rx. Non root users will not be able to write to the default_repo, even if they own the dir.

After copying data from the mage package using e.g. `shutil.copy2` or `shutil.copytree` set the permissions using e.g. `os.chmod(path, stat.S_IRWXU)` to Read, write, and execute by owner.

Not using shutil to copy files, but rather create/write the files from code with correct permissions to start with.

**Additional context**
Needed to enable non-root usage of Docker Image
cc: @wangxiaoyou1993

christopherscholz:
# Description
Resolves #3260

Adds a function to recursively chmod files.
- Adds `700` on top of current permissions to default_repo recursively
- If file permission is `555`, then after `chmod(file, 0x700)` the file permission is `755`
- This is based on bitwise or => `0x555 | 0x700 == 0x755`

# How Has This Been Tested?
- [x] Running the Dockerfile and manually checking the permissions using `ls -l`
- [x] Added unit test `tests/shared/test_io.py IOTests.test_chmod`

# Checklist
- [x] I have made corresponding changes to the documentation; not needed

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/shared/io.py
+++ b/mage_ai/shared/io.py
@@ -1,8 +1,24 @@
-from typing import Callable
-import aiofiles
 import os
 import shutil
 import traceback
+from typing import Callable
+
+import aiofiles
+
+
+def chmod(path, mode, append=True):
+    # recursively runs chmod against the path and either adds or sets the permission(mode)
+    try:
+        current_mode = os.stat(path).st_mode
+    except Exception as e:
+        raise e
+    if append:
+        mode = current_mode | mode
+    os.chmod(path, mode)
+    for dirpath, _dirnames, filenames in os.walk(path):
+        os.chmod(dirpath, mode)
+        for filename in filenames:
+            os.chmod(os.path.join(dirpath, filename), mode)
 
 
 def safe_write(filepath: str, content: str, write_func: Callable = None):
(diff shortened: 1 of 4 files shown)


mage_ai/shared/io.py
```diff
@@ -1,8 +1,24 @@
...
 import os
 import shutil
 import traceback
+from typing import Callable
+
+import aiofiles
+
+
+def chmod(path, mode, append=True):
+    # recursively runs chmod against the path and either adds or sets the permission(mode)
```
dy46, line 10: can you update the comment to the docstring format?

mage_ai/shared/io.py
```diff
@@ -1,8 +1,24 @@
...
+    try:
+        current_mode = os.stat(path).st_mode
+    except Exception as e:
+        raise e
+    if append:
+        mode = current_mode | mode
+    os.chmod(path, mode)
```
dy46, line 17: can this operation raise an exception?

christopherscholz: I have wrapped it in a full try, except block and only print stack trace as with many other functions.

dy46 (approved):

PR #3283, opened 2023-08-15
[dy] Fix bugs for workspaces

dy46:
# Description

* Pipeline runs from other workspaces were not being filtered properly in the `/pipeline-runs` page
* Create default roles differently for subprojects in case they don't already exist

# How Has This Been Tested?

- [x] Unit tests
- [x] Tested with local k8s workspaces

# Checklist

cc:


mage_ai/orchestration/db/models/base.py
```diff
@@ -76,14 +74,16 @@ def save(self, commit=True) -> None:
                 raise e
 
     def update(self, **kwargs) -> None:
+        commit = kwargs.pop('commit', True)
```
dy46, line 77: cc: @wangxiaoyou1993 is this a safe change to make? it seems like it should be similar to the `create` method

wangxiaoyou1993: Yes, i think this is a safe change.

mage_ai/orchestration/db/models/oauth.py
```diff
@@ -188,6 +188,18 @@ def create_default_roles(
                     ],
                     commit=False,
                 )
+            elif permissions:
+                role.update(
+                    permissions=role.permissions + [
```
wangxiaoyou1993, line 193: Can this role already have the permission?

wangxiaoyou1993: Can you add comments to this method?

wangxiaoyou1993: why? does this method always return permissions? https://github.com/mage-ai/mage-ai/blob/master/mage_ai/orchestration/db/models/oauth.py#L277-L299

dy46: it doesn't always return permissions. it will first check to see if permissions already exist for the given entity and entity id, and only create new permissions if they don't exist.

i updated the method to be a little more clear. i'll add some comments too

wangxiaoyou1993: OK, if we want to reduce the unnecessary queries, we can just get the permission from the permissions list instead of making individual filter queries.

wangxiaoyou1993 (approved):

PR #3316, opened 2023-08-18
[jk] Add pagination to pipeline run block runs page

johnson-mage:
# Description
- Add pagination to Block Runs page (for a specific pipeline run). Helps to organize block runs, especially for dynamic pipelines that create many block runs.

# How Has This Been Tested?
Before (no pagination):
before - no block run pagination

After (with pagination):
after - with block run pagination

# Checklist

### DIFF
--- a/mage_ai/api/policies/PipelineRunPolicy.py
+++ b/mage_ai/api/policies/PipelineRunPolicy.py
@@ -92,6 +92,7 @@ PipelineRunPolicy.allow_query([
     'backfill_id',
     'disable_retries_grouping',
     'end_timestamp',
+    'exclude_block_runs',
     'global_data_product_uuid',
     'include_pipeline_type',
     'order_by[]',
--- a/mage_ai/api/policies/PipelineRunPolicy.py
+++ b/mage_ai/api/policies/PipelineRunPolicy.py
@@ -92,7 +92,6 @@ PipelineRunPolicy.allow_query([
     'backfill_id',
     'disable_retries_grouping',
     'end_timestamp',
-    'exclude_block_runs',
     'global_data_product_uuid',
     'include_pipeline_type',
     'order_by[]',
--- a/mage_ai/api/presenters/PipelineRunPresenter.py
+++ b/mage_ai/api/presenters/PipelineRunPresenter.py
@@ -21,9 +21,8 @@ class PipelineRunPresenter(BasePresenter):
     ]
 
     async def present(self, **kwargs):
+        query = kwargs.get('query', {})
         if constants.LIST == kwargs['format']:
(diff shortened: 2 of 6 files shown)


mage_ai/frontend/pages/pipelines/[pipeline]/runs/[run]/index.tsx
```diff
@@ -65,7 +72,9 @@ function PipelineBlockRuns({
 
   const { data: dataPipelineRun } = api.pipeline_runs.detail(
     pipelineRunProp.id,
-    {},
+    {
```
Comment (johnson-mage) [inline mage_ai/frontend/pages/pipelines/[pipeline]/runs/[run]/index.tsx:75]: We still need to periodically fetch the pipeline run so that we can display its running status, as well as its `execution_date` in the breadcrumbs:
image

mage_ai/frontend/pages/pipelines/[pipeline]/runs/[run]/index.tsx
```diff
@@ -65,7 +72,9 @@ function PipelineBlockRuns({
...
+      exclude_block_runs: true,
```
Comment (johnson-mage) [inline mage_ai/frontend/pages/pipelines/[pipeline]/runs/[run]/index.tsx:76]: Added this query param since we need to fetch the pipeline run, but we no longer need to include the block_runs in that payload. We fetch the block runs with a separate block runs request which already handles the `_limit` and `_offset` params.

mage_ai/api/presenters/PipelineRunPresenter.py
```diff
@@ -45,18 +44,24 @@ async def present(self, **kwargs):
...
 
-            arr = []
-            for r in block_runs:
-                block_run = r.to_dict()
-                block_run['pipeline_schedule_id'] = pipeline_schedule.id
-                block_run['pipeline_schedule_name'] = pipeline_schedule.name
-                arr.append(block_run)
-            arr.sort(key=lambda b: b.get('created_at'))
-            data['block_runs'] = arr
+            exclude_block_runs = query.get('exclude_block_runs', [False])
```
dy46, line 49: maybe we can do it like this to keep things consistent between different APIs: https://github.com/mage-ai/mage-ai/blob/master/mage_ai/api/presenters/GitCustomBranchPresenter.py#L33

johnson-mage: ok

mage_ai/api/policies/PipelineRunPolicy.py
```diff
@@ -92,6 +92,7 @@ def entity(self):
     'backfill_id',
     'disable_retries_grouping',
     'end_timestamp',
+    'exclude_block_runs',
```
dy46, line 95: let's remove this

dy46 (approved):

PR #3317, opened 2023-08-18
[jk] Add pipeline "created_at" property

johnson-mage:
# Description
- Add `created_at` property to pipeline model and display in the Pipelines dashboard table, which users can use to sort the pipelines table by when sorting is implemented.

# How Has This Been Tested?
- Updated unit tests
- New pipelines will now have the `created_at` property (old pipelines are not applicable since they were already created):
image

# Checklist
- [X] I have added unit tests that prove my fix is effective or that my feature works

### DIFF
--- a/mage_ai/data_preparation/models/pipeline.py
+++ b/mage_ai/data_preparation/models/pipeline.py
@@ -6,6 +6,7 @@ import shutil
 from typing import Callable, Dict, List
 
 import aiofiles
+import pytz
 import yaml
 from jinja2 import Template
 
@@ -55,7 +56,7 @@ class Pipeline:
         self.block_configs = []
         self.blocks_by_uuid = {}
         self.concurrency_config = dict()
-        self.created_at = datetime.datetime.utcnow()
+        self.created_at = None
         self.data_integration = None
         self.description = None
         self.executor_config = dict()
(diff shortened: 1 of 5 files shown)


mage_ai/data_preparation/models/pipeline.py
```diff
@@ -154,6 +155,7 @@ def create(self, name, pipeline_type=PipelineType.PYTHON, repo_path=None):
         # Update metadata.yaml with pipeline config
         with open(os.path.join(pipeline_path, PIPELINE_CONFIG_FILE), 'w') as fp:
             yaml.dump(dict(
+                created_at=str(datetime.datetime.utcnow()),
```
dy46, line 158: I would use datetime.datetime.now(tz=pytz.UTC)

mage_ai/data_preparation/models/pipeline.py
```diff
@@ -55,6 +55,7 @@ def __init__(self, uuid, repo_path=None, config=None, repo_config=None, catalog=
         self.block_configs = []
         self.blocks_by_uuid = {}
         self.concurrency_config = dict()
+        self.created_at = datetime.datetime.utcnow()
```
dy46, line 58: does it make sense to default this to current time?

mage_ai/data_preparation/models/pipeline.py
johnson-mage: ok

dy46 (approved):
backend changes LGTM

mage_ai/data_preparation/models/pipeline.py
johnson-mage: Initially, i thought it'd just be a placeholder since we read it from the metadata config file, but thanks for pointing it out since it might have unintended side effects whenever the Pipeline class is instantiated.

Issue #3330, opened 2023-08-22
Docker: Python Multi Stage Build


christopherscholz:
**Is your feature request related to a problem? Please describe.**
Currently if any Python dependency or Mage code changes all downstream layers have to be rebuild. The same goes for if any upstream layers change, then all Python downstream layers have to be rebuild. The python layers take quite a while to build.

**Describe the solution you'd like**
Decouple the Python layers from the remaining build by using multi stage build.
- Setup an virtual environment in a separate image, which is the `python-builder`
- Install packages in the virtual environment
- `COPY` the virtual environment from `python-builder` to the main image (runner) using `--link`
- Fix `PATH` to use virtualenv python bin as primary executable

**Describe alternatives you've considered**
Don't use mult stage build

**Additional context**
cc: @dy46, @wangxiaoyou1993

PR #3331, opened 2023-08-22
Docker: Python Multi Stage Build

christopherscholz:
# Description
Resolves: #3330

Set up a builder Image, which builds python packages inside a virtual environment.
`COPY --link` virtual environment to runner Image and set `PATH` to virtual environment bin.

If builder image is untouched by changes, then it does not need to be rebuild. If builder image is changed, then only downstream layers have to be rebuild. e.g. if a new version of an apt package is released, then the python layers do not have to be rebuild.

Also adds `/bin/bash -c` as prefix to start mage-ai/server in dev, which is needed to make `$PATH` available and therefore starting the correct python bin (virtual env). Otherwise it would start the /usr/bin/python binary.

# How Has This Been Tested?
- [x] Ran Dockerfile
- [x] Ran dev.Dockerfile with scripts/dev.sh

# Checklist
- [x] I have added unit tests that prove my fix is effective or that my feature works; not needed
- [x] I have made corresponding changes to the documentation; not needed

# Additional Info
Builds on top of #3329, which needs to be migrated first.
cc: @dy46, @wangxiaoyou1993

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,27 @@
+ARG PIP=pip3
+ARG VIRTUAL_ENV=/opt/venv
+
+FROM python:3.10 as builder
+ARG PIP
+ARG VIRTUAL_ENV
+RUN python3 -m venv ${VIRTUAL_ENV}
+ENV PATH=${VIRTUAL_ENV}/bin:$PATH
+
+RUN ${PIP} install --no-cache-dir sparkmagic
+# Mage Integration
+RUN ${PIP} install --no-cache-dir "git+https://github.com/mage-ai/singer-python.git#egg=singer-python"
+RUN ${PIP} install --no-cache-dir "git+https://github.com/mage-ai/google-ads-python.git#egg=google-ads"
+RUN ${PIP} install --no-cache-dir "git+https://github.com/mage-ai/dbt-mysql.git#egg=dbt-mysql"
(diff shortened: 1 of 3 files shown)


dy46:
@christopherscholz the `--link` flag didn't work for me until i upgraded my docker. is there way to make this change backwards compatible?

christopherscholz:
> @christopherscholz the `--link` flag didn't work for me until i upgraded my docker. is there way to make this change backwards compatible?

hmm its reasonable new (Dockerfile Syntax 1.4+ -> March 2022), but gives a huge benefit for this huge build.
I don't think there is a backwards compatible way to do this. 

Of course I could remove it, but then the main benefit would be gone. Still worth it though.

dy46:
also, some of the python packages depend on packages installed in the `apt-get` command. i believe these packages are required to install the python packages. the `pip install mage_integrations/` step is failing when i do the docker build locally
```
# odbc dependencies
msodbcsql18
unixodbc-dev
# pymssql dependencies
freetds-dev
freetds-bin
```

christopherscholz:
> also, some of the python packages depend on packages installed in the `apt-get` command. i believe these packages are required to install the python packages. the `pip install mage_integrations/` step is failing when i do the docker build locally
> 
> ```
> # odbc dependencies
> msodbcsql18
> unixodbc-dev
> # pymssql dependencies
> freetds-dev
> freetds-bin
> ```

hmm then I don't think the multiple stages idea makes sense. I thought these were run but not build dependencies. After checking the setup.py of these packages I'm now a bit wiser....

PR #3335, opened 2023-08-22
Elasticsearch as Sink for Streaming

sujiplr:
# Description
Add ElasticSearch sink to streaming pipeline. This sink will help in append only and update options for the documents.

# How Has This Been Tested?

Tested this using Kafka as the source and elasticsearch as the destination for single and bulk messages.
cc:


mage_ai/data_preparation/templates/data_exporters/streaming/elasticsearch.yaml
```diff
@@ -0,0 +1,8 @@
+connector_type: elasticsearch
+host: https://localhost
```
wangxiaoyou1993, line 2: localhost usually uses http

wangxiaoyou1993 (approved):

Issue #3202, opened 2023-08-07
Mage dbt runs don't respect overriden schemas


mattppal:
When overriding the default naming convention macro in dbt, Mage does not pick up this behavior and the resulting `SELECT * LIMIT 1000` uses an incorrect naming scheme. This causes blocks using custom names to fail.

https://mageai.slack.com/archives/C03HTTWFEKE/p1691407959751919

PR #3368, opened 2023-08-29
[dy] Get schema from generate_schema_name

dy46:
# Description

This PR is to resolve this issue: https://github.com/mage-ai/mage-ai/issues/3202

~~From what I've gathered from dbt docs, you can overwrite the default schema naming convention by overwriting the `generate_schema_name` macro. The only way that I could figure out how to get the schema from that macro is by doing the following:~~

* ~~create a new macro that calls the `generate_schema_name` macro and prints out the result~~
* ~~execute the new macro with `dbt run-operation`~~
* ~~parse the output of the command to get the schema.~~

Updated (8/30)

Instead of running the macro in Mage which can potentially take a long time, we will try to get the schema from the `manifest.json` file that is updated after every dbt operation. Each model will have its own entry in the `nodes` field. 

# How Has This Been Tested?

- [x] Tested locally
- [ ] Unit tests (will add after the approach in the PR is verified)

# Checklist

cc:

### DIFF
--- a/mage_ai/data_preparation/models/block/dbt/__init__.py
+++ b/mage_ai/data_preparation/models/block/dbt/__init__.py
@@ -371,7 +371,6 @@ class DBTBlock(Block):
                             self,
                             dbt_profile_target,
                             project_full_path,
-                            profiles_dir,
                             limit=limit,
                             variables=variables,
                         )
--- a/mage_ai/data_preparation/models/block/dbt/__init__.py
+++ b/mage_ai/data_preparation/models/block/dbt/__init__.py
@@ -370,6 +370,8 @@ class DBTBlock(Block):
                         df = fetch_model_data(
                             self,
                             dbt_profile_target,
+                            project_full_path,
+                            profiles_dir,
                             limit=limit,
                             variables=variables,
                         )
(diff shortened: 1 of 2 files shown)


dy46:
if any DBT experts can take a look at this PR and let me know if the approach looks ok or if you know of any other methods of resolving this issue, that would be appreciated! @mattppal @wangxiaoyou1993 @tommydangerous

mattppal:
I think this works, but how performant is it? I know calling dbt macros can be slow. Is the model already compiled when this is executed?

Honestly, I need to understand more about how our dbt works— I have some ideas, but it depends on how we handle things.

dy46:
yes, it is compiled. it should be relatively quick unless the macro logic is really complex for some reason. running `dbt run-operation` just takes a couple of seconds

mattppal:
> yes, it is compiled. it should be relatively quick unless the macro logic is really complex for some reason. running `dbt run-operation` just takes a couple of seconds

Hm I guess I'm worried that for a project with 1000 models (not uncommon) this would be more expensive.

Could you grab it from the `manifest.json`? https://schemas.getdbt.com/dbt/manifest/v8/index.html#nodes_additionalProperties_oneOf_i0_schema

`nodes` -> `schema`?

dy46:
hmm that looks like it could work. i'll take a look

christopherscholz:
@dy46 I guess this is not needed anymore, as its solved by #3497.
As we are not using the mage connector from io to load the preview, but instead use `dbt show`, dbt automatically gets the correct schema name.

dy46:
@christopherscholz great, I'll close this then

PR #3373, opened 2023-08-29
[xh]Return dict for GENERATE_COMMENT_FOR_CODE usecase

matrixstone:
# Description
(1) I found for `GENERATE_COMMENT_FOR_CODE` use case, it is now returning a string directly.
This PR updates it to return a dict with key `block_code`
(2) Also a minor fix when generating block documentation, it should `await` instead of `asyncio.run`

# How Has This Been Tested?
Tested locally 
Screenshot 2023-08-28 at 10 46 56 PM

# Checklist

cc:
@wangxiaoyou1993 @dy46

### DIFF
--- a/mage_ai/ai/llm_pipeline_wizard.py
+++ b/mage_ai/ai/llm_pipeline_wizard.py
@@ -491,7 +491,9 @@ class LLMPipelineWizard:
         chain = LLMChain(llm=self.llm, prompt=prompt_template)
         function_comments_json = await chain.arun(block_content=block_content)
         function_comments = json.loads(function_comments_json)
-        return self.__insert_comments_in_functions(block_content, function_comments)
+        return dict(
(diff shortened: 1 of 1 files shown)


mage_ai/ai/llm_pipeline_wizard.py
```diff
@@ -491,7 +491,9 @@ async def async_generate_comment_for_block(self, block_content: str) -> str:
         chain = LLMChain(llm=self.llm, prompt=prompt_template)
         function_comments_json = await chain.arun(block_content=block_content)
         function_comments = json.loads(function_comments_json)
-        return self.__insert_comments_in_functions(block_content, function_comments)
+        return dict(
```
wangxiaoyou1993, line 494: you don't need to update it to return dict now.
The UI already integrates this API response format.

PR #3378, opened 2023-08-29
Influxdb data exporter

mazieb:
# Description
Add data exporter for InfluxDB. 

# How Has This Been Tested?
- [x] Tested Locally

# Checklist

cc:
@wangxiaoyou1993


mage_ai/streaming/sinks/influxdb.py
```diff
@@ -0,0 +1,95 @@
...
+            max_retry_delay=125_000,
+            exponential_base=2,
+        )
+        self.write_api = self.client.write_api(write_options=options)
+        self._print("Finish initializing writer.")
+
+    def write(self, data: Dict):
+        if isinstance(data, dict):
+            record = {
+                "measurement": data.get("measurement", self.config.measurement),
```
wangxiaoyou1993, line 63: nit: we prefer using single quote in the code

mage_ai/streaming/sinks/influxdb.py
```diff
@@ -0,0 +1,95 @@
...
+                "time": data.get("timestamp", get_timestamp_ms()),
+                "tags": data.get("tags", {}),
+                "fields": data.get("fields", data),
+            }
+        elif isinstance(data, Iterable):
+            for i, value in enumerate(data):
```
wangxiaoyou1993, line 69: Do you only get the last record?

mage_ai/data_preparation/templates/data_exporters/streaming/influxdb.yaml
```diff
@@ -0,0 +1,5 @@
+connector_type: influxdb
+url: "http://localhost:8086"
+token: "my-token"
+org: "my-org"
+bucket: "data"
```
wangxiaoyou1993, line 5: nit: add empty line at end of file

wangxiaoyou1993 (commented):
Thanks for putting up the PR! It looks good overall.
Could you also add the new package to setup.py?
https://github.com/mage-ai/mage-ai/blob/master/setup.py#L109
https://github.com/mage-ai/mage-ai/blob/master/setup.py#L101

mage_ai/streaming/sinks/influxdb.py
```diff
@@ -0,0 +1,95 @@
...
+            jitter_interval=0,
+            retry_interval=5_000,
+            max_retries=5,
+            max_retry_delay=125_000,
+            exponential_base=2,
+        )
+        self.write_api = self.client.write_api(write_options=options)
+        self._print("Finish initializing writer.")
+
+    def write(self, data: Dict):
```
wangxiaoyou1993, line 60: To include the metadata in the data, I would update the data format to be
```
{
    "record": {...},
    "metadata": {...}
}
``

File: mage_ai/streaming/sinks/influxdb.py (hunk 2 above)
Comment (mazieb) [reply mage_ai/streaming/sinks/influxdb.py:69]: I am assuming you refer to line 66: It's a bit tricky, not only the last record is taken but the whole record.

- If data["fields"] is found we assume the data can be found there.
- If data["fields"] does not exist, the record is malformed. But there is a chance that the data record is the data itself (but without "fields" and most probably without "tags", "time", "measurement") because the schema was ignored. Then we just take the whole data record as fields. It's not good practice to write everything as fields in InfluxDB, but it works.

Alternatively we could reject dict records, if the schema does not match.

File: mage_ai/streaming/sinks/influxdb.py (hunk 4 above)
Comment (mazieb) [reply mage_ai/streaming/sinks/influxdb.py:60]: I am not used to github, so please excuse my confusion. 
If you are referring to line 57: I think the best place for write_options is the config yaml.
If you are referring to line 60:

Yes that's a way to structure the data. I'am not sure why you call it record and metadata, as record suggests it includes metadata itself. I would suggest data and metadata instead, i.e.

```python
message = {"data": {...}, "metadata": {...}}
```

But that's personal preference. A typical message to influxdb would then look like this:

```python
message = {
  "data": {"bees": 23}
  "metadata": {
    "measurement": "census",
    "time": 1693472675783,
    "tags": {"location": "klamath", "scientist": "anderson"}
  }
}
```

So basically influxdb "fields" can be found in the "data" section, everything else is in "metadata" and has defaults if not given explicitly. I am currently preparing a PR for including metadata in kafka sink and source. A typical message from kafka source could look like this:

```python
message = {
  "data": {
    "bees": {
      "location": "klamath", 
      "scientist": "anderson", 
      "count": 23
    },
  "metadata": {
    "topic": "census",
    "partition": 0,
    "offset": 6,
    "time": 1693472675783
  }
```

Ultimately all sinks and sources should share some basic interpretation of the messages, e.g. time stamps from kafka should be used when writing them to influxdb without the need to restructure the data. Specialties like the distinction between tags and fields in InfluxDB can't or shouldn't be handled automatically (IMHO).

I will update my code to match the new schema.

wangxiaoyou1993: Yes, that makes sense. We can use this format.

mage_ai/streaming/sinks/influxdb.py
```diff
@@ -0,0 +1,109 @@
...
+            retry_interval=5_000,
+            max_retries=5,
+            max_retry_delay=125_000,
+            exponential_base=2,
+        )
+        self.write_api = self.client.write_api(write_options=options)
+        self._print('Finish initializing writer.')
+
+    def write(self, message: Dict):
+        data = message['data']
```
wangxiaoyou1993, line 61: nit: if the message doesn't have the 'data' key, let's set data to message itself as fallback

wangxiaoyou1993 (approved):

wangxiaoyou1993:
Do you want to update the PR status from "Draft" to "Ready for review"?

wangxiaoyou1993:
Did you test this InfluxDB data exporter locally?

mazieb:
> Did you test this InfluxDB data exporter locally?

Yes it's tested locally and working.

Anaisdg:
Wow this is so cool! Thank you! 
One note, I noticed you're using the InfluxDB v2 Client Library. 
While the InfluxDB v1, v2, and v3 Client use the same /write api endpoint, the [InfluxDB v3 Client](https://github.com/InfluxCommunity/influxdb3-python) library has methods for writing more types of data including DataFrames, parquet, feather, json etc:
https://github.com/InfluxCommunity/influxdb3-python/tree/main/Examples/file-import

I would also recommend using the v3 client library for the source as well.

PR #3384, opened 2023-08-29
enhancement: Token Period from env

juancaven1988:
# Description
Add a new env variable to define the token expiration date

# How Has This Been Tested?
Test funtions added in local

cc:

@dy46 @tommydangerous


mage_ai/settings/__init__.py
```diff
@@ -27,10 +29,15 @@ def is_disable_pipeline_edit_access():
...
 
+#------------------------- DISABLE TERMINAL ----------------------#'
+
+DISABLE_TERMINAL = os.getenv('DISABLE_TERMINAL', '0').lower() in ('true', '1', 't')
+
 # ----------------- Authentication settings ----------------
 REQUIRE_USER_AUTHENTICATION = \
     os.getenv('REQUIRE_USER_AUTHENTICATION', 'False').lower() in ('true', '1', 't')
 AUTHENTICATION_MODE = os.getenv('AUTHENTICATION_MODE', 'LOCAL')
+TOKEN_PERIOD = os.getenv('TOKEN_PERIOD', '30d').lower()
```
dy46, line 40: i think it would be better to make this a number value which represents the amount of time in seconds

dy46: let's name it something a little more descriptive like: `MAGE_ACCESS_TOKEN_EXPIRY_TIME`

juancaven1988: Add changes according to suggestions

mage_ai/settings/__init__.py
```diff
@@ -27,10 +29,18 @@ def is_disable_pipeline_edit_access():
 
 DISABLE_TERMINAL = os.getenv('DISABLE_TERMINAL', '0').lower() in ('true', '1', 't')
 
+#------------------------- DISABLE TERMINAL ----------------------#'
+
+DISABLE_TERMINAL = os.getenv('DISABLE_TERMINAL', '0').lower() in ('true', '1', 't')
```
dy46, line 34: it seems like there are some extra changes here that aren't related to the current issue. can you remove them?

juancaven1988:
Add a new clean pull request

PR #3388, opened 2023-08-30
Fixes on MongoDB destination (Data integration)

Luishfs:
Fixed mongodb destination error after INCREMENTAL pipeline run
Added RECORD pre-processing
Added table_name option on destination config

# Description

## Issues

1.  **User reported not being able to modify data types on MongoDB destination**

Solution:
MongoDB sink was modified to support a preprocessing function on each record
This function checks each data type with the given schema, if a data type does not correspond to the given schema, the function will try to do a simple python type conversion. If failed, pipeline will break and user will be notified.

2. **User reported a ERROR message after doing a INCREMENTAL run**

Solution:
The error was being caused by the `state_path` not being defined on MongoDB `__init__.py ` file
State path was added

3. **User reported not being able to overwrite destination table name**

Solution:
Modified MongoDB sink file and added the table_name option on config.json
If no table_name is added. the default name will be used

# How Has This Been Tested?

Tested on PostgreSQL -> MongoDB pipeline

### Data type modification

Base example schema on Postgres:
Screenshot from 2023-08-30 01-41-09

Modifications were:

1. bool_value to string ( instead of bool)
2. char_value to number ( instead of char)
3. int_value to string ( instead of int)

Destination data schema:
Screenshot from 2023-08-30 01-40-57

Since the preprocessing record function tries to modify data types, it is not possible to configure more than one Not Null type on  a given record, if so, the code will raise a error:

Screenshot from 2023-08-30 01-40-16

# Checklist

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_integrations/mage_integrations/destinations/mongodb/target_mongodb/target.py
+++ b/mage_integrations/mage_integrations/destinations/mongodb/target_mongodb/target.py
@@ -126,7 +126,7 @@ class TargetMongoDb(Target):
         """Process a RECORD message.
 
         Args:
-            message_dict: TODO
+            message_dict (dict): JSONL Singer RECORD message
         """
         self._assert_line_requires(message_dict, requires={"stream", "record"})
 
--- a/mage_integrations/mage_integrations/destinations/mongodb/target_mongodb/target.py
+++ b/mage_integrations/mage_integrations/destinations/mongodb/target_mongodb/target.py
@@ -1,5 +1,5 @@
 """MongoDB target class."""
-
+import copy
 import json
 import logging
 import sys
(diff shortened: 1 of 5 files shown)


mage_integrations/mage_integrations/destinations/mongodb/target_mongodb/target.py
```diff
@@ -120,3 +121,46 @@ def _process_lines_internal(self, file_input: t.IO[str]) -> t.Counter[str]:
             stats[record_type] += 1
 
         return Counter(**stats)
+
+    def _process_record_message_mongo(self, message_dict: dict) -> None:
+        """Process a RECORD message.
+
+        Args:
+            message_dict: TODO
```
wangxiaoyou1993, line 129: nit: what's TODO?

wangxiaoyou1993 (approved):

mage_integrations/mage_integrations/destinations/mongodb/target_mongodb/target.py
Luishfs: This was written on Meltano SDK, just updated on https://github.com/mage-ai/mage-ai/pull/3388/commits/4b1ffcc5db054a47c0e0e9a6df5fcdaf51099e50

PR #3396, opened 2023-08-31
Enhancement: token expiry time env 

juancaven1988:
Description
Add a new env variable to define the token expiration date

How Has This Been Tested?
Test funtions added in local

cc:

@dy46 @tommydangerous


mage_ai/settings/__init__.py
```diff
@@ -31,6 +31,10 @@ def is_disable_pipeline_edit_access():
 REQUIRE_USER_AUTHENTICATION = \
     os.getenv('REQUIRE_USER_AUTHENTICATION', 'False').lower() in ('true', '1', 't')
 AUTHENTICATION_MODE = os.getenv('AUTHENTICATION_MODE', 'LOCAL')
+try:
+    MAGE_ACCESS_TOKEN_EXPIRY_TIME = int(os.getenv('MAGE_ACCESS_TOKEN_EXPIRY_TIME', 3600))  
+except ValueError:
+    MAGE_ACCESS_TOKEN_EXPIRY_TIME = 3600
```
dy46, line 37: can we change the default to 30 days in seconds, so that it matches the previous logic?

PR #3399, opened 2023-08-31
[dy] Add timeout for block runs and pipeline runs

dy46:
# Description

Add support for block run and pipeline run timeouts. The block run timeout is configured in the block settings, and the pipeline run timeout is configured in the trigger edit page. The timeouts are checked in the `pipeline_scheduler.schedule` method. One issue is that the actual timeout of the block/pipeline may not be exact (usually within 5 seconds of the set timeout), but I think that is ok for now.

Screenshot 2023-08-31 at 12 09 36 PM
Screenshot 2023-08-31 at 12 09 22 PM
Screenshot 2023-08-31 at 12 09 03 PM

# How Has This Been Tested?

- [x] Tested locally
- [x] Unit tests

# Checklist

cc: @wangxiaoyou1993


mage_ai/data_preparation/models/block/__init__.py
```diff
@@ -258,6 +259,7 @@ def __init__(
         self.color = block_color
         self.configuration = configuration
         self.has_callback = has_callback
+        self.timeout = timeout
```
wangxiaoyou1993, line 262: Do you need to update the BlockPolicy?

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -384,6 +389,51 @@ def build_tags(self, **kwargs):
             base_tags['hostname'] = HOSTNAME
         return merge_dict(kwargs, base_tags)
 
+    @safe_db_query
+    def __check_run_timeout(self) -> None:
```
wangxiaoyou1993, line 393: could you add docstring to the method?

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -508,6 +560,7 @@ def __schedule_blocks(self, block_runs: List[BlockRun] = None) -> None:
                 b.id,
                 self.pipeline_run.get_variables(),
                 self.build_tags(**tags),
+                timeout=block.timeout,
```
wangxiaoyou1993, line 563: Do you use it in the job?

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -384,6 +389,51 @@ def build_tags(self, **kwargs):
...
+                self.logger.error(
+                    f'Pipeline run timed out after {int(time_difference)} seconds',
+                    **self.build_tags(),
+                )
+                stop_pipeline_run(
+                    self.pipeline_run,
+                    self.pipeline,
+                    status=PipelineRun.PipelineRunStatus.FAILED
+                )
+                return
```
wangxiaoyou1993, line 409: do you still proceed with scheduling logic in `schedule` method after `stop_pipeline_run`?

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -183,6 +183,11 @@ def schedule(self, block_runs: List[BlockRun] = None) -> None:
         for b in self.pipeline_run.block_runs:
             b.refresh()
 
+        try:
+            self.__check_run_timeout()
```
wangxiaoyou1993, line 187: this method might not work with integration pipeline, streaming pipeline, and `run_pipeline_in_one_process` scenarios.

mage_ai/data_preparation/models/block/__init__.py
dy46: good call, i forgetting to turn on user auth since it always gets reset

mage_ai/orchestration/pipeline_scheduler.py
dy46: does it make sense for a streaming pipeline to have timeouts? i think i can make it work for integration pipelines, but it will be tricky for the `run_pipeline_in_one_process` method for block run timeouts.

wangxiaoyou1993: no need to support it for the cases. just need to make sure we only run the check for the scenarios that we support.

dy46: ok

dy46:
@wangxiaoyou1993 I updated the PR to check pipeline run timeout for non-streaming pipelines, and block run timeout for only batch pipelines.

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -253,6 +254,7 @@ def schedule(self, block_runs: List[BlockRun] = None) -> None:
             elif self.pipeline.run_pipeline_in_one_process:
                 self.__schedule_pipeline()
             else:
+                self.__check_block_run_timeout()
```
wangxiaoyou1993, line 257: If certain block runs time out and get updated to failed, we should not proceed with `__schedule_blocks`.
You can just return here and let it check `(self.pipeline_run.any_blocks_failed() and not self.allow_blocks_to_fail)` next time the `schedule` method is called.

dy46:
@wangxiaoyou1993 updated

wangxiaoyou1993 (approved):

PR #3405, opened 2023-09-01
Docker: Installation Method of Node using setup_17.x is no longer supported

christopherscholz:
# Description
Nodesource does not support setup scripts anymore. The installation is straight forward.
In addition I have
- updated to Nodejs 20, as its the current version and Nodejs 17 is EOL.
- removed npm as its already included in nodejs package

# How Has This Been Tested?
Ran dev.Dockerfile using scripts/dev.sh and checked whether localhost:3000 still works.

# Checklist
- [x] I have added unit tests that prove my fix is effective or that my feature works - no need

cc: @dy46

### DIFF
--- a/dev.Dockerfile
+++ b/dev.Dockerfile
@@ -7,12 +7,13 @@ USER root
 RUN \
   curl https://packages.microsoft.com/keys/microsoft.asc | apt-key add - && \
   curl https://packages.microsoft.com/config/debian/11/prod.list > /etc/apt/sources.list.d/mssql-release.list && \
-  curl -fsSL https://deb.nodesource.com/setup_17.x | bash - && \
+  curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg && \
+  NODE_MAJOR=20 && \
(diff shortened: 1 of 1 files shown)


johnson-mage:
@christopherscholz Thanks for catching this. Can we use Node v18 instead of v20 since v18 is the latest LTS version? Would the Docker setup still work?

christopherscholz:
> @christopherscholz Thanks for catching this. Can we use Node v18 instead of v20 since v18 is the latest LTS version? Would the Docker setup still work?

@johnson-mage 
Sure? 18 maintenance end is 2025-04-30 but 20 maintenance end is 2026-04-30. Also 20 becomes active lts in October as soon as 21 is release. 18 becomes maintenance lts then.

I can still change it to 18 though, but would not recommend.

johnson-mage:
> > @christopherscholz Thanks for catching this. Can we use Node v18 instead of v20 since v18 is the latest LTS version? Would the Docker setup still work?
> 
> @johnson-mage Sure? 18 maintenance end is 2025-04-30 but 20 maintenance end is 2026-04-30. Also 20 becomes active lts in October as soon as 21 is release. 18 becomes maintenance lts then.
> 
> I can still change it to 18 though, but would not recommend.

Got it. We can stick with 20. Let me test it out first.

johnson-mage (approved):

PR #3414, opened 2023-09-04
Switching from setup.py to pure pyproject.toml

christopherscholz:
# Description
This switches both mage_ai and mage_integration from setup.py to pyproject.toml.

As build backend I keep setuptools and not switch to poetry fully.
Poetry unfortunately does not support [PEP-621](https://peps.python.org/pep-0621/) yet. This means that `poetry install` does not take the dependencies into account yet. Therefore having to install mage_ai and mage_integrations via pip install and not poetry install inside the virtual env.

Also updated
- dev.Dockerfile
- build-and-test github action workflow
  - dropped 3.7 support as its EOL
  - fixed pip install of mage-ai and mage-integration
- publish-to-pypi github action workflow
  - fixed python build command

Also checked all imports in mage_ai and mage_integrations to check missing dependencies.

# How Has This Been Tested?
- [x] See Workflow
- [x] Ran scripts/dev.sh
- Pypi publish not tested

# Checklist

cc: @wangxiaoyou1993, @dy46, @jcorrado76


pyproject.toml
```diff
@@ -1,32 +1,274 @@
...
-interrogate = "*"
-isort = "*"
-commitizen = "*"
-ipykernel = "^6.17.1"
-pandas-stubs = "*"
-flake8-bugbear = "*"
-sourcery = "*"
+dependencies = [
+  'aiofiles==22.1.0',
+  'aiohttp',
```
dy46, line 27: how did you come up with this list of dependnencies? `aiohttp` isn't in the original `requirements.txt` or `setup.py`

pyproject.toml
```diff
@@ -1,32 +1,274 @@
...
+  'alembic>=1.7.5',
+  'async_timeout',
+  'bcrypt==4.0.1',
+  'click',
+  'confluent-avro',
```
dy46, line 32: `confluent-avro` was originally part of the "streaming" optional dependencies. is there a reason that we combined that with the required dependencies?

pyproject.toml
```diff
@@ -1,32 +1,274 @@
...
+  'croniter==1.3.7',
+  'cryptography==36.0.2',
+  'dask>=2022.2.0',
+  'Faker==4.14.0',
+  'flask-cors',
+  'flask',
```
dy46, line 38: i don't think we need flask dependencies

pyproject.toml
christopherscholz: I went through all imports in code. E.g. its used here: https://github.com/mage-ai/mage-ai/blob/f06f6117b5d2d7f676899769de0a017ed216912e/mage_ai/server/api/projects.py#L1

pyproject.toml
christopherscholz: you are right, this should have been an optional dependency. Ill change it

pyproject.toml
dy46: hmm, i don't think this file is used. let me see if i can remove it so we can remove some dependencies

dy46:
when i try to pip install mage-ai from your branch, the `mage start` command fails with this error:
```
Traceback (most recent call last):
  File "/Users/david_yang/mage/mage-ai/.venv/bin/mage", line 5, in <module>
    from mage_ai.cli.main import app
ModuleNotFoundError: No module named 'mage_ai.cli'
```
i'm planning on looking into it, but if you know how to fix it, you can do that too

christopherscholz:
> when i try to pip install mage-ai from your branch, the `mage start` command fails with this error:
> 
> ```
> Traceback (most recent call last):
>   File "/Users/david_yang/mage/mage-ai/.venv/bin/mage", line 5, in <module>
>     from mage_ai.cli.main import app
> ModuleNotFoundError: No module named 'mage_ai.cli'
> ```
> 
> i'm planning on looking into it, but if you know how to fix it, you can do that too

for me everything works. But i saw that i have a fairly new setuptools version. The console script usage from pyproject.toml was introduced in setuptools==61.0. So I will update the pyproject file accordingly. 

I have installed mage via `pip install -e "./[all]"` in editable mode and can run `mage start` without any issue.

dy46:
I tried adding that line in my own branch, but it didn't seem to resolve the issue.

dy46:
sorry, i haven't had time to push this through. the last time i tested it, both the dev and prod workflows seemed to work properly. let me sync it with master and see if we can finish this up

dy46 (approved):

dy46:
@christopherscholz I asked some of my teammates to try building the docker image from this branch, and they also ran into the `mage_integrations` issue that I ran into which is why I haven't merged this change yet. I was able to resolve it by completely reseting my dev environment, but that doesn't seem like an ideal solution.

PR #3462, opened 2023-09-10
Handle transformer renaming columns not breaking destination data types in integration pipelines

splatcollision:
In case of integration pipeline transformer renaming columns, support an optional 'aliases' field in the schema properties, that maps source column data types to the renamed column data types.

Slack thread: https://mageai.slack.com/archives/C03HTTWFEKE/p1694214865146399

# Description

Since data passed to transform blocks is a pandas DataFrame, many operations are available on each row of your data.

It's possible to do a few dangerous things with this, for example if you have a target destination that needs it's columns renamed to match existing schemas.

In this case, there can be edge case mismatches between the data types in the source columns, and the expected data types in the destination. This is particularly important for SQL columns where the INSERT commands generated by the pipelines will use CAST expressions to try and make sure the writes succeed. However, if a column is renamed in the transformer, we will use our best guess by looking at values to decide what to CAST the column values to.

For example, a TIMESTAMPTZ value of `2023-01-09T20:01:00`, once it's renamed, can look like a string, but using an INSERT with `CAST('2023-01-09T20:01:00' AS TEXT)` to a column with data type `TIMESTAMPTZ` will fail.

In order to fix this, you can add column "aliases" in the pipeline's data integration catalog.

(Examples of transformer block and catalog json changes in doc changes within this PR)

# How Has This Been Tested?

1. I've encountered insert problems when renaming dataframe columns in my transformer blocks. In particular, TIMESTAMPTZ columns attempt to insert as TEXT via the CAST statements.
2. I added 'aliases' property to each schema property in the data integration catalog file.
3. I update the code on my local mage with the change in this PR.
4. I added extra pipeline logging to confirm the alias maps were being created correctly (now removed)
5. I checked the pipeline editing UI, and making changes there does not affect the custom changes in the catalog file.
6. Without having to provide default CASTs in the postgres layer, the integration pipeline INSERTs now CAST as the correct data types, and the inserts succeed without SQL errors.

# Checklist

cc:


docs/design/blocks/transformer.mdx
```diff
@@ -44,3 +44,120 @@ import pandas as pd
 def transform_df(df1, df2, df3, **kwargs) -> DataFrame:
     return pd.concat([df1, df2, df3])
 ```
+
+## Handling column renames
```
wangxiaoyou1993, line 48: this section is for data integration pipeline. could you update the section name to "Handling column renames in data integration pipeline"

mage_ai/data_preparation/models/block/integration/__init__.py
```diff
@@ -246,6 +254,14 @@ def __update_col_names(schema_dict: Dict, properties_key: str, new_columns: List
                                         if k in properties_original else v
                                         for k, v in properties_updated.items()
                                     }
+                                    # Separately, update schema types from alias map
+                                    if len(output_col_aliases.keys()):
+                                        for cname in schema_updated['schema']['properties']:
+                                            if cname in output_col_aliases:
+                                                prop = output_col_aliases[cname]
+                                                schema_updated['schema']['properties'][cname] = prop
+
```
wangxiaoyou1993, line 263: nit: remove extra blank line according to the code quality check (flake8)

mage_ai/data_preparation/models/block/integration/__init__.py
```diff
@@ -246,6 +254,14 @@ def __update_col_names(schema_dict: Dict, properties_key: str, new_columns: List
                                         if k in properties_original else v
                                         for k, v in properties_updated.items()
                                     }
+                                    # Separately, update schema types from alias map
+                                    if len(output_col_aliases.keys()):
+                                        for cname in schema_updated['schema']['properties']:
```
wangxiaoyou1993, line 259: I think it should be `for cname in schema_updated['schema']['properties'].keys()`

splatcollision: Sure, updated

mage_ai/data_preparation/models/block/integration/__init__.py
splatcollision: 👍

docs/design/blocks/transformer.mdx
splatcollision: Absolutely

wangxiaoyou1993 (approved):

PR #3463, opened 2023-09-11
[xh] Integrate DuckDB IO class and SQL block

matrixstone:
# Description
Integrate DuckDB IO class and SQL block

# How Has This Been Tested?

## Load Data From DuckDB
In generic SQL data loader, choose connection as `DuckDB`, profile as `default`. It can run query against duck db.

If `Use raw SQL` is unchecked:
image

A complex example with `creating table`, `inserting data` and `reading data`
image

## Insert Data into DuckDB
This test reads data from Postgres db with a loader then insert the data into DuckDB.
Data exporter block of duckDB is:
image

Data read from Postgres DB:
pg_input1

Data write into DuckDB:
duck_output1

# Checklist

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_ai/api/resources/DataProviderResource.py
+++ b/mage_ai/api/resources/DataProviderResource.py
@@ -22,7 +22,7 @@ DATA_PROVIDERS = [
 DATA_PROVIDERS_NAME = {
     DataSource.BIGQUERY: 'BigQuery',
     DataSource.CLICKHOUSE: 'ClickHouse',
-    DataSource.DUCKDB: 'Duck DB',
+    DataSource.DUCKDB: 'DuckDB',
     DataSource.DRUID: 'Druid',
     DataSource.MSSQL: 'Microsoft SQL Server',
     DataSource.MYSQL: 'MySQL',
--- a/mage_ai/api/resources/DataProviderResource.py
+++ b/mage_ai/api/resources/DataProviderResource.py
@@ -10,6 +10,7 @@ DATA_PROVIDERS = [
     DataSource.BIGQUERY,
     DataSource.CLICKHOUSE,
     DataSource.DRUID,
+    DataSource.DUCKDB,
     DataSource.MSSQL,
     DataSource.MYSQL,
     DataSource.POSTGRES,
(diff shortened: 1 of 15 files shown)


mage_integrations/requirements.txt
```diff
@@ -4,6 +4,7 @@ backoff==2.0.0
 clickhouse_sqlalchemy
 couchbase==4.1.1
 deltalake==0.7.0
+duckdb==0.8.1
```
wangxiaoyou1993, line 7: it should be added to `requirements.txt` instead of `mage_integrations/requirements.txt`
you should also add it to setup.py

mage_ai/data_preparation/models/block/sql/__init__.py
```diff
@@ -279,6 +280,50 @@ def execute_sql_code(
                             verbose=False,
                         ),
                     ]
+    elif DataSource.DUCKDB.value == data_provider:
+        from mage_ai.io.duckdb import DuckDB
+
+        loader = DuckDB.with_config(config_file_loader)
+        schema = schema or 'main'
```
wangxiaoyou1993, line 287: Does user need to enter this schema in the io_config.yaml?

matrixstone: Make sense! I thought user enter schema at Mage UI Block header. It is also a good idea to add it into io_config. I created a `DUCKDB_SCHEMA` in io_config also tested it in the Mage UI.

requirements.txt
```diff
@@ -10,6 +10,7 @@ croniter==1.3.7
 cryptography==36.0.2
 dask>=2022.2.0
 datadog==0.44.0
+duckdb==0.8.1
```
wangxiaoyou1993, line 13: Could you move this package below the "#extras" in this file?

mage_ai/api/resources/DataProviderResource.py
```diff
@@ -21,6 +22,7 @@
 DATA_PROVIDERS_NAME = {
     DataSource.BIGQUERY: 'BigQuery',
     DataSource.CLICKHOUSE: 'ClickHouse',
+    DataSource.DUCKDB: 'Duck DB',
```
wangxiaoyou1993, line 25: nit: call it 'DuckDB'

wangxiaoyou1993 (approved):

PR #3474, opened 2023-09-11
Feat: allow user to overwrite SQL types

Luishfs:
# Description

Feature to allow users to overwrite their SQL types in multipe DBs

This feature aims to allow users to overwrite SQL Table types creation by receiving an external python dictionary containing both the column name and type.

With this, Mage users won't be restricted to internal table generation types or the need to create a table before running any pipeline within the io classes.

IO classes that are being modified:

1.  PostgreSQL
2.  Redshift
3. Trino
4. MsSQL
5. MySQL
6. Clickhouse

# How Has This Been Tested?

Tested on a local postgres instance

Python Data Exporter Block:
Screenshot from 2023-09-11 17-38-04

Python Data Exporter Block logs:
Screenshot from 2023-09-11 17-35-49

Result table:
Screenshot from 2023-09-11 17-36-11

cc:
@wangxiaoyou1993


mage_ai/io/redshift.py
```diff
@@ -225,16 +225,24 @@ def __process():
                     cur.execute(query)
                 else:
                     if should_create_table:
-                        col_with_types = ', '.join(
-                            [f'{col} {col_type}' for col, col_type in columns_with_type],
-                        )
+                        if overwrite_type is not None:
+                            user_mod_columns, col_with_usr_types = get_user_type(overwrite_type)
+
+                        col_with_types = [f'{col} {col_type}' for col, col_type in columns_with_type
```
wangxiaoyou1993, line 231: Let's not change the order of the columns in the create table command.
Same in other databases.

mage_ai/io/clickhouse.py
```diff
@@ -162,15 +165,27 @@ def build_create_table_command(
...
         }
         fields = []
-        for cname in db_dtypes:
-            fields.append(f'{cname} {db_dtypes[cname]}')
+        if user_types is not None:
+            user_mod_columns, col_with_usr_types = get_user_type(user_types)
+
+            for cname in db_dtypes:
+                if cname in user_mod_columns:
+                    pass
```
wangxiaoyou1993, line 182: why do you pass?

wangxiaoyou1993 (approved):

mage_ai/io/redshift.py
```diff
@@ -227,16 +226,26 @@ def __process():
...
+
+                        col_with_types = ', '.join(col_with_types)
                         query = f'CREATE TABLE IF NOT EXISTS {full_table_name} ({col_with_types})'
                         cur.execute(query)
 
                     columns = ', '.join([t[0] for t in columns_with_type])
                     values = [f"""({', '.join([format_value(x) for x in v])})""" for v in df.values]
                     values = ', '.join(values)
                     query = f'INSERT INTO {full_table_name} ({columns})\nVALUES {values}'
+                    with self.printer.print_msg(query):
```
tommydangerous, line 247: @wangxiaoyou1993 @Luishfs why are we printing here and not checking the value of `verbose`?

wangxiaoyou1993: will fix

PR #3477, opened 2023-09-12
[mp] Update pipeline overview screenshot

mattppal:
Quick PR to update the pipeline overview screenshot that appears in a few locations


wangxiaoyou1993:
Why do we update the screenshot?
The new image's resolution is low. And we should put sensor block at the bottom

mattppal:
> Why do we update the screenshot? The new image's resolution is low. And we should put sensor block at the bottom

The old screenshot was really old. Just uploaded a new one

wangxiaoyou1993:
> > Why do we update the screenshot? The new image's resolution is low. And we should put sensor block at the bottom
> 
> The old screenshot was really old. Just uploaded a new one

Can we give the blocks better names in the new screenshot like the old one?

mattppal:
> > > Why do we update the screenshot? The new image's resolution is low. And we should put sensor block at the bottom
> > 
> > 
> > The old screenshot was really old. Just uploaded a new one
> 
> Can we give the blocks better names in the new screenshot like the old one?

Blocks & pipeline renamed :)

wangxiaoyou1993 (approved):

PR #3481, opened 2023-09-12
[jk] Local timezone setting

johnson-mage:
# Description
- Add project setting for displaying datetimes in local timezone. The timezone conversions are all done on the frontend. Dates are still stored in config files and the metadata db in UTC time.
- Certain pages (namely a pipeline's "Monitor" page) and components (e.g. the pipeline run completed vertical bar chart on the "Overview" page) are still in UTC time since those dates are parsed from the UTC datetimes on the backend. Will require a separate PR to convert these dates in the graphs to the local timezone.

# How Has This Been Tested?
image

The dates are in local time (they have a timezone offset):
pages with dates in local time

# Checklist

### DIFF
--- a/mage_ai/api/resources/ProjectResource.py
+++ b/mage_ai/api/resources/ProjectResource.py
@@ -53,11 +53,16 @@ class ProjectResource(GenericResource):
 
         if 'features' in payload:
             for k, v in payload.get('features', {}).items():
-                feature = (repo_config.features or {}).get(k)
-                if (feature and not v) or (not feature and v):
+                if v is not None:
                     if 'features' not in data:
                         data['features'] = {}
                     data['features'][k] = v
+            features = data.get('features')
+            if features is not None:
+                data['features'] = merge_dict(
+                    repo_config.features or {},
+                    features,
+                )
 
         if 'help_improve_mage' in payload:
             if payload['help_improve_mage']:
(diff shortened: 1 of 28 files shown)


mage_ai/api/resources/ProjectResource.py
```diff
@@ -53,11 +53,16 @@ async def update(self, payload, **kwargs):
 
         if 'features' in payload:
             for k, v in payload.get('features', {}).items():
-                feature = (repo_config.features or {}).get(k)
-                if (feature and not v) or (not feature and v):
+                if v is not None:
```
johnson-mage, line 56: @wangxiaoyou1993 I added the `display_local_timezone` feature setting to the project config, but needed to update the logic here cuz when trying to update both the `add_new_block_v2` and `display_local_timezone` settings in one API request, one of the settings would get removed.

For example, if the project `metadata.yaml` config had this setting for `features` below:
```
features:
  add_new_block_v2: true
```
If we make an API request with this payload:
```
{
  "project": {
    "features": {
      "add_new_block_v2": true,
      "display_local_timezone": true
    }
  }
}
```
Then the new config no longer has the `add_new_block_v2` setting, but only the `display_local_timezone` setting. The new `metadata.yaml` `features` config looked like this after the API request:
```
features:
  display_local_timezone: true
```
The changes to this file keep existing settings and make any changes to the "features" settings consistent with what a user would expect. So in the example above, with the new changes, the API request would have updated the config to look like this:
```
features:
  add_new_block_v2: true
  display_local_timezone: true
```

PR #3498, opened 2023-09-14
[dy] Fix how table name is parsed for sql blocks

dy46:
# Description

This PR fixes 2 issues related to schema/table name parsing when creating the upstream block table.

The first issue is that when the `data_provider` was the same for the current block and the upstream block, the schema name was being taken from the upstream block even if the `data_provider_profile` was different. 

The second issue is that the table_name from the upstream block can potentially be `database_name.table_name`, and the `table_name_parts` function was not accounting for that and just using the entire string as the table name.

# How Has This Been Tested?

- [x] Tested locally with postgres and mysql blocks

# Checklist

cc: @wangxiaoyou1993


wangxiaoyou1993 (approved):
Could you also update the method docstring?

PR #3521, opened 2023-09-16
Fix: paths include posix like forward slash in many instances, which will not work on non-posix systems

### LINKED ISSUE #3518: paths include posix like forward slash in many instances, which will not work on non-posix systems
**Describe the bug**
in the code in many instances a path is created or defined using posix like `/` syntax.
Instead the path should be joined atomically e.g. `os.path.join('parent', 'child', 'file')` instead of `os.path.join('parent', 'child/file')`.
Also when resolving the user dir using something like `~/path` it must be joined using e.g. `os.path.join('~', 'path')`
Only when writing a path to the .gitconfig you need to use posix in any case as git expects also paths on windows to be written using forward slashes.

**To Reproduce**
Run unittests on windows

**Expected behavior**
paths should work on all systems.

**Desktop (please complete the following information):**
 - OS: Windows
 - Version 0.9.24

christopherscholz:
# Description
Resolves #3518

replaced many instances. There are probably more.
WHen using paths as names (e.g. uuids for blocks) this should probably be handled more strict. e.g. always use forward slash in these cases. Also rather use posixpath instead of manually doing some concatenating of strings to form a path.

Fixed gitconfig for windows users. Running unittest on windows will not destroy the .gitconfig anymore.

# How Has This Been Tested?
- [x] Unittest

# Checklist

### DIFF
--- a/mage_ai/orchestration/db/__init__.py
+++ b/mage_ai/orchestration/db/__init__.py
@@ -32,7 +32,7 @@ elif not db_connection_url:
     else:
         if is_test():
             db_connection_url = f'sqlite:///{TEST_DB}'
-        elif os.path.exists('mage_ai/orchestration/db/'):
+        elif os.path.exists(os.path.join('mage_ai', 'orchestration', 'db')):
             # For local dev environment
             db_connection_url = 'sqlite:///mage_ai/orchestration/db/mage-ai.db'
         elif os.path.exists('mage-ai.db'):
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,6 +16,7 @@ repos:
         types: [python]
     -   id: name-tests-test
         types: [python]
+        exclude: "base_test"
         args:
           - "--pytest-test-first"
     -   id: requirements-txt-fixer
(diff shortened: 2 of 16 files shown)


mage_ai/orchestration/db/__init__.py
```diff
@@ -32,7 +32,7 @@
     else:
         if is_test():
             db_connection_url = f'sqlite:///{TEST_DB}'
-        elif os.path.exists('mage_ai/orchestration/db/'):
+        elif os.path.exists(os.path.join('mage_ai', 'orchestration', 'db')):
             # For local dev environment
             db_connection_url = 'sqlite:///mage_ai/orchestration/db/mage-ai.db'
```
dy46, line 37: we should probably change this as well

dy46 (commented):
thanks for doing this!

mage_ai/orchestration/db/__init__.py
christopherscholz: the sqlite path is using a file uri, which i don't like, but should probably be fine: https://en.wikipedia.org/wiki/File_URI_scheme and e.g. https://docs.sqlalchemy.org/en/20/core/engines.html#sqlite

dy46 (approved):

PR #3553, opened 2023-09-20
[xh] Add DuckDB loader and exporter

matrixstone:
# Description
This PR includes
- Add DuckDB loader and exporter
- Convert array type into text in duckdb

# How Has This Been Tested?
Loader template and loaded data:
duck_loader
duck_load_data

Loader template and exported data:
duckdb_exporter
duck_data_exporter

Write Int array, String array, Json in duckdb
duck_db_export_complicated_type

# Checklist

cc:
@wangxiaoyou1993

### DIFF
--- a/docs/design/data-loading.mdx
+++ b/docs/design/data-loading.mdx
@@ -8,6 +8,7 @@ Mage provides data loading clients that simplify loading and exporting data in y
 -   [AWS Redshift](#redshift)
 -   [AWS S3](#s3)
 -   [Azure Blob Storage](#azureblobstorage)
+-   [DuckDB](#duckdb)
 -   [Filesystem](#fileio)
 -   [Google BigQuery](#bigquery)
 -   [Google Cloud Storage](#googlecloudstorage)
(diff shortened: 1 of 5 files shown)


docs/design/data-loading.mdx
```diff
@@ -11,6 +11,7 @@ Mage provides data loading clients that simplify loading and exporting data in y
 -   [Filesystem](#fileio)
 -   [Google BigQuery](#bigquery)
 -   [Google Cloud Storage](#googlecloudstorage)
+-   [DuckDB](#duckdb)
```
wangxiaoyou1993, line 14: nit: alphabetical order

mage_ai/io/duckdb.py
```diff
@@ -89,10 +89,7 @@ def get_type(self, column: Series, dtype: str) -> str:
             PandasTypes.UNKNOWN_ARRAY,
             PandasTypes.COMPLEX,
         ):
-            raise BadConversionError(
-                f'Cannot convert column \'{column.name}\' with data type \'{dtype}\' to '
-                'a DuckDB datatype.'
-            )
+            return 'text'
```
wangxiaoyou1993, line 92: Should it use upper case?

wangxiaoyou1993 (approved):

mage_ai/io/duckdb.py
matrixstone: I tried so both upper case and lower case works. But I update here to be upper case to be consistent with other type string.

Issue #3566, opened 2023-09-21
Export to redshift doesn't handle datetime


RobinFrcd:
Hi,
I'm trying to export a dataframe with `datetime64[ns]` dtype but it seems like the [`convert_python_type_to_redshift_type`](https://github.com/mage-ai/mage-ai/blob/81218015ab8aa0f0010df10344ac311aa47f0ba3/mage_ai/shared/utils.py#L77) converter does not support `datetime`.
My exported redshift table has a `VARCHAR` instead of `TIMESTAMP`.

RobinFrcd:
@dy46 lol sorry, this one wasn't meant !

PR #3567, opened 2023-09-21
Support TIMESTAMP in redshift convert

RobinFrcd:
Fix https://github.com/mage-ai/mage-ai/issues/3566

### DIFF
--- a/mage_ai/shared/utils.py
+++ b/mage_ai/shared/utils.py
@@ -81,6 +81,8 @@ def convert_python_type_to_redshift_type(python_type):
         return 'FLOAT8'
     elif python_type is bool:
         return 'BOOLEAN'
+    elif python_type is datetime:
+        return 'TIMESTAMP'
     return 'VARCHAR'
 
 


wangxiaoyou1993:
Have you tested this change?

RobinFrcd:
@wangxiaoyou1993 
I've tested up to 
```
[(
    col,
    convert_python_type_to_redshift_type(
        convert_pandas_dtype_to_python_type(df.dtypes[col]),
    ),
) for col in columns]
```

And it works fine. I've also checked the [documentation](https://docs.aws.amazon.com/redshift/latest/dg/r_Datetime_types.html#r_Datetime_types-timestamp) and TIMESTAMP seems to be what we needed.
However, doing and e2e test on my environment with redshift is not that easy so I didn't do a full test, no, sorry.

wangxiaoyou1993:
@Luishfs will help test out this change.

wangxiaoyou1993 (approved):
I've tested this change. It works.

PR #3591, opened 2023-09-25
Scheduler trigger interval as environment variable setting

PopaRares:
# Description
Added a new setting called SCHEDULER_TRIGGER_INTERVAL that sets the LoopTimeTrigger's sleep time between scheduler calls. Value is of type float. Default value is kept at 10 seconds.
Example call:
	 
	 SCHEDULER_TRIGGER_INTERVAL=3 ./scripts/dev.sh

# How Has This Been Tested?
Set the **SCHEDULER_TRIGGER_INTERVAL** to some numerical value (default value is 10 seconds).

	SCHEDULER_TRIGGER_INTERVAL=3.5 ./scripts/dev.sh

# Checklist

cc:
@adelcast 
@hugabora


PopaRares:
@wangxiaoyou1993 I've been looking at the classes LoopTimeTrigger and TimeTrigger and I have some light concerns. Should the trigger_interval property be held in the TimeTrigger class? How things work currently is that LoopTimeTrigger does the waiting and TimeTrigger triggers the scheduler; going by that logic, the trigger_interval property should be a property of the entity that does the actual waiting: LoopTimeTrigger. Expandig from this, what are the advantages of having LoopTimeTrigger and TimeTrigger be 2 separate classes? TimeTrigger is only used with the loop part attached so maybe a good solution would be to merge these two? Let me know what you guys think!

wangxiaoyou1993 (approved):

PR #3630, opened 2023-09-29
[dy] Raise exception if there are serialization errors

dy46:
# Description

Raise exception when fetching input variables if parquet or json variable cannot be read.

Screenshot 2023-10-04 at 3 39 38 PM

# How Has This Been Tested?

- [x] Tested handling an exception when serializing both json and parquet input variables in notebook
- [x] Tested when running trigger

# Checklist

cc: @wangxiaoyou1993


wangxiaoyou1993:
Could you add more details to the test section?

mage_ai/data_preparation/storage/local_storage.py
```diff
@@ -39,22 +39,36 @@ def remove(self, path: str) -> None:
     def remove_dir(self, path: str) -> None:
         shutil.rmtree(path, ignore_errors=True)
 
-    def read_json_file(self, file_path: str, default_value: Dict = None) -> Dict:
+    def read_json_file(
```
wangxiaoyou1993, line 42: Could you also update S3Storage and GCSStorage? Otherwise, it'll throw exceptions when using those two storage classes.

mage_ai/data_preparation/storage/local_storage.py
```diff
@@ -39,22 +39,36 @@ def remove(self, path: str) -> None:
...
+        file_path: str,
+        default_value: Dict = None,
+        raise_exception: bool = False,
+    ) -> Dict:
         if not self.path_exists(file_path):
             return default_value or {}
         async with aiofiles.open(file_path, mode='r') as file:
             try:
-                return json.loads(await file.read())
+                return json.load(await file.read())
```
wangxiaoyou1993, line 68: Why do we update it to json.load?

dy46: oops i didn't see that, will fix

mage_ai/data_preparation/storage/local_storage.py
dy46: I made it so that the behavior doesn't change for those classes. The default value of `raise_exception` is `False`, so the exceptions shouldn't be thrown in those cases.

wangxiaoyou1993: Have you tested it? 
The S3Storage class doesn't support `raise_exception` param in the `read_json_file` method right now. If you call it, I suppose it'll throw exception: https://github.com/mage-ai/mage-ai/blob/master/mage_ai/data_preparation/storage/s3_storage.py#L53

dy46: oh sorry, i thought you were talking about something else. not sure i missed those, i'll update them

dy46:
@wangxiaoyou1993 updated

wangxiaoyou1993 (approved):

PR #3659, opened 2023-10-03
[jk] Project-wide pipeline run filters

johnson-mage:
# Description
- Allow multiple filters on the project-wide Pipeline Runs page for `status`, `pipeline uuid`, and `pipeline tag`.

# How Has This Been Tested?
pipeline run filters with tags

# Checklist


mage_ai/cache/tag.py
```diff
@@ -23,6 +27,34 @@ async def initialize_cache(self, replace: bool = False) -> 'TagCache':
     def get_tags(self) -> Dict:
         return self.get(self.cache_key) or {}
 
+    def get_tags_by_pipeline_uuid(self) -> Dict:
```
johnson-mage, line 30: We needed a cache for the `pipeline_uuid_to_tags_mapping` so that we can quickly filter pipeline runs by pipeline tags and also to display the pipeline tags for each pipeline run in the Pipeline Runs table. Not sure if it'd be better to create an entirely different cache class, but I added this method here since it was related to the `TagCache`.

mage_ai/data_preparation/models/pipeline.py
```diff
@@ -1013,6 +1013,14 @@ async def update(self, data, update_content=False):
 
             cache = await TagCache.initialize_cache()
 
+            # Update tag cache if tag was removed from pipeline
+            tags_cached = cache.get_tags_for_specific_pipeline(self)
+            tags_removed = set(tags_cached) - set(self.tags)
+            if len(tags_removed) > 0:
+                for tag_uuid in tags_removed:
+                    if old_uuid is None:
+                        cache.remove_pipeline(tag_uuid, self.uuid)
```
johnson-mage, line 1022: I noticed fetching pipelines with tag filters in the pipelines list wasn't getting updated immediately when a tag was removed from a pipeline, so this fixes that. Calling the `remove_pipeline` method also clears the `pipeline_uuid_to_tags_mapping` cache, so that an outdated cache would not be used for the Pipeline Runs table.

For example in the screenshot, the filter for tags with the `integration` tag was applied, but a pipeline (`cloud_postgres_to_mysql`) that no longer has that tag was still included. You can see that it doesn't have the `integration` tag in the "Tags" column.
image

mage_ai/api/resources/PipelineRunResource.py
```diff
@@ -114,20 +130,44 @@ async def process_collection(self, query_arg, meta, user, **kwargs):
...
-        if pipeline_type is not None:
-            pipeline_type_by_pipeline_uuid = dict()
+        filter_by_pipeline_type = pipeline_type is not None
+        filter_by_pipeline_tag = pipeline_tags is not None
+        if filter_by_pipeline_type or filter_by_pipeline_tag:
             try:
+                pipeline_type_by_pipeline_uuid = dict()
+                if filter_by_pipeline_tag:
+                    pipeline_tag_cache = TagCache()
+                    pipeline_tags_by_pipeline_uuid = \
```
wangxiaoyou1993, line 152: This way of filtering pipeline runs by tags is not efficient.
Here is the recommend way
* Filter the pipelines by tags using the TagCache and get the pipeline uuids
* Filter the pipeline runs by the pipeline uuids

mage_ai/cache/tag.py
wangxiaoyou1993: This method is actually not needed if you implement the pipeline run filtering by tags with my recommended way.

mage_ai/api/resources/PipelineRunResource.py
johnson-mage: Updated query based on what you suggested

mage_ai/cache/tag.py
johnson-mage: @wangxiaoyou1993 I added a property `pipeline_tags` that uses this method to the PipelineRun [model](https://github.com/mage-ai/mage-ai/pull/3659/files#diff-81ce2b1693f591260806fa018d2c1f03076f636fafbc682bc6318d6d64d002cdR622) so that we can display the pipeline tags in the PipelineRuns table. I considered not displaying it in the table, but it seemed necessary to provide some assurance that the pipeline tag filter was properly filtering the pipeline runs and so that users can see which runs had certain pipeline tags if multiple tags were being filtered.
 
image

wangxiaoyou1993: You can just get the tags by pipeline using `pipeline.tags`

johnson-mage: Gotcha, i was overthinking it.

mage_ai/api/resources/PipelineRunResource.py
```diff
@@ -199,6 +239,10 @@ async def process_collection(self, query_arg, meta, user, **kwargs):
             'next': has_next,
         }
 
+        if include_pipeline_uuids:
```
wangxiaoyou1993, line 242: qq: what's the purpose of returning all pipeline uuids?

johnson-mage: I need it to display to users so they can filter the project-wide pipeline runs by multiple pipelines, which actually seems useful (e.g. user just wants to see the pipeline runs for 2 or 3 pipelines).

johnson-mage: image

mage_ai/api/resources/PipelineRunResource.py
```diff
@@ -108,26 +140,34 @@ def collection(self, query_arg, meta, user, **kwargs):
...
+                pipeline_type_by_pipeline_uuid = dict()
                 pipeline_runs = total_results.all()
                 results = []
                 for run in pipeline_runs:
-                    if run.pipeline_uuid not in pipeline_type_by_pipeline_uuid:
-                        pipeline_type_by_pipeline_uuid[run.pipeline_uuid] = run.pipeline_type
-                    run_pipeline_type = pipeline_type_by_pipeline_uuid[run.pipeline_uuid]
-                    if run_pipeline_type == pipeline_type:
+                    filters = []
+                    if filter_by_pipeline_type:
```
wangxiaoyou1993, line 165: You already have a `if filter_by_pipeline_type` above. It's a duplicate check here.

wangxiaoyou1993 (approved):

PR #3691, opened 2023-10-05
[mp] Update ETL demo and remove Postgres requirement

mattppal:
# Description

This PR overhauls the formatting of our ETL demo and removes the need for Postgres as a destination by writing to a DuckDB database.

It also creates a quickstart that allows users to get up and running with a single command:

```
git clone https://github.com/mage-ai/etl-demo mage-etl-demo \
&& cd mage-etl-demo \
&& cp dev.env .env && rm dev.env \
&& docker compose up
```

I removed the visualization functionality from this demo, as I feel it is tangential to Mage's core functionality and only serves to distract from orchestration/transformation. I think we could add a separate, in-depth tutorial on creating visualizations with charts.

# How Has This Been Tested?
- [x] `mintlify dev`

# Checklist


wangxiaoyou1993:
I saw two images added in this PR. Where do we use them?

mattppal:
Something's up with my git setup. It's duplicating images and adding a `2`, I haven't figured out why yet. I'll delete these. Nice catch

docs/guides/load-api-data.mdx
```diff
@@ -2,294 +2,203 @@
...
-        />
-    </Frame>
+## Tutorial
 
----
+<AccordionGroup>
+    <Accordion title="1️⃣ Create a pipeline">
+        Open the pipelines list page (`/pipelines`). This is the default page when navigating to Mage in your web browser. If you have Mage running, just click [here](https://localhost:6789/pipelines).
 
-## `3.` Visualize data
```
wangxiaoyou1993, line 97: why don't we keep the "Visualize data" section?

mattppal: I think it's tangential to the process of ETL. I haven't noticed a ton of users engaging with charts. 

I think we could make a separate guide that walks though charts, but for this demo my goal was to get users on board transforming data as fast as possible.

wangxiaoyou1993: ok, got it. we can make a separate guide for charts then.

wangxiaoyou1993 (approved):

docs/guides/load-api-data.mdx
mattppal: I'll make a to-do

PR #3696, opened 2023-10-06
[jk] Cron expression syntax err display for triggers in code

johnson-mage:
# Description
- Display error popup if trigger in code has syntax error for cron expression. Note that this checks if the cron syntax is invalid, not if the `triggers.yaml` config file has proper yaml syntax. For example, `schedule_interval: */15 * * * *` may not be parsed correctly as yaml because `*/15 * * * *` requires quotes around it.

# How Has This Been Tested?
cron expression syntax error

# Checklist

### DIFF
--- a/mage_ai/data_preparation/models/triggers/__init__.py
+++ b/mage_ai/data_preparation/models/triggers/__init__.py
@@ -69,6 +69,7 @@ class Trigger(BaseConfig):
 
     @property
     def has_valid_schedule_interval(self) -> bool:
+        # Check if trigger has valid cron expression
         if self.schedule_interval is not None and \
             self.schedule_type == ScheduleType.TIME and \
             self.schedule_interval not in [e.value for e in ScheduleInterval] and \
--- a/mage_ai/data_preparation/models/triggers/__init__.py
+++ b/mage_ai/data_preparation/models/triggers/__init__.py
@@ -6,6 +6,7 @@ from datetime import datetime
 from typing import Dict, List
 
 import yaml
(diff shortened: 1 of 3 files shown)


mage_ai/data_preparation/models/triggers/__init__.py
```diff
@@ -66,6 +67,17 @@ def __post_init__(self):
...
             raise Exception(f'Please provide valid env values inside {list(VALID_ENVS)}.')
 
+    @property
+    def has_valid_schedule_interval(self) -> bool:
+        # Check if trigger has valid cron expression
+        if self.schedule_interval is not None and \
+            self.schedule_type == ScheduleType.TIME and \
+            self.schedule_interval not in [e.value for e in ScheduleInterval] and \
+                not croniter.is_valid(self.schedule_interval):
+            return False
```
dy46, line 77: might be good to return False if `schedule_type == TIME` and `schedule_interval` is None

dy46 (approved):

mage_ai/data_preparation/models/triggers/__init__.py
johnson-mage: I tested out adding a check for this as well, but decided not to include it, since the main issue with the invalid cron syntax in the `triggers.yaml` file was that an error would keep occurring on the server and prevent the trigger from actually being created (and would be hard to detect from the UI). A null `schedule_interval` is still able to create the pipeline_schedule without any issues, so the error popup isn't as critical to show.

PR #3703, opened 2023-10-06
[ac] Replace runtime variables card list with table

anniexcheng:
# Description

Previously, runtime variables were displayed in a horizontally-scrollable card list like so: 
before copy

Feedback was that "it is hard to see all the runtime variables the way it is now, and you have to scroll a lot too." This PR replaces the card list with a table instead ([Airtable Task](https://airtable.com/applEuqJsK4zF5yJK/tbl5bIlFkk2KhCpL4/viwCmaGgcnwD9IqOM/recrumg67BH0zbgaB?blocks=hide)):
image
* If there are more than 5 variables, only 5 variables will be visible in the table, and there will be a scrollbar to vertically scroll and see more variables.
* If there are ≤ 5 variables, all variables will be visible in the table, and there will be no scrollbar.

# How Has This Been Tested?

1. Add global variables here: http://localhost:3000/pipelines/example_pipeline/edit?sideview=variables
2. Test when there's no variables
3. Test when there's ≤ 5 variables
4. Test when there's > 5 variables

**Before:**
before

**After**

- [ ] No variables
    image
- [ ] ≤ 5 variables
    image
- [ ] > 5 variables
    image

cc:

### DIFF
--- a/mage_ai/frontend/components/RuntimeVariables/index.style.tsx
+++ b/mage_ai/frontend/components/RuntimeVariables/index.style.tsx
@@ -1,28 +1,12 @@
 import styled from 'styled-components';
 
 import dark from '@oracle/styles/themes/dark';
-import { BORDER_RADIUS } from '@oracle/styles/units/borders';
-import { PADDING, UNIT } from '@oracle/styles/units/spacing';
-import { ScrollbarStyledCss } from '@oracle/styles/scrollbars';
+import { PADDING } from '@oracle/styles/units/spacing';
 
-export const ContainerStyle = styled.div`
-  border-bottom: 1px solid ${dark.borders.medium};
-  padding: ${PADDING}px;
-`;
-
-export const CardsStyle = styled.div`
-  ${ScrollbarStyledCss}
-
-  height: 80px;
-
+export const ContainerStyle = styled.div<{ $height: number }>`
   display: flex;
-  overflow-x: scroll;
-`;
-
-export const VariableCardStyle = styled.div`
-  background-color: ${dark.background.output};
-  border-radius: ${BORDER_RADIUS}px;
-  flex-shrink: 0;
-  margin-right: ${UNIT}px;
+  flex-direction: column;
+  height: ${props => props.$height}px;
   padding: ${PADDING}px;
+  border-bottom: 1px solid ${dark.borders.medium};
 `;
--- a/mage_ai/frontend/oracle/components/Table/index.style.tsx
+++ b/mage_ai/frontend/oracle/components/Table/index.style.tsx
@@ -3,17 +3,19 @@ import styled from 'styled-components';
 import light from '@oracle/styles/themes/light';
 import { BORDER_RADIUS } from '@oracle/styles/units/borders';
 import { UNIT } from '@oracle/styles/units/spacing';
+import { ScrollbarStyledCss } from '@oracle/styles/scrollbars';
 import { transition } from '@oracle/styles/mixins';
 
 export const PADDING_SIZE_UNITS = 1.5;
 
 export const TableStyle = styled.div<any>`
+  ${ScrollbarStyledCss}
+
   overflow-y: auto;
   position: relative;
   width: 100%;
   z-index: 3;
-  border-bottom-left-radius: ${BORDER_RADIUS}px;
-  border-bottom-right-radius: ${BORDER_RADIUS}px;
+  border-radius: ${BORDER_RADIUS}px;
 
   ${props => `
     background-color: ${(props.theme.background || light.background).page};
@@ -21,9 +23,11 @@ export const TableStyle = styled.div<any>`
   ${props => props.height && `
     height: ${props.height}px;
   `}
-
   ${props => props.flex && `
     flex: 1;
+  `}
+  ${props => props.noBorder && `
+    border-radius: 0;
   `}
 `;
 
--- a/mage_ai/frontend/pages/pipelines/[pipeline]/triggers/index.tsx
+++ b/mage_ai/frontend/pages/pipelines/[pipeline]/triggers/index.tsx
@@ -197,25 +197,39 @@ function PipelineSchedules({
   const buildSidekick = useMemo(() => {
     const variablesOverride = selectedSchedule?.variables;
     const hasOverride = !isEmptyObject(variablesOverride);
-
-    const showVariables = hasOverride
-      ? selectedSchedule?.variables
-      : !isEmptyObject(variablesOrig) ? variablesOrig : null;
+    const hasVariables = !isEmptyObject(variablesOrig);
 
     return (props: DependencyGraphProps) => {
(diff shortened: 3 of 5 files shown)


mage_ai/frontend/components/RuntimeVariables/index.style.tsx
```diff
@@ -1,28 +1,12 @@
...
-  overflow-x: scroll;
-`;
-
-export const VariableCardStyle = styled.div`
-  background-color: ${dark.background.output};
-  border-radius: ${BORDER_RADIUS}px;
-  flex-shrink: 0;
-  margin-right: ${UNIT}px;
+  flex-direction: column;
+  height: ${props => props.$height}px;
```
anniexcheng, line 9: We set a fixed height here so that overflow for the table will be triggered with > 5 variables.

mage_ai/frontend/oracle/components/Table/index.style.tsx
```diff
@@ -3,28 +3,32 @@ import styled from 'styled-components';
...
 
 export const TableStyle = styled.div<any>`
+  ${ScrollbarStyledCss}
+
   overflow-y: auto;
   position: relative;
   width: 100%;
   z-index: 3;
-  border-bottom-left-radius: ${BORDER_RADIUS}px;
-  border-bottom-right-radius: ${BORDER_RADIUS}px;
```
anniexcheng, line 16: I figured it was visually more consistent to have all the corners rounded or not instead of just the bottom corners. Let me know if there was a specific reason for just rounding the bottom ones.

`noBorder = false`:
image

`noBorder = true`:
image

mage_ai/frontend/pages/pipelines/[pipeline]/triggers/index.tsx
```diff
@@ -197,25 +197,39 @@ function PipelineSchedules({
   const buildSidekick = useMemo(() => {
     const variablesOverride = selectedSchedule?.variables;
     const hasOverride = !isEmptyObject(variablesOverride);
-
-    const showVariables = hasOverride
-      ? selectedSchedule?.variables
-      : !isEmptyObject(variablesOrig) ? variablesOrig : null;
+    const hasVariables = !isEmptyObject(variablesOrig);
```
Comment (anniexcheng) [inline mage_ai/frontend/pages/pipelines/[pipeline]/triggers/index.tsx:200]: The code in `RuntimeVariables` indicates that a variable override is only relevant if the variable exists in `variablesOrig`, so I simplified the logic here to only check if original variables are set.

mage_ai/frontend/pages/pipelines/[pipeline]/triggers/index.tsx
```diff
@@ -197,25 +197,39 @@ function PipelineSchedules({
...
+      if (hasVariables) {
+        const maxVisibleRows = 5;
+        const rowHeight = 44;
+        const numVariables = Object.keys(variablesOrig).length;
+        const numVisibleRows = Math.min(maxVisibleRows, numVariables);
+        // This accounts for title + spacing + table in RuntimeVariables
+        runtimeVariablesHeight = 116 + numVisibleRows * rowHeight;
+      }
+   
+      const dependencyGraphHeight = props.height - runtimeVariablesHeight;
```
Comment (anniexcheng) [inline mage_ai/frontend/pages/pipelines/[pipeline]/triggers/index.tsx:219]: A bit hacky but height was already previously hardcoded, so this was a minimal solution I could think of without a much larger refactor.

anniexcheng:
@johnson-mage I was mainly testing with the Runtime Variables section on the Triggers page but let me know if there's other places I should be testing as well!

mage_ai/frontend/components/RuntimeVariables/index.style.tsx
```diff
@@ -1,28 +1,12 @@
...
-  padding: ${PADDING}px;
-`;
-
-export const CardsStyle = styled.div`
-  ${ScrollbarStyledCss}
-
-  height: 80px;
+import { PADDING } from '@oracle/styles/units/spacing';
 
+export const ContainerStyle = styled.div<{ $height: number }>`
```
johnson-mage, line 6: @anniexcheng jw is there any significance behind the `$` in front of the variable name? Versus just using `height`? In terms of consistency with the rest of our prop names, I don't think we name any other FE variables with the `$` in front.

anniexcheng: Yea, the difference is that with `$height`, `height` won't show up in the DOM but without it, it will! Here's the longer explanation from the [docs](https://styled-components.com/docs/basics#passed-props).

`$height`
image

`height`
image

I'm down with either approach though! If we want consistency, I can switch to `height`.

johnson-mage: Ah, is there a situation where it'd not be ideal to show `height` in the DOM? If it is innocuous, we can remove the `$` for consistency, but good to know.

mage_ai/frontend/oracle/components/Table/index.style.tsx
```diff
@@ -3,28 +3,32 @@ import styled from 'styled-components';
 import light from '@oracle/styles/themes/light';
 import { BORDER_RADIUS } from '@oracle/styles/units/borders';
 import { UNIT } from '@oracle/styles/units/spacing';
+import { ScrollbarStyledCss } from '@oracle/styles/scrollbars';
```
johnson-mage, line 6: Alphabetize imports.

johnson-mage:
@anniexcheng Discussed with kai for some suggestions on styling. Can you please do the following?
1. Remove the `Runtime variables` text, rename `Variable` in the column header to `Runtime variable` instead, and make the table fill that entire section.
2. Move the Runtime variables section below the dependency graph.
3. Replace the `light` theme with `dark` in the `mage_ai/frontend/oracle/components/Table/index.style.tsx`, and replace the `background-color` in `ColumnHeaderRowStyle` with the `panel: '#232429'` color from the `mage_ai/frontend/oracle/styles/themes/dark.ts` file.
4. Add the variables count to the column header (e.g. `Runtime variables (5)`).

johnson-mage:
@anniexcheng When the PR is ready for me to take another look, you can re-request review so i get the notification. thanks!

mage_ai/frontend/oracle/components/Table/index.style.tsx
anniexcheng: Ah, is it based on the imported file name (e.g. `'@oracle/styles/scrollbars'`) vs. the imported item name (`ScrollbarStyledCss`)?

johnson-mage: Based on the variable name, so `ScrollbarStyledCss`  would come before `UNIT`. Not a huge deal though. And we did look into adding an eslint rule, but i think it had organized the imports slightly different than how we do it so we didn't use it. If you know of a way to do it, please feel free to add!

johnson-mage (approved):

PR #3785, opened 2023-10-17
[ac] Show server time in app

anniexcheng:
# Description

This PR adds a new current time display in the top right of the app header ([Airtable task](https://airtable.com/applEuqJsK4zF5yJK/tbl5bIlFkk2KhCpL4/viwCmaGgcnwD9IqOM/recTr1owTCG5JOVCx?blocks=hide)).  By default, the time display shows in the UTC format. If you click on the time display, a dropdown menu shows up:
image

The dropdown displays the current time in both UTC format and your local timezone format. You have the option to toggle between showing UTC vs. local time format AND between showing vs. hiding the seconds. You can click anywhere outside the dropdown to dismiss it.

# How Has This Been Tested?

- [x] Current time display shows up in the top right of the app header
  image
- [x] Dropdown menu shows up when the time display is clicked
  image
- [x] Seconds are shown in the time display if "Include seconds" is checked
  image
- [x] Local time is displayed if "Show as local time" is checked
   image

cc: @johnson-mage


mage_ai/frontend/components/ServerTimeDropdown/utils.ts
```diff
@@ -0,0 +1,34 @@
+import {
```
johnson-mage, line 1: @anniexcheng Let's move this file to the `mage_ai/frontend/storage` directory and rename the file, since we keep most of our local storage utility methods there.

mage_ai/frontend/components/ServerTimeDropdown/utils.ts
```diff
@@ -0,0 +1,34 @@
+import { 
+  LOCAL_STORAGE_KEY_DISPLAY_LOCAL_SERVER_TIME, LOCAL_STORAGE_KEY_INCLUDE_SERVER_TIME_SECONDS, 
+} from '@storage/constants';
+import { get, set } from '@storage/localStorage';
+
+function getLocalStorageValue(storageKey: string): boolean {
+  const storageValue = get(storageKey, null);
```
johnson-mage, line 7: If you just called `get(storageKey, false)` which returns a default `false` value, would this `getLocalStorageValue` method still be needed?

mage_ai/frontend/components/ServerTimeDropdown/index.tsx
```diff
@@ -0,0 +1,157 @@
+import { useCallback, useEffect, useState } from 'react';
+
+import ClickOutside from '@oracle/components/ClickOutside';
+import ServerTimeButton from './ServerTimeButton';
+import Text from '@oracle/elements/Text';
+import ToggleSwitch from '@oracle/elements/Inputs/ToggleSwitch';
+
+import { abbreviatedTimezone, currentTimes, TimeZoneEnum, TIME_ZONE_NAMES } from '@utils/date';
```
johnson-mage, line 8: alphabetize

mage_ai/frontend/components/ServerTimeDropdown/index.style.tsx
```diff
@@ -0,0 +1,71 @@
...
+const SHARED_CELL_PADDING = '12px 20px';
+const SHARED_GAP = '1px';
+
+export const DropdownContainerStyle = styled.div<{
+  top?: number;
+}>`
+  position: absolute;
+  top: ${props => props.top || 0};
+  right: 0;
+  width: 360px;
```
johnson-mage, line 14: If a unit value is a multiple of 8, we usually import `UNIT` from `@oracle/styles/units/spacing` and multiply that by a number to come up with the total value. For example, we might put `width: ${UNIT * 45}px;` here instead. The `UNIT` of 8 is basically the standard pixel unit Mage uses throughout the app.

mage_ai/frontend/components/ServerTimeDropdown/index.style.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  display: flex;
+  flex-direction: column;
+  gap: ${SHARED_GAP};
+
+  overflow: hidden;
+  background-color: ${props => (props.theme.borders || dark.borders).darkLight};
+  border-radius: 8px;
```
johnson-mage, line 22: You can also import `BORDER_RADIUS` from `@oracle/styles/units/borders`.

mage_ai/frontend/components/ServerTimeDropdown/index.style.tsx
```diff
@@ -0,0 +1,71 @@
...
+`;
+
+export const DropdownHeaderStyle = styled.div`
+  padding: ${SHARED_CELL_PADDING};
+
+  background-color: ${props => (props.theme.background || dark.background).panel};
+`;
+
+export const TimeListContainerStyle = styled.div`
+  height: 160px;
```
johnson-mage, line 33: can also put `${UNIT * 20}` instead of 160

mage_ai/frontend/components/ServerTimeDropdown/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  return storageValue || false;
+}
+
+function setLocalStorageValue(storageKey: string, value: boolean): boolean {
```
johnson-mage, line 12: We could possibly move this method to the `mage_ai/frontend/storage/localStorage.ts` file, since it seems more generalized and something we could use again in the future.

mage_ai/frontend/utils/date.ts
```diff
@@ -135,6 +150,40 @@ export function utcNowDate(opts?: { dateObj?: boolean }): any {
...
+  includeSeconds?: boolean;
+}) {
+  const currentMoment = tzMoment.utc();
+  const zoneTimes = new Map(
+    timeZones.map((timeZone) => {
+      let moment = currentMoment;
+      switch (timeZone) {
+        case TimeZoneEnum.LOCAL:
+          moment = currentMoment.local();
+        case TimeZoneEnum.UTC:
```
johnson-mage, line 168: Is this case clause still needed since it is empty?

mage_ai/frontend/components/ServerTimeDropdown/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+          mountedCallback={setDropdownPosition}
+          onClick={handleButtonClick}
+          time={times.get(defaultTimeZone)}
+          timeZone={abbreviatedTimezone(defaultTimeZone)}
+        />
+
+        {!isSmallBreakpoint && showDropdown && (
+          <DropdownContainerStyle top={top}>
+            <DropdownHeaderStyle>
+              <Text bold muted uppercase>Server Time</Text>
```
johnson-mage, line 115: I know we initially were thinking of `Server time` like in the designs, but does `Current time` feel more accurate? e.g. In the case of the server having some non-UTC and non-local timezone, "Server time" may not accurately reflect what the server time actually is.

mage_ai/frontend/components/ServerTimeDropdown/utils.ts
anniexcheng: ah true, that's a good simplification

mage_ai/frontend/components/ServerTimeDropdown/index.tsx
anniexcheng: could you clarify which portion should be alphabetized? If it's just the imports on line 8, my thought process was to alphabetize regardless of casing:

1. abbreviatedtimezone
2. currenttimes
3. timezoneenum
4. timezonenames

mage_ai/frontend/utils/date.ts
anniexcheng: It's not technically needed but I just left it there for increased clarity that UTC is the default in the case that we add a lot more time zones here!

mage_ai/frontend/components/ServerTimeDropdown/index.tsx
johnson-mage: Oh I meant this entire line should go below the `import { PURPLE2 } from '@oracle/styles/colors/main';` line. We usually go uppercase bracketed alphabetized > lowercase bracketed alphabetized, so we do group by casing generally/

johnson-mage (approved):

mage_ai/frontend/components/ServerTimeDropdown/index.tsx
anniexcheng: ah oops missed that

PR #3795, opened 2023-10-18
Bugfix: Remove lower-casing from postgres unique-constraints

Duckapple:
# Description
Since lower-casing isn't performed on the columns inserted, this fix will ensure they both aren't changed. Then, it will be up to the user of the function to completely clean any wrong columns.

_Perhaps_ it would be appropriate to entirely align the use of `_clean_column_name` across the function, but this should work.

# How Has This Been Tested?
```py
from mage_ai.io.postgres import Postgres
c = Postgres()

# Old behaviour
print(c._clean_column_name('Domain'))  # '_domain'
print(c._clean_column_name('Domain', allow_reserved_words=True))  # 'domain'

# Added behaviour
print(c._clean_column_name('Domain'), perform_lower_casing=False))  # '_Domain'
print(c._clean_column_name('Domain'), allow_reserved_words=True, perform_lower_casing=False))  # 'Domain'
```

I have also used this to insert data with preserved casing in my own Mage pipelines.

# Checklist
- [ ] ~~The PR is tagged with proper labels (bug, enhancement, feature, documentation)~~ (no access)
  - Just run the above block on the branch, should be enough

### DIFF
--- a/mage_ai/io/base.py
+++ b/mage_ai/io/base.py
@@ -286,8 +286,10 @@ class BaseSQLDatabase(BaseIO):
         """
         return query_string.strip(' \n\t')
 
-    def _clean_column_name(self, column_name: str, allow_reserved_words: bool = False) -> str:
-        col_new = re.sub(r'\W', '_', column_name.lower())
+    def _clean_column_name(self, column_name: str, allow_reserved_words: bool = False, perform_lower_casing: bool = True) -> str:
+        if perform_lower_casing:
+            column_name = column_name.lower()
+        col_new = re.sub(r'\W', '_', column_name)
         if not allow_reserved_words and col_new.upper() in SQL_RESERVED_WORDS:
             col_new = f'_{col_new}'
         return col_new
--- a/mage_ai/io/base.py
+++ b/mage_ai/io/base.py
@@ -286,7 +286,12 @@ class BaseSQLDatabase(BaseIO):
         """
         return query_string.strip(' \n\t')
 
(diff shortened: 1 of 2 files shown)


wangxiaoyou1993:
lower-casing is actually performed on the columns inserted: 
https://github.com/mage-ai/mage-ai/blob/master/mage_ai/io/sql.py#L264-L268

Duckapple:
> lower-casing is actually performed on the columns inserted: https://github.com/mage-ai/mage-ai/blob/master/mage_ai/io/sql.py#L264-L268

That is assuming one uses the parent function (`export`). I explicitly opted to use the inner function (`upload_dataframe`), as it didn't perform this cleaning.

Otherwise, I could expand the scope for this feature and make it a flag like `perform_lower_casing` on the parent function.

wangxiaoyou1993:
> > lower-casing is actually performed on the columns inserted: https://github.com/mage-ai/mage-ai/blob/master/mage_ai/io/sql.py#L264-L268
> 
> That is assuming one uses the parent function (`export`). I explicitly opted to use the inner function (`upload_dataframe`), as it didn't perform this cleaning. No underscore in front means it is public in Python, right?
> 
> Otherwise, I could expand the scope for this feature and make it a flag like `perform_lower_casing` on the parent function.

I see. Because `upload_dataframe` currently is mostly used as the inner function in the `export`, so it assumes the column names are using lowercase. Simply changing the `upload_dataframe` to not use lowercase will break the use cases of using export method.
So it would be better to add the `use_lowercase` param to both `export` and `upload_dataframe` methods.

Duckapple:
I'm not going to finish this.

PR #3813, opened 2023-10-20
Download functionality for files and pipelines

PopaRares:
# Description
Added download functionality for files and pipelines. Users are now able to right click on any file in the **Files** tab and chose the new **Download file** option, or click the **save icon** on the pipelines page to download an entire pipeline in the form of a zip file.

file_download
*Downloading a file*

pipeline_download
*Downloading a pipeline*

When downloading a pipeline, users can chose to preserve its folder structure, or download the files only.

context_menu
comparison
*zip comparison between normal and file-only pipeline download*

## How it works
Each download request involves two separate calls.
1. **Token Generation**:
   - The first call creates a JWT token that contains the file path of the requested resource.
   - In the case of a pipeline download, the token will contain a list of file paths of all related files.
   - The token is sent to the client which generates the dynamic download link used in the second call.

2. **File Retrieval**:
   - The second call takes place when the client programmatically clicks on the generated link.
   - In the case of a pipeline download, all related files get zipped together in a temporary zip file.
   - This action prompts the server to send the requested file using a [Content-Disposition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) call.
   - The temporary zip is automatically deleted after download (in the case of a pipeline download). 

# How Has This Been Tested?
Manually tested from the UI.

# Checklist

cc:
@hugabora @adelcast


wangxiaoyou1993:
cc: @johnson-mage

mage_ai/frontend/components/FileBrowser/index.tsx
```diff
@@ -90,6 +90,33 @@ function FileBrowser({
   const { data: serverStatus } = api.statuses.list();
   const repoPath = useMemo(() => serverStatus?.statuses?.[0]?.repo_path, [serverStatus]);
 
+  const [downloadFile] = useMutation(
+    (fullPath: string) => api.downloads.files.useCreate(fullPath)(),
+    {
+      onSuccess: (response: any) => onSuccess(
+        response, {
+          callback: () => {
+            const url = response.data.download.uri
```
johnson-mage, line 99: linting: missing semicolon

mage_ai/frontend/components/FileBrowser/index.tsx
```diff
@@ -474,6 +501,13 @@ function FileBrowser({
             showModalNewFile({ file: selectedFile, moveFile: true });
           },
           uuid: 'move_file',
+        }, {
+          label: () => 'Download file',
+          onClick: () => {
+            const fp = getFullPathWithoutRootFolder(selectedFile);
+            downloadFile(encodeURIComponent(fp));
```
johnson-mage, line 508: Saw a linting error for a missing dependency, so you may need to include `downloadFile` in the useMemo dependency array here: https://github.com/mage-ai/mage-ai/pull/3813/files#diff-5df45198d443d4d572868f9c02c6c2d71a000c912c5a5b254ce42033ae6c930cR568-R581

mage_ai/frontend/pages/pipelines/index.tsx
```diff
@@ -1377,6 +1432,16 @@ function PipelineListPage() {
             flex={1} justifyContent="flex-end"
             key={`chevron_icon_${idx}`}
           >
+            <Button
+              {...sharedOpenButtonProps}
+              onClick={() => {
+                downloadPipeline({pipelineUUID: uuid});
```
johnson-mage, line 1438: linting: Include space between brackets (i.e. `downloadPipeline({ pipelineUUID: uuid });`)

mage_ai/frontend/pages/pipelines/index.tsx
```diff
@@ -1204,6 +1239,26 @@ function PipelineListPage() {
...
+                pipelineUUID: selectedPipeline?.uuid,
+                filesOnly: false
+              });
+            },
+            uuid: 'download',
+          },
+          {
+            label: () => 'Download (without folder structure)',
+            onClick: () => {
+              downloadPipeline({
```
johnson-mage, line 1255: Missing dependency `downloadPipeline` in useCallback dependency array here: https://github.com/mage-ai/mage-ai/pull/3813/files#diff-22d0517fd466848fa8e87718671265bcaecd728c9a6a4407b128c3532a88c6d9R1479-R1494

mage_ai/frontend/pages/pipelines/index.tsx
```diff
@@ -288,6 +289,40 @@ function PipelineListPage() {
...
 
+  const [downloadPipeline] = useMutation(
+    ({
+      pipelineUUID,
+      filesOnly = false,
+    }: {
+      pipelineUUID: string;
+      filesOnly?: boolean;
+    }) => 
+      api.downloads.pipelines.useCreate(pipelineUUID)({'download': {'ignore_folder_structure': filesOnly}}),
```
johnson-mage, line 300: linting: spaces between brackets. May also be more readable to format this over multiple lines like this:
```
api.downloads.pipelines.useCreate(pipelineUUID)(
  { 'download': { 'ignore_folder_structure': filesOnly } },
),
```

mage_ai/frontend/pages/pipelines/index.tsx
```diff
@@ -1204,6 +1239,26 @@ function PipelineListPage() {
...
           },
+          {
+            label: () => 'Download (keep folder structure)',
+            onClick: () => {
+              downloadPipeline({
+                pipelineUUID: selectedPipeline?.uuid,
+                filesOnly: false
+              });
+            },
+            uuid: 'download',
```
johnson-mage, line 1250: Let's differentiate the uuids (e.g. `download_keep_folder_structure`).

mage_ai/frontend/pages/pipelines/index.tsx
```diff
@@ -1204,6 +1239,26 @@ function PipelineListPage() {
...
+          },
+          {
+            label: () => 'Download (without folder structure)',
+            onClick: () => {
+              downloadPipeline({
+                pipelineUUID: selectedPipeline?.uuid,
+                filesOnly: true
+              });
+            },
+            uuid: 'download',
```
johnson-mage, line 1260: Let's differentiate the uuids (e.g. `download_without_folder_structure`).

johnson-mage:
@wangxiaoyou1993 Can you take a look at the backend code please?

johnson-mage:
@PopaRares Can you please test with [user authentication enabled](https://docs.mage.ai/production/authentication/overview#require-user-authentication) and as an "editor" role? You'll need to update the `DownloadPolicy` to avoid some policy errors. For example: 
image

mage_ai/frontend/pages/pipelines/index.tsx
PopaRares: Left the curly brackets next to the parentheses; hope this works.

PopaRares:
> @PopaRares Can you please test with [user authentication enabled](https://docs.mage.ai/production/authentication/overview#require-user-authentication) and as an "editor" role? You'll need to update the `DownloadPolicy` to avoid some policy errors. For example: image

I added another clause to my DownloadPolicy which fixed the issue. I've also lowered the required role to VIEWER which I think makes more sense.

mage_ai/frontend/pages/pipelines/index.tsx
johnson-mage: yup, that works too!

mage_ai/api/resources/DownloadResource.py
```diff
@@ -0,0 +1,61 @@
...
+            for block in pipeline.blocks_by_uuid.values():
+                rel_path = block.file_path.replace(repo_path, '')
+                arcname = block.file_path.split('/')[-1] if ignore_folder_structure else rel_path
+                pipeline_zip.write(block.file_path, arcname)  # write individual blocks to pipeline
+        return zip_name
+
+    def generate_download_token(file_path):
+        now = datetime.datetime.utcnow()
+        payload = {
+            "file_path": file_path,
```
wangxiaoyou1993, line 57: nit: we prefer using single quote in the python code

mage_ai/server/api/downloads.py
```diff
@@ -62,3 +66,39 @@ async def get(self, pipeline_uuid, block_uuid, **kwargs):
...
+class ApiResourceDownloadHandler(BaseHandler):
+
+    def get(self, token):
+        try:
+            decoded_payload = jwt.decode(token, JWT_SECRET, algorithms=["HS256"])
+
+            file_path = decoded_payload['file_path']
+            file_name = file_path.split('/')[-1]
+            is_temporary = file_path.startswith(TEMPORARY_DOWNLOAD_LOCATION)
+            with open(file_path, "rb") as f:
```
wangxiaoyou1993, line 80: I feel there's a security vulnerability.

User can manually generate the token if they know the JWT_SECRET (the default value is public) and then they can download any file in the instance.

mage_ai/server/api/downloads.py
```diff
@@ -62,3 +66,39 @@ async def get(self, pipeline_uuid, block_uuid, **kwargs):
...
+
+
+class ApiResourceDownloadHandler(BaseHandler):
+
+    def get(self, token):
+        try:
+            decoded_payload = jwt.decode(token, JWT_SECRET, algorithms=["HS256"])
+
+            file_path = decoded_payload['file_path']
+            file_name = file_path.split('/')[-1]
```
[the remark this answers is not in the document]
PopaRares: refactored this part so this comment is no longer necessary

mage_ai/server/api/downloads.py
PopaRares: While I believe having a default value for something as critical as a JWT secret is a security vulnerability in and of itself, we've taken measures to address this concern. At startup, we now replace the default secret with a randomly generated string value. This step effectively eliminates the risk of token forgery. However, it's important to note that this change has consequences: in the event of an environment reset, all active user sessions will become invalid.
Since the two token types are different in both scope and lifespan, we've decided to separate their secrets. The OAuth tokens will continue to use the original JWT_SECRET, while the download tokens will be generated using a new secret, JWT_DOWNLOAD_SECRET, which would get randomly generated if a specific value is not provided at startup. We're confident about this solution as it strikes a balance between security and user experience, hope you guys like it as well.

PopaRares:
Hello there,

It's been a while since I last updated this feature, but I'm thrilled to share the significant improvements I've made:

* I've created a separate JWT secret for generating download tokens. This secret is now randomly generated at startup if no value is set, addressing concerns raised by @wangxiaoyou1993.

* In the event of a pipeline download, creating the zipfile now occurs during the second call, offering several advantages:

  * The first call is faster, reducing the time the UI is locked in place.
  * This change makes the feature easily extensible for other types of downloads, such as folders or multiple files.
  * Temporary files are now created and destroyed within the same context.
* I've enhanced the process of generating [temporary files](https://docs.python.org/3/library/tempfile.html).

* My code now includes checks for relative paths that point to files outside the project's directory.

* In the first call, I've updated the response to return the download token itself instead of a pre-built URI.

* I've extracted the frontend code responsible for initiating the second download call and placed it into a utils script for better organization and maintainability.

I am eager to hear your feedback on this. Cheers!

mage_ai/server/api/downloads.py
```diff
@@ -62,3 +68,72 @@ async def get(self, pipeline_uuid, block_uuid, **kwargs):
...
+        zip_file.seek(0)  # set cursor to start of file to prepare for content extraction
+        return zip_file
+
+    def relative_path_mapping(self, path):
+        abs_path = os.path.abspath(path)
+        common_ground = os.path.commonpath([self.abs_repo_path, abs_path])
+
+        # trying to access files outside of the project folder
+        if common_ground != self.abs_repo_path:
+            raise Exception(abs_path)
```
adelcast, line 135: Can we make this Exception more specific? https://docs.python.org/3/library/exceptions.html#bltin-exceptions.. I am thinking PermissionError makes a bit more sense

mage_ai/api/constants.py
```diff
@@ -52,3 +52,6 @@ class AuthorizeStatusType(str, Enum):
 
 META_KEY_LIMIT = '_limit'
 META_KEY_OFFSET = '_offset'
+
+DOWNLOAD_TOKEN_LIFESPAN = 60  # Lifespan of generated download token in seconds
+TEMPORARY_DOWNLOAD_LOCATION = '/tmp'  # Temporary location of generated zipfiles
```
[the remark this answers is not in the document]
PopaRares: Addressed this.

mage_ai/server/api/downloads.py
PopaRares: I think **PremissionError** makes it sound like there are certain permissions that would allow you to download files outside the scope of the project, which there are none. As such I replaced it with [**ValueError**](https://docs.python.org/3/library/exceptions.html#ValueError) as I think it makes the most sense.

wangxiaoyou1993 (approved):

PR #3815, opened 2023-10-20
[ac] Add up/down arrow keyboard shortcut for navigating b/w pipeline runs

anniexcheng:
# Description

This PR addresses the following user feedback:

>  I often want to view the "Run details" tab for multiple runs, and clicking each individual run's row is cumbersome. Would love to be able to just use the UP or DOWN arrow to iterate through them quickly to find the run that I'm after.

When you're on a page with a pipeline runs table and rows are selectable, you can click on a row to select it. Once a row is selected, you can use the keyboard up arrow (↑) to navigate to the previous run and the keyboard down arrow (↓) to go to next run in the table on the current page. 
* If the first run is selected and you click the ↑ arrow, we loop to the last run in the table on the current page.
* If the last run is selected and you click the ↓ arrow, we loop to the first run in the table on the current page.
* We automatically scroll the table such that the selected row is in view.
* We don't yet enable navigation between runs on different pages of a table via keyboard shortcuts.

# How Has This Been Tested?

- [x] Select a row in the Pipeline Runs table and see it's selected
- [x] Press the ↑ arrow and see that the previous row is selected (when the current selection is not the first row)
- [x] Press the ↑ arrow and see that the last row is selected (when the current selection is the first row)
- [x] Press the ↓ arrow and see that the next row is selected (when the current selection is not the last row)
- [x] Press the ↓ arrow and see that the first row is selected (when the current selection is the last row)
- [x] Check that keyboard navigation is disabled when an action menu is open
- [x] Check that pressing other keyboard shortcuts (e.g. left/right arrows) don't do anything
- [x] Check that you can navigate interchangeably with the up/down arrows and clicking on rows

- [x] Test all above on Pipeline Runs page

  https://github.com/mage-ai/mage-ai/assets/8130751/32d26be0-cef3-40ca-8377-c6c4eb70ccb9

- [x] Test all above on Triggers Detail page

    https://github.com/mage-ai/mage-ai/assets/8130751/725bc077-7cdb-4c8a-87c8-91c040db6dd4

- [x] Test all above on Backfills Detail page

    https://github.com/mage-ai/mage-ai/assets/8130751/34480fa7-9d52-46b9-beee-dc9b88988444

cc: @johnson-mage

### DIFF
--- a/mage_ai/frontend/components/Backfills/Detail/index.tsx
+++ b/mage_ai/frontend/components/Backfills/Detail/index.tsx
@@ -163,6 +163,7 @@ function BackfillDetail({
           pipelineRuns={pipelineRuns}
           selectedRun={selectedRun}
           setErrors={setErrors}
+          setSelectedRun={setSelectedRun}
         />
         <Spacing p={2}>
           <Paginate
--- a/mage_ai/frontend/components/PipelineDetail/Runs/Table.tsx
+++ b/mage_ai/frontend/components/PipelineDetail/Runs/Table.tsx
@@ -37,6 +37,7 @@ import {
   DELETE_CONFIRM_TOP_OFFSET_DIFF,
   DELETE_CONFIRM_TOP_OFFSET_DIFF_FIRST,
   TIMEZONE_TOOLTIP_PROPS,
+  getTableRowUuid,
 } from '@components/shared/Table/constants';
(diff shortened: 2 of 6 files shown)


johnson-mage (approved):

johnson-mage:
LGTM. The scrolling while navigating between rows is a nice touch.

PR #3816, opened 2023-10-20
[jk] Detach block from shared pipelines

johnson-mage:
# Description
- Add block "Detach" feature, which lets user duplicate a block and remove any associations with other pipelines (would require to choose a new, unique block name). Feature is visible for blocks with shared pipelines.
- Supported for python, sql, R, and data integration YAML blocks (not supported for dbt blocks).

# How Has This Been Tested?
With add-on blocks (conditional and callback) and upstrea/downstream dependencies:
detach block with PUT

Dep graph before detaching `custom_sql_block2_c2`:
image
Dep graph after detaching `custom_sql_block2_c2` (replaced with `custom_sql_block2_c2_copy`):
image

# Checklist

### DIFF
--- a/mage_ai/api/resources/BlockResource.py
+++ b/mage_ai/api/resources/BlockResource.py
@@ -650,46 +650,6 @@ class BlockResource(GenericResource):
                 replicated_block=replicated_block,
             )
 
-        async def _create_callback(resource):
-            # Replace existing block in pipeline with newly created copy of that block
-            block_uuid_to_remove = payload.get('block_uuid_to_remove')
-            if pipeline and block_uuid_to_remove is not None:
-                callback_blocks = payload.get('callback_blocks')
-                if callback_blocks:
-                    callbacks_by_uuid = pipeline.callbacks_by_uuid
-                    for callback_uuid in callback_blocks:
-                        callback_block = callbacks_by_uuid.get(callback_uuid)
-                        if callback_block:
-                            existing_upstream_block_uuids = \
-                                [b.uuid for b in callback_block.upstream_blocks]
-                            callback_block.update(
-                                dict(upstream_blocks=existing_upstream_block_uuids + [block.uuid])
-                            )
-
-                conditional_blocks = payload.get('conditional_blocks')
-                if conditional_blocks:
-                    conditionals_by_uuid = pipeline.conditionals_by_uuid
-                    for conditional_uuid in conditional_blocks:
-                        conditional_block = conditionals_by_uuid.get(conditional_uuid)
-                        if conditional_block:
-                            existing_upstream_block_uuids = \
-                                [b.uuid for b in conditional_block.upstream_blocks]
-                            conditional_block.update(
-                                dict(upstream_blocks=existing_upstream_block_uuids + [block.uuid])
-                            )
-
-                block_to_remove = Block.get_block(
-                    block_uuid_to_remove,
-                    block_uuid_to_remove,
-                    block_type,
-                    language=language,
-                    pipeline=pipeline,
-                )
-                cache.remove_pipeline(block_to_remove, pipeline.uuid)
-                block_to_remove.delete()
-
-        self.on_create_callback = _create_callback
-
         return self(block, user, **kwargs)
 
     @classmethod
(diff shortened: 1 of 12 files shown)


johnson-mage:
Empty `default.jinja` template files added in streaming template folders so error would not be thrown when detaching a data loader or data exporter block in a streaming pipeline.

mage_ai/api/resources/BlockResource.py
```diff
@@ -649,6 +649,22 @@ async def create(self, payload, user, **kwargs):
                 replicated_block=replicated_block,
             )
 
+        async def _create_callback(resource):
```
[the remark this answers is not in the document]
wangxiaoyou1993: There could be some corner cases that are not covered.
For example, if conditional blocks or callback blocks are associated with the block, do we associate them with the new block?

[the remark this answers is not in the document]
johnson-mage: @wangxiaoyou1993 Made an [update](https://github.com/mage-ai/mage-ai/pull/3816/commits/6e4390282fa06de794a29e6eb911d78167cacfa2) to also include conditonal and callback blocks when detaching a block, so we do associate them with the newly created block. Added gif and screenshots in the PR description.

mage_ai/api/resources/BlockResource.py
```diff
@@ -649,6 +649,46 @@ async def create(self, payload, user, **kwargs):
...
+            # Replace existing block in pipeline with newly created copy of that block
+            block_uuid_to_remove = payload.get('block_uuid_to_remove')
+            if pipeline and block_uuid_to_remove is not None:
+                callback_blocks = payload.get('callback_blocks')
```
wangxiaoyou1993, line 656: You don't need to pass the callback blocks and conditional blocks from the API. You can just get them from the the block of `block_uuid_to_remove` (maybe `block_uuid_to_detach` is a better param name). `block_to_detach = pipeline.get_block(block_uuid_to_detach)`

And I just checked the code, there're actually more associated objects with the block: widget, conditional blocks, callback blocks, extensions, replicated block, variables directory. These need to be associated with the new block. I would suggest adding a method in the Pipeline class to transfer the associated objects.

Here is some code of updating the directories when updating the block uuid. I think they also need to be done for this case:
https://github.com/mage-ai/mage-ai/blob/master/mage_ai/data_preparation/models/pipeline.py#L1387-L1393
https://github.com/mage-ai/mage-ai/blob/master/mage_ai/data_preparation/models/pipeline.py#L1426-L1441

wangxiaoyou1993: Actually, i think there's a better and simpler way to implement this.
You can just update this method and add an option to detach: https://github.com/mage-ai/mage-ai/blob/master/mage_ai/data_preparation/models/block/__init__.py#L2787-L2829
When detach is true, you can just update this line to create the new file instead of rename the file name: https://github.com/mage-ai/mage-ai/blob/master/mage_ai/data_preparation/models/block/__init__.py#L2820C14-L2820C14

wangxiaoyou1993: You can use Block PUT endpoint instead of POST endpoint. Just update the block with the new name and "detach: true".

johnson-mage: @wangxiaoyou1993 Thank you for the suggestions, made an update to reuse the logic from renaming a block, which does work out better since it handles the associated objects, as well as upstream and downstream dependencies.

mage_ai/data_preparation/models/block/__init__.py
```diff
@@ -2861,16 +2866,39 @@ def __update_name(self, name) -> None:
...
             os.makedirs(parent_dir, exist_ok=True)
 
-            os.rename(old_file_path, new_file_path)
+            if detach:
+                """"
+                Detaching a block creates a copy of the block file while keeping the existing block
+                file the same. Without detaching a block, the existing block file is simply renamed.
+                """
+                with open(new_file_path, 'w') as f:
+                    f.write(self.content)
```
wangxiaoyou1993, line 2875: I would suggest you getting the file content before updating the uuid. 
If the file content was not read before, `self.content` will read the file content from the new path with the new uuid.
https://github.com/mage-ai/mage-ai/blob/master/mage_ai/data_preparation/models/block/__init__.py#L360-L361
https://github.com/mage-ai/mage-ai/blob/master/mage_ai/data_preparation/models/block/__init__.py#L527-L531

wangxiaoyou1993 (approved):

mage_ai/data_preparation/models/block/__init__.py
johnson-mage: got it, ty

johnson-mage:
This PR addresses part of https://github.com/mage-ai/mage-ai/issues/2742

PR #3819, opened 2023-10-21
[Fix] Reposition file browser context menu

### LINKED ISSUE #3800: File browser context menu is cut off
**Describe the bug**
- File browser context menu is cut off when right-clicking towards bottom of window.
- Need to position file browser context menu within viewport (e.g. if the context menu is opened at the bottom of the window) so that it is not cut off.

**To Reproduce**
1. Go to `/files` endpoint or any pipeline's "Edit" page.
2. Expand folders so there is a folder/file at the very bottom of the window (shrink window height if necessary).
3. Right-click one of the last items at the bottom of the window.
4. Context menu should be cut off.

**Expected behavior**
Context menu should not be cut off (e.g. should appear above the cursor instead of below it).

**Screenshots**
file browser context menu bottom of window

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome
 - Version: 0.9.36

edmondwinston:
# Description

Moved the menu up whenever it overflows off the screen. Also moved it a bit to the right, for good measure.
Resolves #3800.

# How Has This Been Tested?

- Before
image

- After
image

# Checklist

cc: @johnson-mage

### DIFF
--- a/mage_ai/frontend/components/FileBrowser/index.tsx
+++ b/mage_ai/frontend/components/FileBrowser/index.tsx
@@ -394,7 +394,6 @@ function FileBrowser({
     const {
       x: xContainer,
       width,
-      height,
     } = ref?.current?.getBoundingClientRect() || {};
     const {
       x = 0,
@@ -517,8 +516,8 @@ function FileBrowser({
 
     let yFinal = y;
     const menuHeight = MENU_ITEM_HEIGHT * items.length;
-    if (y + menuHeight >= height) {
-      yFinal = height - menuHeight;
+    if (y + menuHeight >= window.innerHeight) {
+      yFinal = y - menuHeight;
     }
 
     return (
--- a/mage_ai/frontend/components/FileBrowser/index.tsx
+++ b/mage_ai/frontend/components/FileBrowser/index.tsx
@@ -400,7 +400,7 @@ function FileBrowser({
       x = 0,
       y = 0,
     } = coordinates || {};
-    let xFinal = x;
+    let xFinal = x + UNIT;
     if (x + MENU_WIDTH >= xContainer + width) {
       xFinal = (xContainer + width) - (MENU_WIDTH + UNIT);
     }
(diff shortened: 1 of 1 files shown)


johnson-mage:
@edmondwinston Can you try testing opening the context menu when there are many files displayed (requiring you to scroll vertically in the file browser) and when all folders are collapsed (resulting in some empty space at the bottom)? I was testing out the positioning of the context menu, and it doesn't seem consistently positioned within the viewport. Please see attached gif. Note how my `transformers` folder has many files, and after opening that folder up, right-clicking the very bottom files/folders still cuts off the context menu.
file broswer context menu position2

johnson-mage:
@edmondwinston We do something similar in the Pipelines list page, so you can look at the logic from this [commit](https://github.com/mage-ai/mage-ai/pull/3648/commits/a56889870e407a3683e6fda411ce11ae9b80ab32) to see if it might be helpful. 
pipelines list context menu

mage_ai/frontend/components/FileBrowser/index.tsx
```diff
@@ -393,12 +394,13 @@ function FileBrowser({
     const {
       x: xContainer,
       width,
+      height,
```
johnson-mage, line 397: `height` here may not consistently get you the correct offset for calculating the context menu position. You might be able to use the `y` prop from the `coordinates` state to help with your calculation.

edmondwinston:
Animation

@johnson-mage Fixed, but what do you think about the change of the `x` position like below? I feel like it'd be a minor UX improvement, but don't want to push the change without getting your opinion.

Animation

mage_ai/frontend/components/FileBrowser/index.tsx
edmondwinston: This helped, you're awesome, mate. It was a bit misleading at first, because the ref is used at 2 different places, but maybe it's documented somewhere I didn't know haha.

johnson-mage (approved):

johnson-mage:
> Fixed, but what do you think about the change of the `x` position like below? I feel like it'd be a minor UX improvement, but don't want to push the change without getting your opinion.

PR looks good now! Gonna merge it. We actually don't need to adjust the `x` position because we have another PR that will show the file browser context menu above the adjacent container, making repositioning of it on the x-axis unnecessary.

PR #3820, opened 2023-10-22
chore(docs): scripts setup guide

edmondwinston:
# Description

I recently contributed to this repo, and saw chances of improvement in the doc.

cc: @mattppal

### DIFF
--- a/README_dev.md
+++ b/README_dev.md
@@ -2,16 +2,20 @@
 
 These are the instructions for setting up the development dependencies and conforming to the style guide for the Mage.ai repo.
 
+> [!WARNING]
+> _All commands below, without any notes, assume you are at the root of the repo._
+
 ## Using helper scripts
+
 Initialise a `mage` repo so you have a starting point.
+
 ```bash
-# Assuming you are at the root.
 ./scripts/init.sh default_repo
 ```
 
 Generate a Docker image, start the dev server for the backend at `localhost:6789` and frontend at `localhost:3000`.
+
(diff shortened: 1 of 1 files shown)


mattppal:
Thanks @edmondwinston this looks great, @wangxiaoyou1993 will take a look as well :)

README_dev.md
```diff
@@ -35,51 +63,103 @@ which python
...
 ```
 
-This command just uses Python and the `toml` library to output the dev dependencies from the `pyproject.toml` as a space-delimited list, and passes that output to the `pip install` command.
-## Git Hooks
+<sup>The above command uses the `toml` library to output the dev dependencies from the `pyproject.toml` as a space-delimited list, and passes that output to the `pip install` command.</sup>
+
+Then, start the dev server:
+
+```bash
+python mage_ai/server/server.py
```
wangxiaoyou1993, line 112: This command doesn't run Mage in docker.
If you just want to run the server only, you can use `./scripts/start.sh default_repo`

mattppal: Updated

wangxiaoyou1993 (approved):

README_dev.md
edmondwinston: @mattppal @wangxiaoyou1993 

I think there's a misunderstanding here. The note to start the server only with the script should be placed at Docker section, the section you are quoting is to setup a local server, hence the Python command. I'll make the changes.

PR #3828, opened 2023-10-24
[mp] Add Google Sheets IO Class

mattppal:
# Description

This PR adds a Google Sheets IO Class, using the [gspread](https://docs.gspread.org/en/latest/index.html) library for basic operations with a google sheet.

It supports:
- Loading single-worksheet data from an existing Google Sheet by:
  - Sheet URL
  - Sheet ID
  - Sheet name
- Checking if a worksheet exists
- Exporting data to an existing worksheet via:
  - Worksheet name
  - Worksheet position
- Arbitrary header rows may be specified with the `header_rows` argument.

It requires the following to be enabled via GCP: 
1. Google Drive API 
2. Google Sheets API

And the following:
1. Google Sheet must be shared with the service account used in Mage

# How Has This Been Tested?

## Load sheet

image

### True
image

### False
image

## Export sheet
image

Exported data:
image

Locally, using an existing sheet.

- [ ] Test A
- [ ] Test B

# Checklist

cc:


dy46 (commented):
general comment: we use single quotes for strings in python. if you're using black you can add the `--skip-string-normalization` option to your black settings

mage_ai/io/google_sheets.py
```diff
@@ -0,0 +1,287 @@
...
+        if worksheet_name is not None:
+            try:
+                worksheet = sheet.worksheet(worksheet_name)
+            except Exception as name_ex:
+                raise name_ex
+        else:
+            try:
+                worksheet = sheet.get_worksheet(worksheet_position)
+            except Exception as postition_ex:
+                raise postition_ex
```
dy46, line 160: why are we raising the exception like this? you can just remove the try/except for these if you want it to raise an exception

mage_ai/io/google_sheets.py
```diff
@@ -0,0 +1,287 @@
...
+                sheet_url=sheet_url,
+                sheet_id=sheet_id,
+                sheet_name=sheet_name,
+                worksheet_name=worksheet_name,
+                worksheet_position=worksheet_position,
+            )
+            return True
+
+        except Exception as ex:
+            print(ex)
```
dy46, line 261: if we print the exception here, the exception will show up in the output for the block when running it from the notebook. not sure how useful it is to log the exception here

requirements.txt
```diff
@@ -59,6 +59,7 @@ google-cloud-iam==2.9.0
 google-cloud-pubsub==2.16.0
 google-cloud-run==0.5.0
 google-cloud-storage==2.5.0
+gspread==5.7.2
```
dy46, line 62: you'll need to add this to `setup.py` as well

mage_ai/data_preparation/templates/constants.py
```diff
@@ -531,6 +544,13 @@
         name='Google BigQuery',
         path='data_exporters/bigquery.py',
     ),
+    dict(
```
wangxiaoyou1993, line 547: This seems a duplicate item

mage_ai/io/google_sheets.py
```diff
@@ -0,0 +1,287 @@
...
+            raise ValueError("There was an error loading the worksheet.")
+
+    def load(
+        self,
+        sheet_url: str = None,
+        sheet_id: str = None,
+        sheet_name: str = None,
+        worksheet_name: str = None,
+        worksheet_position: int = 0,
+        header_rows: int = 1,
```
wangxiaoyou1993, line 177: You mentioned the default value is 0 in the comment but the default value is 1 here

mage_ai/io/google_sheets.py
mattppal: Uh yeah admittedly this is weird. I'm used to just try/except, but flake8 threw an error so I basically hacked a solution that didn't fail linting. Do you have an example of how we should be raising exceptions?

mage_ai/io/google_sheets.py
mattppal: Oh. Yeah it looks like a json payload and isn't great. Is there a better way to log?

mage_ai/io/google_sheets.py
mattppal: Default should be 1. I made an error elsewhere. Will fix

mage_ai/io/google_sheets.py
dy46: you don't have to catch the exception here. right now, you're catching the exception and then immediately raising it again which is the same behavior as if the try/except block wasn't there

mage_ai/io/google_sheets.py
dy46: is there another way to check if the worksheet exists? it's a bit awkward to just try to fetch it and catch an exception since the worksheet could fail to be fetched for many reasons

mattppal: Yes sir!

https://stackoverflow.com/a/73385054

mattppal:
@dy46 any ideas why I'm getting this blank line:

image

mage_ai/io/google_sheets.py
mattppal: image

Ended up being a slightly different solution— looks like an APIError 404 is thrown instead

mage_ai/io/google_sheets.py
```diff
@@ -0,0 +1,289 @@
...
+                        mapping_obj, scopes=scopes
+                    )
+            if 'path_to_credentials' in kwargs:
+                path = kwargs.pop('path_to_credentials')
+                if path is not None:
+                    credentials = service_account.Credentials.from_service_account_file(
+                        path, scopes=scopes
+                    )
+            if 'credentials' in kwargs:
+                kwargs.pop('credentials')
```
dy46, line 79: I think this should be outside of the `if credentials is None:` block

mage_ai/io/google_sheets.py
```diff
@@ -0,0 +1,289 @@
...
+        else:
+            worksheet = sheet.get_worksheet(worksheet_position)
+
+        if worksheet is not None:
+            with self.printer.print_msg(
+                f'Loaded worksheet \'{worksheet.title}\' at index \'{worksheet.index}\''
+            ):
+                return worksheet
+        else:
+            raise ValueError('There was an error loading the worksheet.')
```
dy46, line 167: I think you can just `return None` here. If the worksheet fails to load for some reason, I'm guessing the `gspread` code will throw an error.

setup.py
```diff
@@ -72,7 +72,10 @@ def readme():
             'dbt-trino==1.4.0',
             'dbt-clickhouse==1.4.0',
         ],
-        'google-cloud-storage': ['google-cloud-storage==2.5.0'],
+        'google-cloud-storage': [
+            'google-cloud-storage==2.5.0',
+            'gspread==5.7.2',
```
dy46, line 77: you'll want to add this to the `'all'` section as well

mage_ai/io/google_sheets.py
```diff
@@ -0,0 +1,289 @@
...
+            raise KeyError(
+                'Please set one of sheet_url, sheet_id, or sheet_name to a valid Google Sheet.'
+            )
+        if sheet is not None:
+            with self.printer.print_msg(
+                f'Loaded Google Sheet \'{sheet.title}\' with ID \'{sheet.id}\''
+            ):
+                return sheet
+        else:
+            raise ValueError('There was an error loading the Google Sheet.')
```
dy46, line 120: same here, you can just return None, and check for None when calling this method. As a developer, I think I'd rather know that it returned None than just a vague error message.

mage_ai/io/google_sheets.py
mattppal: I took this directly from the BigQuery code [here](https://github.com/mage-ai/mage-ai/blob/f07ba491bd048eceadc1010ed97dd9841f25220b/mage_ai/io/bigquery.py#L52), so I _think_ this is correct.

dy46: ok i guess i could see how that would make sense too

mage_ai/io/google_sheets.py
```diff
@@ -0,0 +1,297 @@
...
+
+        Returns:
+            gspread.worksheet.Worksheet: A worksheet object.
+        """
+        sheet = self.fetch_sheet(
+            sheet_url=sheet_url,
+            sheet_id=sheet_id,
+            sheet_name=sheet_name,
+        )
+
```
dy46, line 148: can you add a check for `sheet is None` here?

dy46 (approved):

PR #3829, opened 2023-10-24
[ac] Make current time button transparent by default

anniexcheng:
# Description
Previously, we would show the current time button in the top right of the app header with a dark background by default (#3785):

image

This PR makes a slight style adjustment such that the button has a transparent background by default. When you hover over the button OR if the dropdown menu is open, the button displays a dark background. This was a design request by Kai!

# How Has This Been Tested?

- [x] Check that current time button has a transparent background by default (when the dropdown is closed)
- [x] Check that the current time button has a dark background on hover
- [x] Check that the current time button has a dark background when the dropdown is open

    https://github.com/mage-ai/mage-ai/assets/8130751/2bee93b3-da72-49e4-9152-8625264af4b4

cc:
@johnson-mage

### DIFF
--- a/mage_ai/frontend/components/ServerTimeDropdown/index.style.tsx
+++ b/mage_ai/frontend/components/ServerTimeDropdown/index.style.tsx
@@ -1,11 +1,24 @@
 import styled from 'styled-components';
 
+import Button from '@oracle/elements/Button';
 import dark from '@oracle/styles/themes/dark';
 import { BORDER_RADIUS } from '@oracle/styles/units/borders';
 import { UNIT } from '@oracle/styles/units/spacing';
 
 const SHARED_CELL_PADDING = '12px 20px';
 const SHARED_GAP = '1px';
+
+export const ButtonStyle = styled(Button)<{ active?: boolean }>`
+  background-color: transparent;
+
+  ${props => (props.active) && `
+    background-color: ${(props.theme.background || dark.background).dashboard};
+  `};
+
+  &:hover {
+    background-color: ${props => (props.theme.background || dark.background).dashboard};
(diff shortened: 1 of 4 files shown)


mage_ai/frontend/components/ServerTimeDropdown/index.style.tsx
```diff
@@ -1,12 +1,25 @@
...
+import Button from '@oracle/elements/Button';
 import dark from '@oracle/styles/themes/dark';
 import { BORDER_RADIUS } from '@oracle/styles/units/borders';
 import { UNIT } from '@oracle/styles/units/spacing';
 
 const SHARED_CELL_PADDING = '12px 20px';
 const SHARED_GAP = '1px';
 
+export const ButtonStyle = styled(Button)<{ active?: boolean }>`
+  background-color: transparent;
```
johnson-mage, line 12: There's actually a `transparent` prop on the `Button` element you can use.

mage_ai/frontend/components/ServerTimeDropdown/index.style.tsx
```diff
@@ -1,12 +1,25 @@
...
+
+  ${props => (props.active) && `
+    background-color: ${(props.theme.background || dark.background).dashboard};
+  `};
+
+  &:hover {
```
johnson-mage, line 18: Can we add a `highlightOnHoverAlt` prop to the `Button` element so we can potentially use it for other places? We already have a `highlightOnHover` prop, but it highlights buttons with a very light background, so this would be another option for highlighting a button dark.

johnson-mage (approved):

PR #3836, opened 2023-10-24
[dy] Lifecycle post start script

dy46:
# Description

This PR adds support for post start scripts. Kubernetes containers already have a [`postStart` lifecycle hook](https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/), so we use that to support post start scripts. The config for post start looks like this:

```
post_start:
    command: str
    hook_path: str
```

If the user just provides the `command` argument, we will just add that to the postStart lifecycle hook. If a user also provides `hook_path`, we will mount the file at `hook_path` to the container, so that it can be referenced in the `command`.

This PR also updates the frontend so that users can configure the pre start and post start lifecycle fields.

# How Has This Been Tested?

- [x] Tested with post start command
- [x] Tested with both pre start and post start command 
- [x] Tested with post start command that uses the post start hook
Screenshot 2023-10-24 at 4 05 23 PM

# Checklist

cc: @wangxiaoyou1993


mage_ai/cluster_manager/config.py
```diff
@@ -65,4 +84,16 @@ def parse_config(self, config: Dict = None) -> Dict:
                 config=termination_policy
             )
 
+        post_start = config.get('post_start')
```
wangxiaoyou1993, line 87: You don't need to manually parse the subclass. This is already done in the base class `load` method: https://github.com/mage-ai/mage-ai/blob/master/mage_ai/shared/config.py#L41-L46

mage_ai/cluster_manager/kubernetes/workload_manager.py
```diff
@@ -67,7 +72,7 @@ def load_config(cls) -> bool:
             pass
 
         try:
-            config.load_kube_config()
+            config.load_kube_config('/home/src/testfiles/kubeconfig')
```
wangxiaoyou1993, line 75: what's this path?

mage_ai/cluster_manager/kubernetes/workload_manager.py
```diff
@@ -144,8 +149,24 @@ def create_workload(
         name: str,
         workspace_config: KubernetesWorkspaceConfig,
         project_type: str = ProjectType.STANDALONE,
-        **kwargs,
```
wangxiaoyou1993, line 147: why do you remove the `**kwargs`?

mage_ai/cluster_manager/kubernetes/workload_manager.py
```diff
@@ -165,9 +186,21 @@ def create_workload(
 
         ingress_name = workspace_config.ingress_name
 
+        self.__create_persistent_volume(
+            name,
+            volume_host_path='/Users/david_yang/mage/mage-ai/testfiles',
```
wangxiaoyou1993, line 191: Is this for testing?

mage_ai/cluster_manager/kubernetes/workload_manager.py
```diff
@@ -434,6 +493,12 @@ def delete_workload(self, name: str):
             # The delete operation will return a 404 response if the config map does not exist
             if ex.status != 404:
                 raise
+        try:
+            self.core_client.delete_namespaced_config_map(f'{name}-hooks', self.namespace)
```
wangxiaoyou1993, line 497: duplicate code

mage_ai/cluster_manager/kubernetes/workload_manager.py
```diff
@@ -478,30 +543,44 @@ def restart_workload(self, name: str) -> None:
...
         name: str,
-        pre_start_script_path: str,
-        mage_container_config: Dict,
-    ) -> None:
-        self.__validate_pre_start_script(pre_start_script_path, mage_container_config)
+        pre_start_script_path: str = None,
+        mage_container_config: Dict = None,
+        post_start_config: PostStart = None,
+    ) -> Dict:
+        config_map_data = {}
```
wangxiaoyou1993, line 553: You want to return or raise Exception if the `mage_container_config` is None, right?

mage_ai/cluster_manager/kubernetes/workload_manager.py
```diff
@@ -294,6 +329,30 @@ def create_workload(
...
+            volumes.append(
+                {
+                    'name': 'lifecycle-hooks',
+                    'configMap': {
+                        'name': f'{name}-hooks',
+                        'items': [
+                            {
+                                'key': key,
+                                'path': key,
+                                'mode': 0o0755,
```
wangxiaoyou1993, line 348: python script doesn't need to use 0755 mode
bash script needs it

mage_ai/cluster_manager/kubernetes/workload_manager.py
dy46: yes i will remove this

mage_ai/cluster_manager/kubernetes/workload_manager.py
dy46: i removed the `kwargs` because i'm expecting all the parameters to be passed in through the `workspace_config` argument now

wangxiaoyou1993: Is this code still being used? https://github.com/mage-ai/mage-ai/blob/master/mage_ai/server/api/clusters.py#L129-L134

dy46: good call i don't think this is used anymore. i'll remove it

dy46:

wangxiaoyou1993 (approved):

PR #3843, opened 2023-10-25
[ac] Add `ctrl/cmd + click` keyboard shortcut for selecting pipeline run rows

anniexcheng:
# Description

In the Pipeline Runs table, you can click on a row to **highlight** it and open the sidekick with run details. You can also click on a row's checkbox (on the far left) to **select/deselect** it for performing actions on one or more rows. Previously, clicking on a row's checkbox both selected/deselected and highlighted the row.

This PR makes a few updates to how we highlight/select rows:
* When you click on a row's checkbox, we no longer highlight the row and only select/deselect it. We've also made a broader change to prevent a table row from being highlighted if the user clicked on an interactive element (e.g. `a`, `button`, `input`) in the row.
* We also add a new keyboard shortcut for selecting rows. If you hold the meta key (`Cmd` on Mac or `Ctrl` on Windows) and click on an unselected row, that row will be selected without being highlighted. If the clicked row is already selected, that row will be unselected. ([Airtable task](https://airtable.com/applEuqJsK4zF5yJK/tbl5bIlFkk2KhCpL4/viwCmaGgcnwD9IqOM/rectr4unRQGOn8xjz?blocks=hide))
* The new keyboard shortcut only works when rows are selectable in the Pipeline Runs table. If rows are not selectable (e.g. on the Backfills page), the keyboard shortcut does not take effect. For example, if you hold the meta key + click a row in the Pipeline Runs table on the Backfills page, the row will be highlighted.

# How Has This Been Tested?

- [x] Tested that clicking on a row's checkbox selects/deselects it without highlighting it
- [x] Tested that clicking on a row highlights it (but does not select/unselect it)
- [x] Tested that clicking on an unselected row while holding the meta key down selects it without highlighting it
- [x] Tested that clicking on a selected row while holding the meta key down deselects it without highlighting it
- [x] Tested that clicking on an interactive element in the row doesn't highlight the row

https://github.com/mage-ai/mage-ai/assets/8130751/436fca61-796c-4959-ae5b-cfc3385b483c

cc: @johnson-mage

### DIFF
--- a/mage_ai/frontend/components/shared/Table/index.tsx
+++ b/mage_ai/frontend/components/shared/Table/index.tsx
@@ -1,6 +1,5 @@
 import NextLink from 'next/link';
 import React, {
-  createRef,
   useCallback,
   useEffect,
   useMemo,
@@ -36,8 +35,9 @@ import {
   TableWrapperStyle,
 } from './index.style';
 import { goToWithQuery } from '@utils/routing';
-import { pushAtIndex, sortByKey } from '@utils/array';
+import { isInteractiveElement } from '@context/shared/utils';
 import { set } from '@storage/localStorage';
+import { sortByKey } from '@utils/array';
 
 export type ColumnType = {
   center?: boolean;
@@ -428,11 +428,13 @@ function Table({
           noHover={!(linkProps || onClickRow || renderExpandedRowWithObject)}
           // @ts-ignore
           onClick={(e) => {
-            if (onClickRow) {
-              handleRowClick(rowIndex, e);
+            if (!isInteractiveElement(e)) {
+              if (onClickRow) {
+                handleRowClick(rowIndex, e);
+              }
+  
+              onClickRowInternal(rowIndex, e);
             }
-
-            onClickRowInternal(rowIndex, e);
           }}
           onContextMenu={hasRightClickMenu
             ? (e) => {
--- a/mage_ai/frontend/components/shared/Table/index.tsx
+++ b/mage_ai/frontend/components/shared/Table/index.tsx
@@ -88,7 +88,7 @@ type TableProps = {
   menu?: any;
   noBorder?: boolean;
   noHeader?: boolean;
-  onClickRow?: (index: number) => void;
+  onClickRow?: (index: number, event?: any) => void;
   onDoubleClickRow?: (index: number) => void;
   onRightClickRow?: (index: number, event?: any) => void;
   renderExpandedRowWithObject?: (index: number, object: any) => any;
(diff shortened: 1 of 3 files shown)


mage_ai/frontend/components/shared/Table/index.tsx
```diff
@@ -428,11 +428,13 @@ function Table({
           noHover={!(linkProps || onClickRow || renderExpandedRowWithObject)}
           // @ts-ignore
           onClick={(e) => {
-            if (onClickRow) {
-              handleRowClick(rowIndex, e);
-            }
+            if (!isInteractiveElement(e)) {
```
anniexcheng, line 431: This is a broader change to the shared table to not trigger the `onClickRow` functionality if the clicked element is an interactive element (e.g. a button in the row). It should capture most (but not all) cases of interactive elements. 

We also don't have to do this change and can individually stop event propagation in the individual interactive element's event handlers as an alternative. Let me know what you think works best!

anniexcheng:
I ended up going with the `meta key + click` rather than `shift key + click` shortcut for the following reasons:

* The former feels more intuitive for selecting individual rows and follows the same keyboard selection behavior as in file browsers. Shift + click is typically used for selecting multiple consecutive rows in one click.
* Shift + click also triggers the default text selection behavior for browsers, which we could handle if we also wanted to add that shortcut and override the default selection behavior.
  image

mage_ai/frontend/components/shared/Table/index.tsx
johnson-mage: Including this sounds great. In the use case of our Pipeline Runs table, does adding this mean that the row should not be selected (and highlighted with the black background) when clicking the Retry button under the `Status` column even if we remove the `e.stopPropagation()` line of code? Or do we also have to pass in the `event` for the row cells or something?

johnson-mage (approved):

johnson-mage:
> I ended up going with the `meta key + click` rather than `shift key + click` shortcut for the following reasons:
> 
> * The former feels more intuitive for selecting individual rows and follows the same keyboard selection behavior as in file browsers. Shift + click is typically used for selecting multiple consecutive rows in one click.
> * Shift + click also triggers the default text selection behavior for browsers, which we could handle if we also wanted to add that shortcut and override the default selection behavior.
>   image

I agree with this decision.

mage_ai/frontend/components/shared/Table/index.tsx
anniexcheng: Oh this change shouldn't affect the current behavior of clicking the Retry button! The main idea was if you click on any interactive element in the row, it wouldn't also highlight the row. Clicking on the button under the Status column (e.g. Ready, Done) pops up the retry menu and doesn't highlight the row. However, clicking on a button within the retry menu should still have the same behavior! Only the top-level element within the row matters. Let me know if I can clarify more!

PR #3855, opened 2023-10-26
[ac] Make sure file browser context menu is always fully visible in the viewport when open

anniexcheng:
# Description

Previously, the file browser context menu would be partially hidden in scenarios where the file browser area was too small to fit the full menu. This was because overflow was hidden in the file browser aside:

https://github.com/mage-ai/mage-ai/assets/8130751/91073d14-39f7-4688-ad97-5e4ca81990c5

This PR fixes this bug and makes sure that the file browser context menu is always fully visible within the viewport ([Airtable Task Pt. 1](https://airtable.com/applEuqJsK4zF5yJK/tbl5bIlFkk2KhCpL4/viwCmaGgcnwD9IqOM/recrnz90jwBp3lxJJ?blocks=hide)). The fix makes use of [React Portals](https://react.dev/reference/react-dom/createPortal) to render the context menu into a DOM node that exists outside the DOM hierarchy of the parent component, which has `overflow: hidden`.

Note: This is a one-off fix that works because the menu uses fixed positioning; the longer term solution would be a much bigger refactor to `FlyoutMenu` (potentially using React Portals while supporting absolute positioning).

# How Has This Been Tested?

- [x] Tested that the file browser context menu is fully visible when aside has a narrow width
- [x] Tested that the file browser context menu is fully visible when clicking a file near the bottom of the viewport
- [x] Tested that the above behavior works for the file, block file, and folder context menus

https://github.com/mage-ai/mage-ai/assets/8130751/7f934ad8-2b3e-4b30-951d-d94ec3cfbcb7

cc: @johnson-mage

### DIFF
--- a/mage_ai/frontend/components/FileBrowser/index.tsx
+++ b/mage_ai/frontend/components/FileBrowser/index.tsx
@@ -515,7 +515,7 @@ function FileBrowser({
       }
     }
 
-    let yFinal = y;
+    let yFinal = y + (UNIT / 2);
     const menuHeight = MENU_ITEM_HEIGHT * items.length;
     if (y + menuHeight >= window.innerHeight) {
       yFinal = y - menuHeight;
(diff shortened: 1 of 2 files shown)


johnson-mage (approved):
approved after fixing merge conflict

anniexcheng:
@johnson-mage fixed the merge conflict! also uploaded a new video to show the new behavior

PR #3867, opened 2023-10-30
[jk] Add "No tags" filtering option for pipelines

edmondwinston:
# Description
This PR adds a `No tags` filtering option on the `/pipelines` page.

# How Has This Been Tested?
image

filter pipelines with no tags

# Checklist

cc: @johnson-mage

### DIFF
--- a/mage_ai/frontend/pages/pipelines/index.tsx
+++ b/mage_ai/frontend/pages/pipelines/index.tsx
@@ -657,15 +657,7 @@ function PipelineListPage() {
   ]);
 
   const { data: dataTags } = api.tags.list();
-  const tags: TagType[] = useMemo(() => {
-    const t = dataTags?.tags
-    if (t?.length > 0 && t[0].uuid !== 'no_tag') { // <- So as to not to re-add the `no_tag` tag.
-      t.push({
-        uuid: 'no_tag',
-      })
-    }
-    return sortByKey(t || [], ({ uuid }) => uuid)
-  }, [
+  const tags: TagType[] = useMemo(() => sortByKey(dataTags?.tags || [], ({ uuid }) => uuid), [
     dataTags,
   ]);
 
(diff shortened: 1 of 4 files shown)


johnson-mage:
@edmondwinston You can install the `ESLint` extension in VSCode to catch minor linting issues like missing semicolons.

mage_ai/frontend/pages/pipelines/index.tsx
```diff
@@ -661,7 +657,15 @@ function PipelineListPage() {
   ]);
 
   const { data: dataTags } = api.tags.list();
-  const tags: TagType[] = useMemo(() => sortByKey(dataTags?.tags || [], ({ uuid }) => uuid), [
+  const tags: TagType[] = useMemo(() => {
+    const t = dataTags?.tags
+    if (t?.length > 0 && t[0].uuid !== 'no_tag') { // <- So as to not to re-add the `no_tag` tag.
```
johnson-mage, line 662: This seems to cause some unintended duplicates for the `no_tag` uuid.
image

johnson-mage:
When including a `no_tag` filter plus some other tags, we would expect the pipelines without tags to be included, as well as the pipelines with the selected tags. It looks like if you include the `no_tag` filter plus some other tag, no pipelines will be displayed in the UI.
image

mage_ai/frontend/pages/pipelines/index.tsx
edmondwinston: Ah, I was short-sighted - I didn't test the feature in every ways. I also saw you pushing, so it'd be lovely if I could take another issue to make it up.

johnson-mage:
@edmondwinston Sorry to take over the PR. There were several changes that I thought should be made, but it would've taken a lot longer to go through every one, so I thought it'd be easier if you saw the diff after I committed the changes to get an idea of what I was thinking.

edmondwinston:
No worries boss, I truly learnt a lot.

dy46 (approved):

PR #3868, opened 2023-10-30
[dy] Add pipeline run limit for a pipeline

dy46:
# Description

Add new config options to `ConcurrencyConfig`:
* `pipeline_run_limit_all_triggers`: set the maximum number of pipeline runs for a single pipeline
* `on_pipeline_run_limit_reached`: set behavior for pipeline runs when the pipeline run limit is reached. if set to "skip", the scheduler will cancel the current run when the limit is reached.

I had to update the scheduling logic in `schedule_all` to handle the new configuration option to loop through schedules grouped by pipeline.

# How Has This Been Tested?

- [x] Tested locally
- [x] Unit tests

# Checklist

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/orchestration/pipeline_scheduler.py
+++ b/mage_ai/orchestration/pipeline_scheduler.py
@@ -1539,7 +1539,7 @@ def schedule_all():
             pipeline_quota = None
 
         quota_filtered_runs = pipeline_runs_to_start
-        if pipeline_quota is not None:
+        if pipeline_quota is not None and pipeline_quota >= 0:
             pipeline_runs_to_start.sort(key=lambda x: x.execution_date)
             quota_filtered_runs = pipeline_runs_to_start[:pipeline_quota]
             pipeline_runs_excluded_by_limit.extend(
@@ -1547,8 +1547,6 @@ def schedule_all():
             )
 
         for r in quota_filtered_runs:
-            if r.block_runs_count == 0 and r.pipeline_uuid is not None:
-                r.create_block_runs()
             PipelineScheduler(r).start()
 
         # If on_pipeline_run_limit_reached is set as SKIP, cancel the pipeline runs that
--- a/mage_ai/orchestration/pipeline_scheduler.py
+++ b/mage_ai/orchestration/pipeline_scheduler.py
@@ -1539,7 +1539,8 @@ def schedule_all():
             pipeline_quota = None
 
         quota_filtered_runs = pipeline_runs_to_start
-        if pipeline_quota is not None and pipeline_quota >= 0:
+        if pipeline_quota is not None:
+            pipeline_quota = pipeline_quota if pipeline_quota > 0 else 0
             pipeline_runs_to_start.sort(key=lambda x: x.execution_date)
             quota_filtered_runs = pipeline_runs_to_start[:pipeline_quota]
             pipeline_runs_excluded_by_limit.extend(
--- a/mage_ai/orchestration/pipeline_scheduler.py
+++ b/mage_ai/orchestration/pipeline_scheduler.py
@@ -2,6 +2,7 @@ import asyncio
 import os
 import traceback
 from datetime import datetime, timedelta
+from itertools import groupby
 from typing import Any, Dict, List, Set, Tuple
 
 import pytz
@@ -30,7 +31,7 @@ from mage_ai.data_preparation.models.triggers import (
 )
 from mage_ai.data_preparation.repo_manager import get_repo_config
 from mage_ai.data_preparation.sync.git_sync import get_sync_config
-from mage_ai.orchestration.concurrency import ConcurrencyConfig
+from mage_ai.orchestration.concurrency import ConcurrencyConfig, OnLimitReached
 from mage_ai.orchestration.db import db_connection, safe_db_query
 from mage_ai.orchestration.db.models.schedules import (
     Backfill,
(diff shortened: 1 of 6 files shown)


mage_ai/server/api/triggers.py
```diff
@@ -44,10 +48,27 @@ def post(self, pipeline_schedule_id, token: str = None):
                 payload['event_variables'][k] = v
 
         pipeline = Pipeline.get(pipeline_schedule.pipeline_uuid)
-        pipeline_run = create_and_start_pipeline_run(
-            pipeline,
-            pipeline_schedule,
-            payload,
-        )
+
+        pipeline_runs = PipelineRun.active_runs_for_pipelines([pipeline.uuid])
```
wangxiaoyou1993, line 52: We can simplify the logic here. The API trigger only needs to create the pipeline run and doesn't need to start the pipeline run like https://github.com/mage-ai/mage-ai/blob/master/mage_ai/api/resources/PipelineRunResource.py#L262

The pipeline run start or cancellation logic can be centralized in the `schedule_all` method

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -1520,23 +1598,59 @@ def schedule_with_event(event: Dict = None):
...
-        for p in matched_pipeline_schedules:
-            payload = dict(
-                execution_date=datetime.now(tz=pytz.UTC),
-                pipeline_schedule_id=p.id,
-                pipeline_uuid=p.pipeline_uuid,
-                variables=merge_dict(p.variables or dict(), dict(event=event)),
-            )
-            pipeline_run = PipelineRun.create(**payload)
-            pipeline_scheduler = PipelineScheduler(pipeline_run)
+        matched_pipeline_uuids = list(set([s.pipeline_uuid for s in matched_pipeline_schedules]))
```
wangxiaoyou1993, line 1601: Same here for event trigger. This method just needs to create the pipeline run.
The scheduling or cancellation logic can be put in `schedule_all` method.

mage_ai/server/api/triggers.py
dy46: ah that's smart. i'll change it

dy46:

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -1397,106 +1411,156 @@ def schedule_all():
...
+            pipeline_runs_to_start.sort(key=lambda x: x.execution_date)
+            quota_filtered_runs = pipeline_runs_to_start[:pipeline_quota]
+            pipeline_runs_excluded_by_limit.extend(
+                pipeline_runs_to_start[pipeline_quota:]
+            )
 
-        lock.release_lock(lock_key)
+        for r in quota_filtered_runs:
+            if r.block_runs_count == 0 and r.pipeline_uuid is not None:
+                r.create_block_runs()
```
wangxiaoyou1993, line 1551: You don't need to create block runs here. The `start` method does the job to create block runs: https://github.com/mage-ai/mage-ai/blob/master/mage_ai/orchestration/pipeline_scheduler.py#L137-L149

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -1397,106 +1411,156 @@ def schedule_all():
...
+        pipeline_run_limit = concurrency_config.pipeline_run_limit_all_triggers
+        if pipeline_run_limit is not None:
+            pipeline_quota = pipeline_run_limit - len(
+                pipeline_runs_by_pipeline.get(pipeline_uuid, [])
+            )
+        else:
+            pipeline_quota = None
+
+        quota_filtered_runs = pipeline_runs_to_start
+        if pipeline_quota is not None:
```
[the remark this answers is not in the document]
dy46: i set to check for >= 0 since this logic works if it's 0 as well.

mage_ai/orchestration/pipeline_scheduler.py
```diff
@@ -1397,106 +1411,154 @@ def schedule_all():
...
+        pipeline_run_limit = concurrency_config.pipeline_run_limit_all_triggers
+        if pipeline_run_limit is not None:
+            pipeline_quota = pipeline_run_limit - len(
+                pipeline_runs_by_pipeline.get(pipeline_uuid, [])
+            )
+        else:
+            pipeline_quota = None
+
+        quota_filtered_runs = pipeline_runs_to_start
+        if pipeline_quota is not None and pipeline_quota >= 0:
```
wangxiaoyou1993, line 1542: actually, i think we should set pipeline_quota to 0 if it's a negative number.
otherwise, if it's a negative number, we start ll the pipeline runs, which is not right.

dy46: good point, let me also add a unit test for that

mage_ai/tests/orchestration/test_pipeline_scheduler.py
```diff
@@ -142,38 +152,51 @@ def new_schedule(self):
...
-            PipelineRun.
-            query.
-            filter(
-                PipelineRun.pipeline_schedule_id == pipeline_schedule.id,
-                PipelineRun.status == PipelineRun.PipelineRunStatus.RUNNING,
-            ).
-            all()
-        )), 0)
+        self.assertEqual(
+            len(
```
wangxiaoyou1993, line 166: nit: this can be simplified as `PipelineRun.query.filter(...).count()`

wangxiaoyou1993 (approved):

PR #3870, opened 2023-10-30
[jk] Revert pipeline type when error occurs switching type

johnson-mage:
# Description
- When users switch the `python` kernel to `pyspark` and an error occurs, the pipeline `type` should not update from `python` to `pyspark`, but should revert back to `python`. This PR reverts the pipeline type when trying to update the kernel and an error occurs.

# How Has This Been Tested?
Before (pipeline type is updated):
BEFORE - revert pipeline type

After (pipeline type is reverted back to its original type):
AFTER - revert pipeline type

# Checklist

### DIFF
--- a/mage_ai/api/resources/PipelineResource.py
+++ b/mage_ai/api/resources/PipelineResource.py
@@ -415,7 +415,7 @@ class PipelineResource(BaseResource):
             if pipeline_type_updated is not None:
                 await self.model.update(dict(type=pipeline_type))
 
-            raise Exception(e)
+            raise e
 
         @safe_db_query
         def update_schedule_status(status, pipeline_uuid):
--- a/mage_ai/api/resources/PipelineResource.py
+++ b/mage_ai/api/resources/PipelineResource.py
@@ -403,11 +403,19 @@ class PipelineResource(BaseResource):
             if pipeline_doc:
                 await _add_markdown_block(pipeline_doc, self.model.uuid, 0)
 
+        pipeline_type = self.model.type
         await self.model.update(
             ignore_keys(payload, ['add_upstream_for_block_uuid']),
             update_content=update_content,
         )
-        switch_active_kernel(PIPELINE_TO_KERNEL_NAME[self.model.type])
+        try:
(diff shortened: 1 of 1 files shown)


mage_ai/api/resources/PipelineResource.py
```diff
@@ -403,11 +403,19 @@ async def _add_markdown_block(block_doc: str, block_uuid: str, priority: int):
...
         await self.model.update(
             ignore_keys(payload, ['add_upstream_for_block_uuid']),
             update_content=update_content,
         )
-        switch_active_kernel(PIPELINE_TO_KERNEL_NAME[self.model.type])
+        try:
+            switch_active_kernel(PIPELINE_TO_KERNEL_NAME[self.model.type])
+        except Exception as e:
+            pipeline_type_updated = payload.get('type')
+            if pipeline_type_updated is not None:
```
wangxiaoyou1993, line 415: also check `pipeline_type_updated != orig_pipeline_type`

mage_ai/api/resources/PipelineResource.py
```diff
@@ -403,11 +403,19 @@ async def _add_markdown_block(block_doc: str, block_uuid: str, priority: int):
...
+                await self.model.update(dict(type=pipeline_type))
+
+            raise Exception(e)
```
wangxiaoyou1993, line 418: nit: you can just do `raise e`

mage_ai/api/resources/PipelineResource.py
johnson-mage: @wangxiaoyou1993 I was gonna add that check but saw that we have one [here](https://github.com/mage-ai/mage-ai/blob/master/mage_ai/data_preparation/models/pipeline.py#L831-L836) already. Should we still check if the pipeline type is different from the original type here?

wangxiaoyou1993: that check includes many fields. if any of the field is different, set `should_save` to true.

the check here is to decide whether another DB update is needed.

johnson-mage: ah, got it. thx

wangxiaoyou1993 (approved):

PR #3886, opened 2023-11-02
[td] Tree v2.0 and bug bash

tommydangerous:
Reverts mage-ai/mage-ai#3874

### DIFF
--- a/mage_ai/data_preparation/models/block/__init__.py
+++ b/mage_ai/data_preparation/models/block/__init__.py
@@ -580,6 +580,7 @@ class Block(DataIntegrationMixin, SparkBlock):
         pipeline = kwargs.get('pipeline')
         if pipeline is not None:
             priority = kwargs.get('priority')
+            downstream_block_uuids = kwargs.get('downstream_block_uuids', [])
             upstream_block_uuids = kwargs.get('upstream_block_uuids', [])
 
             if BlockType.DBT == block.type and block.language == BlockLanguage.SQL:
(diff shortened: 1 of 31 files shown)


mage_ai/data_preparation/models/pipeline.py
```diff
@@ -1255,8 +1255,10 @@ def update_block(
         self,
         block: Block,
         callback_block_uuids: List[str] = None,
+        check_downstream_block_order: bool = False,
```
wangxiaoyou1993, line 1258: These two params seem unused.

mage_ai/data_preparation/models/block/__init__.py
```diff
@@ -598,6 +599,17 @@ def after_create(self, block: 'Block', **kwargs) -> None:
                     priority=priority,
                     widget=widget,
                 )
+                if downstream_block_uuids:
```
wangxiaoyou1993, line 602: I would add this logic to `pipeline.add_lock` so that we can validate the pipeline circular dependency issues after setting upstream and downstream blocks.

mage_ai/data_preparation/models/block/__init__.py
```diff
@@ -2111,6 +2125,36 @@ def update(self, data, **kwargs) -> 'Block':
                 check_upstream_block_order=check_upstream_block_order,
             )
 
+        if 'downstream_blocks' in data and self.pipeline:
```
wangxiaoyou1993, line 2128: Same here I would add this logic to to `pipeline.update_block` method so that we can validate the pipeline after updating the dependency.

mage_ai/data_preparation/models/block/__init__.py
```diff
@@ -2111,6 +2125,36 @@ def update(self, data, **kwargs) -> 'Block':
...
+                    dict(
+                        upstream_blocks=list(filter(
+                            lambda x, uuid=self.uuid: x != uuid,
+                            block.upstream_block_uuids or [],
+                        )),
+                    ),
+                    check_upstream_block_order=check_upstream_block_order,
+                )
+
+            for block_uuid in arr:
```
wangxiaoyou1993, line 2147: Not all the block uuid in the `arr` are all new downstream block.

tommydangerous: I don’t understand what you’re suggesting here.

mage_ai/data_preparation/models/block/__init__.py
tommydangerous: Done.

mage_ai/data_preparation/models/pipeline.py
tommydangerous: Done.

mage_ai/data_preparation/models/block/__init__.py
tommydangerous: Done.

PR #3889, opened 2023-11-02
[ac] Add keyboard shortcuts for inserting new scratchpad cell

anniexcheng:
# Description

This PR adds the same keyboard shortcuts as Jupyter notebooks for inserting new scratchpad cells:

* If a block is selected and the user presses `a`, a new scratchpad cell is inserted above the selected block.
* If a block is selected and the user presses `b`, a new scratchpad cell is inserted below the selected block.
* If no block is selected or the user is typing in an input / text area, the `a` and `b` keyboard shortcuts do nothing.

# How Has This Been Tested?

- [x] Test that `a` inserts a new scratchpad cell **above** the selected block when a block is selected and an input / text area isn't focused
- [x] Test that `b` inserts a new scratchpad cell **below** the selected block when a block is selected and an input / text area isn't focused
- [x] Test that the `a` and `b` keyboard shortcuts do nothing when no block is selected
- [x] Test that the `a` and `b` keyboard shortcuts do nothing when an input / text area is focused

cc: @johnson-mage

### DIFF
--- a/mage_ai/frontend/components/CodeEditor/index.tsx
+++ b/mage_ai/frontend/components/CodeEditor/index.tsx
@@ -179,6 +179,10 @@ function CodeEditor({
       setTextareaFocused?.(true);
     });
 
+    editor.onDidBlurEditorWidget(() => {
+      setTextareaFocused?.(false);
+    });
+
     editor.onDidContentSizeChange(({
       contentHeight,
       contentHeightChanged,
--- a/mage_ai/frontend/components/PipelineDetail/ConfigureBlock/index.tsx
+++ b/mage_ai/frontend/components/PipelineDetail/ConfigureBlock/index.tsx
@@ -1,6 +1,6 @@
 import { DndProvider } from 'react-dnd';
 import { HTML5Backend } from 'react-dnd-html5-backend';
-import { useEffect, useMemo, useRef, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { useMutation } from 'react-query';
 
 import BlockCubeGradient from '@oracle/icons/custom/BlockCubeGradient';
@@ -90,6 +90,30 @@ function ConfigureBlock({
     name: defaultName,
     type: block?.type,
   });
+
+  const handleOnSave = useCallback(() => {
+    onSave({
+      ...blockAttributes,
+      name: blockAttributes?.name || defaultName,
+    });
+  }, [blockAttributes, defaultName, onSave]);
+
+  useEffect(() => {
+    const handleKeyDown = (event) => {
+      event.stopPropagation();
+      if (event.key === 'Escape') {
+        onClose();
+      } else if (event.key === 'Enter') {
+        handleOnSave();
+      }
+    };
+
+    document?.addEventListener('keydown', handleKeyDown);
+
+    return () => {
+      document?.removeEventListener('keydown', handleKeyDown);
+    };
+  }, [handleOnSave, onClose]);
 
   useEffect(() => {
     refTextInput?.current?.focus();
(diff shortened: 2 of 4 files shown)


mage_ai/frontend/components/CodeEditor/index.tsx
```diff
@@ -163,6 +163,10 @@ function CodeEditor({
       setTextareaFocused?.(true);
     });
 
+    editor.onDidBlurEditorWidget(() => {
+      setTextareaFocused?.(false);
```
anniexcheng, line 167: `textareaFocused` was previously not accurate because we didn't set it to false when text area focus was lost

mage_ai/frontend/components/PipelineDetail/index.tsx
```diff
@@ -490,7 +541,21 @@ function PipelineDetail({
           const selectedBlockIndex =
             blocks.findIndex(({ uuid }: BlockType) => selectedBlock.uuid === uuid);
 
-          if (keyMapping[KEY_CODE_ESCAPE]) {
+          if (onlyKeysPresent([KEY_CODE_A], keyMapping) && !event.repeat && selectedBlockIndex !== -1) {
+            if (!isInputElement(event)) {
```
anniexcheng, line 545: In most cases, we don't need this because we already check that `textareaFocused = false` but this is here in case there's another input field (e.g. filter) on the page

mage_ai/frontend/components/PipelineDetail/index.tsx
```diff
@@ -465,6 +516,20 @@ function PipelineDetail({
       ) {
         event.preventDefault();
         savePipelineContent();
+      } else if (onlyKeysPresent([KEY_CODE_A], keyMapping) || onlyKeysPresent([KEY_CODE_B], keyMapping)) {
+        if (!event.repeat && !isInputElement(event)) {
```
anniexcheng, line 520: We check if the an input / text area is focused directly rather than using `textareaFocused` because `textareaFocused` is not always accurate :'(

johnson-mage:
@anniexcheng The keyboard shortcut for adding scratchpad block seems to work well! For `scratchpad` blocks specifically, can you enable the `escape` key shortcut to close the `ConfigureBlock` modal that pops up please? If a user is using keyboard shortcuts to add scratchpad blocks, they prob also would like to be able to close the modal via the keyboard as well.

anniexcheng:
> ConfigureBlock

@johnson-mage Is there a specific reason we want to enable the `escape` keyboard shortcut specifically for `scratchpad` blocks only? It might be nice to just enable `escape` to close the modal and `enter` to save the block for `ConfigureBlock` generally!

mage_ai/frontend/components/PipelineDetail/ConfigureBlock/index.tsx
```diff
@@ -91,6 +91,30 @@ function ConfigureBlock({
...
 
+  const handleOnSave = useCallback(() => {
+    onSave({
+      ...blockAttributes,
+      name: blockAttributes?.name || defaultName,
+    });
+  }, [blockAttributes, defaultName, onSave]);
+
+  useEffect(() => {
+    const handleKeyDown = (event) => {
```
anniexcheng, line 102: We don't use `useKeyboardContext` here b/c it doesn't work well with modals - it still captures keyboard shortcuts for the underlying page (even when stopping event propagation).

johnson-mage (approved):
LGTM. @anniexcheng You might've seen this `disableEscape` [setting](https://github.com/mage-ai/mage-ai/blob/master/mage_ai/frontend/pages/pipelines/%5Bpipeline%5D/edit.tsx#L1854) for the `useModal` hook already, but just mentioning it here in case you didn't see it or need to use it in the future. Setting it to `false` (or not including it at all) allows you to press `Escape` to close the modal, but your approach of adding the `handleKeyDown` event listener works better for avoiding keyboard conflicts as you mentioned. In this case, pressing `escape` would have also deselected the block in addition to closing the modal if we just used the built-in escape-to-close shortcut in the `useModal` hook.

johnson-mage:
@anniexcheng This PR was reverted in this [commit](https://github.com/mage-ai/mage-ai/commit/39a7e8665b3ac5ac085346dc15f65ce63b5efb0d) due to a bug discovered that prevented the Chart block code editor in the Sidekick (on Pipeline Editor page) from being focused. As a result, the chart blocks could not be edited.

PR #3899, opened 2023-11-03
[ac] Show dependency graph zoom options

anniexcheng:
# Description

Previously, we only displayed zoom controls for the dependency graph on the Pipeline Edit page; other instances of the dependency graph were not accompanied by zoom controls:

<div style="display: inline;">
  image
  image
</div>

This PR extends the zoom controls to all instances of the dependency graph. We move the zoom controls into the `DependencyGraph` component itself and use a new floating action bar design. When the dependency graph container width is too narrow, we hide the action buttons and only show the zoom level:
<div style="display: inline;">
  image
  image
</div>

As per Kai's request, we also update the zoom factor to be 50% rather than 10%. When you zoom in, the zoom level increases by 50% and when you zoom out, the zoom level decreases by 50%. The min zoom level (30%) and max zoom level (200%) remain unchanged.

# How Has This Been Tested?

- [x] Check that full zoom controls are shown when the dependency graph is wide enough
- [x] Check that only the zoom display is shown when the dependency graph is too narrow
- [x] Check that reset, zoom in, and zoom out functionalities work as expected
- [x] Check that there are tool tips for all zoom controls
- [x] Check that the zoom controls accompany all dependency graph instances

    https://github.com/mage-ai/mage-ai/assets/8130751/9fa829b8-18ad-47ca-a0cb-14da4fc5aeb2

cc: @johnson-mage


johnson-mage:
@anniexcheng Looks great. Just one comment. When zooming in or out by clicking the buttons in the zoom controls, can we disable the "double-click to reset" shortcut on the canvas? If user tries to zoom in or out quickly, it can cause an odd zooming loop where it zooms a bit but then resets and continues zooming if the user is clicking the zoom buttons quickly (which I've done if i want to get to the max or min zoom fast).

zoom reset loop

anniexcheng:
> @anniexcheng Looks great. Just one comment. When zooming in or out by clicking the buttons in the zoom controls, can we disable the "double-click to reset" shortcut on the canvas? If user tries to zoom in or out quickly, it can cause an odd zooming loop where it zooms a bit but then resets and continues zooming if the user is clicking the zoom buttons quickly (which I've done if i want to get to the max or min zoom fast).
> 
> zoom reset loop [ zoom reset loop ](https://user-images.githubusercontent.com/78053898/281165918-e8878097-b05e-4b74-9afe-161670daa5ca.gif) [ ](https://user-images.githubusercontent.com/78053898/281165918-e8878097-b05e-4b74-9afe-161670daa5ca.gif)

ooo great catch!! will fix 😄

johnson-mage:
@anniexcheng Instead of completely removing the double-click to reset zoom shortcut, can we just disable it when zooming in or out? Like maybe disabling it on hovering over the zoom buttons. I do think it's nice to have the shortcut there since there are instances where the dep graph zoom isnt fitted after loading and double-clicking to reset is a lot easier.

johnson-mage (approved):

PR #3905, opened 2023-11-05
[ac] Enable users to cancel in progress runs when disabling a pipeline trigger

anniexcheng:
# Description

Previously, in the Pipeline Triggers table, you could disable a pipeline trigger and prevent future pipeline runs for that trigger. However, current in progress pipeline runs would continue to run to completion. **This PR adds the option to cancel in progress runs when disabling a pipeline trigger.**

If a pipeline trigger is enabled in the Pipeline Triggers table and a user clicks the toggle to disable it, the following happens:

* If the pipeline trigger has current in progress run(s), we show a popup modal that gives the user 2 options: allow in progress runs to continue OR stop them immediately.
* If the user clicks "Allow", the pipeline trigger will be disabled, and the current in progress runs will continue to run (like the previous behavior).
* If the user clicks "Stop", the pipeline trigger will be disabled, and the current in progress runs will be immediately cancelled.

# How Has This Been Tested?

- [x] Test that enabling a disabled trigger works properly
- [x] Test that disabling an enabled trigger with no current in progress runs works properly
- [x] Test that disabling an enabled trigger with current in progress runs shows the pop up modal
- [x] Test that clicking "Allow" in the popup modal allows the in progress runs to continue
- [x] Test that clicking "Stop" in the popup modal immediately cancels the in progress runs

https://github.com/mage-ai/mage-ai/assets/8130751/36f1e083-5b15-48dd-9389-bb194d35a28e

https://github.com/mage-ai/mage-ai/assets/8130751/7426dc5b-7ddf-4aa3-9c01-b4884b596327

cc: @johnson-mage


mage_ai/frontend/components/Triggers/Table.tsx
```diff
@@ -211,6 +232,68 @@ function TriggersTable({
...
+          id: pipelineScheduleId,
+          status: ScheduleStatusEnum.INACTIVE,
+        });
+      }}
+      onStop={(pipelineScheduleId) => {
+        hideDisableTriggerModal();
+        // Cancel all in progress runs
+        updatePipeline({
+          pipeline: {
+            pipelineScheduleId: pipelineScheduleId,
```
johnson-mage, line 257: @anniexcheng I believe the object key here should be `pipeline_schedule_id` instead of `pipelineScheduleId`.

johnson-mage:
@anniexcheng Can you also test with [user authentication](https://docs.mage.ai/production/authentication/overview#require-user-authentication) enabled and as an `editor` role? I ran into a `PipelinePolicy` error so you may need to add `pipeline_schedule_id` [here](https://github.com/mage-ai/mage-ai/blob/master/mage_ai/api/policies/PipelinePolicy.py#L110-L113).

mage_ai/frontend/oracle/icons/index.tsx
```diff
@@ -1644,3 +1644,20 @@ export const LayoutSplit = build([
   fill: 'none',
   viewBox: '0 0 16 16',
 });
+
+export const Warning = build([
```
johnson-mage, line 1648: @anniexcheng We actually have this icon already. It's called `AlertTriangle` so you can remove the `Warning` icon since it's a duplicate.
image

mage_ai/frontend/components/Triggers/Table.tsx
johnson-mage: The reason it's still able to cancel the running pipeline runs is because it's hitting the `else` block of the `cancel_pipeline_runs` method in `PipelineResource`, so it cancels ALL running pipeline runs for the pipeline (not just for the runs under the trigger).

johnson-mage:
@anniexcheng You'll need to test this feature on the `/triggers` page as well. The `pipeline` doesn't actually get passed into the `TriggersTable` component since it's the overall Triggers list page for all pipelines, so we get a 405 error when trying to stop the running pipeline runs.

image

mage_ai/frontend/components/Triggers/Table.tsx
anniexcheng: oh derp, really good catch!

johnson-mage:
@anniexcheng Let's also add an `errors` state to the `mage_ai/frontend/pages/triggers/index.tsx` file so we can pass in the `setErrors` prop to the `TriggersTable`/`Dashboard` components, and the `errors` prop to the `Dashboard`. Otherwise an error like a policy error will just be silent if something goes wrong when making the PUT requests.

johnson-mage:
> Can we also include the stop/allow running runs popup in the Trigger detail page for when a user pauses the trigger and there are running runs? image

This will be done in a separate PR.

johnson-mage:
@wangxiaoyou1993 I'm ready to approve this PR, but there was some small changes to the BE code. It looks good to me, but can you please take a quick look in case I missed anything?

johnson-mage (approved):
Also should have a second look for the BE changes

wangxiaoyou1993 (approved):
BE change looks good.
Generally, there're some performance issues (N+1 query issue on `pipeline_runs_count`, `pipeline_in_progress_runs_count` and `tags` fields) of the pipeline_schedule LIST API. But I can optimize them together later.

PR #3972, opened 2023-11-13
[Fix] Backfill's date-picker date value mismatch

edmondwinston:
# Description

The problem existed due to the use of `.toISOSString()` when formatting the selected date, which will add time zone offset. Fixed using (the already exists) `moment`.

# How Has This Been Tested?

- Before
Animation

- After
a

# Checklist

cc: @johnson-mage

### DIFF
--- a/mage_ai/frontend/components/Triggers/utils.ts
+++ b/mage_ai/frontend/components/Triggers/utils.ts
@@ -3,9 +3,9 @@ import moment from 'moment';
 import BlockRunType from '@interfaces/BlockRunType';
 import PipelineScheduleType from '@interfaces/PipelineScheduleType';
 import {
-  DATE_FORMAT_DISPLAY,
   DATE_FORMAT_LONG_NO_SEC,
   DATE_FORMAT_LONG_NO_SEC_WITH_OFFSET,
+  DATE_FORMAT_LONG_T_SEP,
   dateFormatLong,
 } from '@utils/date';
 import { DEFAULT_PORT } from '@api/utils/url';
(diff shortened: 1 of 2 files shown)


johnson-mage:
@edmondwinston After saving a backfill date and then going back to edit it, does the start/end date go back 1 day for you? For example, please see the gif below. I would expect the date to stay the same upon clicking `Edit`, making no changes, and then saving again, but if I click `Edit`, make no changes, and then save, the start/end date goes back 1 day.

date goes back 1 day on edit and save

edmondwinston:
@johnson-mage 

I figure the problem is that we store datetime in UTC, and then when displaying we're parsing them to UTC again. Tested the latest code with different time zones, but please also try it yourself.

Also, I'm having the idea of making another PR removing the `convertToUtc` option, since we would never be converting if we always store them in UTC from the first place. `shouldDisplayLocalTimezone()` doesn't have to be call at every components here and there anymore - they could be called when parsing. What do you think?

johnson-mage:
> I figure the problem is that we store datetime in UTC, and then when displaying we're parsing them to UTC again. Tested the latest code with different time zones, but please also try it yourself.

Cool, I'll test it out.

> Also, I'm having the idea of making another PR removing the `convertToUtc` and `localTimezone` option, since we would never be converting if we always store them in UTC from the first place. `shouldDisplayLocalTimezone()` doesn't have to be call at every components here and there anymore - they could be called when parsing. What do you think?

This would require a lot of testing, since there's a lot of different areas in the app that display times and could be hard to catch everything. I think there are just a bunch of different scenarios where the datetimes are not always a consistent format, which is why you have several different methods for parsing the dates. It could definitely use some cleanup, but doesn't seem high-priority if the times are being displayed correctly already. The issue you ran into with selecting the date from the datepicker definitely needs to be fixed, though it seems to be an edge case that doesn't happen for most users.

johnson-mage (approved):

PR #4087, opened 2023-11-28
[dy] Optimize calculate metrics for integration pipelines

dy46:
# Description

When running integration pipelines, the `calculate_metrics` function was being run before the pipeline run started, after every block run, and after the pipeline run was completed. The `calculate_metrics` was calculating pipeline run metrics and block run metrics every time it was called, so if there were a large number of block runs, calculating metrics could take a long time. 

I changed a couple of things to optimize this operation:
1. instead of calculating the metrics for all block runs after each block run was completed, we instead only calculate the metrics for the block run that completed. 
2. instead of calculating every metric when `calculate_metrics` was called, I created separate functions for calculating block run metrics and pipeline run metrics so pipeline run metrics will be calculate before and after the pipeline run, and block run metrics will only be calculated when the pipeline run is in progress.

# How Has This Been Tested?

- [x] Tested with an integration pipeline to ensure the metrics before and after the change were unchanged
- [x] Verified that the blocks were run more efficiently by running an integration pipeline with 10000 block runs
  * before change -> 2 blocks completed in 5 minutes
  * after change -> 354 blocks completed in 5 minutes

# Checklist

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/data_integrations/utils/scheduler.py
+++ b/mage_ai/data_integrations/utils/scheduler.py
@@ -24,7 +24,7 @@ from mage_ai.data_preparation.models.pipelines.integration_pipeline import (
 from mage_ai.data_preparation.models.triggers import ScheduleInterval
 from mage_ai.orchestration.db import db_connection
 from mage_ai.orchestration.db.models.schedules import BlockRun, PipelineRun
-from mage_ai.orchestration.metrics.pipeline_run import calculate_metrics
+from mage_ai.orchestration.metrics.pipeline_run import calculate_pipeline_run_metrics
 from mage_ai.shared.array import find
 from mage_ai.shared.hash import index_by, merge_dict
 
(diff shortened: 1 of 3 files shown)


wangxiaoyou1993:
Did you test the performance change?

dy46:
I haven't yet. Let me try to test the performance change

dy46:
@wangxiaoyou1993 i tested the performance change. i put the results in the PR description. i also changed a lot of the previous logic because i realized it wasn't working properly.

wangxiaoyou1993:
@dy46 can you describe what wasn't working properly before?

dy46:
@wangxiaoyou1993 oh i meant in the first iteration of this PR. it was still taking a while to calculate metrics

dy46:
when there were a lot of block runs, it was iterating through every block run and fetching the logs for the block run, so it was really slow

wangxiaoyou1993 (approved):

PR #4091, opened 2023-11-28
[jk] Display error in UI when variables directories configured incorrectly

johnson-mage:
# Description
- In the project's `metadata.yaml` configuration file, it could be difficult or cryptic to identify that the `variables_dir` or `remote_variables_dir` was configured incorrectly (e.g. due to an empty variable or missing value). This PR checks for issues with those properties and displays an error in the UI with a link to the config file, allowing a user to easily fix their project's configuration.

# How Has This Been Tested?
Tested with an empty `variables_dir` in the config file, as well as an empty interpolated variable for the `remote_variables_dir`:
misconfigured variables directories

# Checklist

### DIFF
--- a/mage_ai/api/presenters/PipelinePresenter.py
+++ b/mage_ai/api/presenters/PipelinePresenter.py
@@ -66,6 +66,14 @@ class PipelinePresenter(BasePresenter):
                     FeatureUUID.DATA_INTEGRATION_IN_BATCH_PIPELINE,
                 )
 
+            include_variables_dir = query.get('include_variables_dir', [False])
+            if include_variables_dir:
+                include_variables_dir = include_variables_dir[0]
+
+            include_remote_variables_dir = query.get('include_remote_variables_dir', [False])
+            if include_remote_variables_dir:
+                include_remote_variables_dir = include_remote_variables_dir[0]
+
             return await self.model.to_dict_async(
                 include_block_catalog=include_block_catalog,
                 include_block_metadata=include_block_metadata,
@@ -77,6 +85,8 @@ class PipelinePresenter(BasePresenter):
                 include_extensions=include_extensions,
                 include_outputs=include_outputs,
                 include_outputs_spark=include_outputs_spark,
+                include_remote_variables_dir=include_remote_variables_dir,
+                include_variables_dir=include_variables_dir,
                 sample_count=DATAFRAME_SAMPLE_COUNT_PREVIEW,
             )
         elif constants.UPDATE == display_format:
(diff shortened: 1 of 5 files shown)


mage_ai/frontend/pages/pipelines/[pipeline]/edit.tsx
```diff
@@ -1331,6 +1325,46 @@ function PipelineDetailPage({
     onChangeCodeBlock,
   ]);
 
+  // Check for pipeline or project config errors
+  useEffect(() => {
+    let dataWithPotentialError = data;
+    let configFileLinks = [];
+    if (!data?.hasOwnProperty('error') && dataDataProviders?.hasOwnProperty('error')) {
+      dataWithPotentialError = dataDataProviders;
+    } else if ((pipeline?.variables_dir?.includes('.mage_data')
```
Comment (johnson-mage) [inline mage_ai/frontend/pages/pipelines/[pipeline]/edit.tsx:1334]: @wangxiaoyou1993 Is checking the `variables_dir` value for `.mage_data` (the default Mage data directory used if no `variables_dir` value provided in the config) a valid way to check for a missing `variables_dir` property? Jw if there's a scenario where users might actually want to use that folder for their variables.

Comment (wangxiaoyou1993) [reply mage_ai/frontend/pages/pipelines/[pipeline]/edit.tsx:1334]: That's the default value for variables_dir so that's a valid value.

Comment (johnson-mage) [reply mage_ai/frontend/pages/pipelines/[pipeline]/edit.tsx]: We just check if the `variables_dir` and `remote_variables_dir` values have `None` in it, which could be caused by using an env var that has an empty or no value. If either of those do, we display an error popup with a link to the project's `metadata.yaml` file.

Leaving the `variables_dir` and `remote_variables_dir` values blank in the `metadata.yaml` config is valid, so no error popup will appear if those values are blank, since the `variables_dir` will just be the default value and the `remote_variables_dir` is not required.

mage_ai/api/presenters/PipelinePresenter.py
```diff
@@ -66,6 +66,14 @@ async def present(self, **kwargs):
                     FeatureUUID.DATA_INTEGRATION_IN_BATCH_PIPELINE,
                 )
 
+            include_variables_dir = query.get('include_variables_dir', [False])
```
wangxiaoyou1993, line 69: You don't necessarily need to add these two params.
Including variables_dir and remote_variables_dir don't need extra IO or DB operations. So you can just include them in the base response payload.

johnson-mage: Ah ok. Updated.

PR #4128, opened 2023-12-05
[dy] Fix external cloud storage logs for k8s blocks

dy46:
# Description

When using external cloud storage for blocks that use the k8s executor, the logs for the block were not getting uploaded correctly. This was happening because the logs were getting written to the destination twice, so the logs in the 1st write were getting overwritten. This PR will try to fetch logs from the destination first and append the new logs to the existing logs if they exist.

# How Has This Been Tested?

- [x] Tested locally with a block using a k8s executor and S3 external log storage
Before:
Screenshot 2023-12-05 at 11 57 07 AM

After:
Screenshot 2023-12-05 at 3 13 12 PM

# Checklist

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/services/k8s/job_manager.py
+++ b/mage_ai/services/k8s/job_manager.py
@@ -11,7 +11,6 @@ from mage_ai.services.k8s.constants import (
     DEFAULT_NAMESPACE,
     KUBE_CONTAINER_NAME,
     KUBE_POD_NAME_ENV_VAR,
-    KUBE_POD_NAMESPACE_ENV_VAR,
 )
 from mage_ai.shared.hash import merge_dict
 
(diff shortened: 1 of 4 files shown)


dy46:
actually i just noticed there are some extra logs in the screenshot... let me figure out why that's happening

dy46:
it was an issue caused by the way my testing environment was set up. i added an updated screenshot to the testing section

mage_ai/services/k8s/job_manager.py
```diff
@@ -36,7 +35,7 @@ def __init__(
 
         self.pod_config = self.core_api_client.read_namespaced_pod(
             name=os.getenv(KUBE_POD_NAME_ENV_VAR),
-            namespace=os.getenv(KUBE_POD_NAMESPACE_ENV_VAR),
+            namespace=self.namespace,
```
wangxiaoyou1993, line 38: what's this for?

wangxiaoyou1993: This is a bit different.
The namespace passed to the JobManager `__init__` is the namespace for the job to be created.
The `os.getenv(KUBE_POD_NAMESPACE_ENV_VAR)` is the namespace of the current pod. The code here gets the config of the current pod.
The two values can be different.

I think the code can be updated to `os.getenv(KUBE_POD_NAMESPACE_ENV_VAR, self.namespace)` in case the  KUBE_POD_NAMESPACE_ENV_VAR environment variable is not set.

PR #4133, opened 2023-12-06
[jk] Manually run pipeline once in same trigger

johnson-mage:
# Description
- Add button to Trigger Detail page for running pipeline once in same trigger, which can be more convenient than clicking the `Run@once` button on the Triggers list page because it does not require the user to re-enter overwritten runtime variables. This button is hidden for streaming pipelines.
- For `@once` triggers, the trigger does not need to be `active` status before running the pipeline once using the `Run@once` button in the Trigger Detail page. The trigger will automatically be updated to `active` status. However, if the trigger is NOT an `@once` trigger (e.g. a recurring interval or API trigger), the trigger must be set to `active` status before manually running the pipeline once from the Trigger Detail page.
- Renamed "start/pause" trigger on Trigger Detail page to "enable/disable" trigger.

# How Has This Been Tested?
Added `Run@once` button to Trigger Detail page:
image

Manually running pipeline once from an `@once` trigger (trigger does not need to be `active` status):
run once trigger

Manually running pipeline once from a recurring trigger (button is disabled if the trigger is NOT `active` status):
recurring trigger

# Checklist


johnson-mage:
> should we just remove the button if the user can't click on it?

We have it disabled cuz user can click it after they enable the trigger.
image

dy46:
I'm just worried because the tooltip doesn't show up for a little bit, so it could be confusing

johnson-mage:
> I'm just worried because the tooltip doesn't show up for a little bit, so it could be confusing

dy46 (approved):

PR #4149, opened 2023-12-08
[edw] Add custom 404 page

edmondwinston:
# Description
This PR will add a custom 404 page, which (1) display an error toast telling the user the previous URL doesn't exist, and then redirect them to `/pipelines`.

# How Has This Been Tested?
404 page redirect

# Checklist

cc: @johnson-mage

### DIFF
--- a/mage_ai/frontend/pages/404.tsx
+++ b/mage_ai/frontend/pages/404.tsx
@@ -3,6 +3,7 @@ import { redirectToUrl } from '@utils/url';
 import Headline from '@oracle/elements/Headline';
 import Text from '@oracle/elements/Text';
 import Spacing from '@oracle/elements/Spacing';
+import GradientLogoIcon from '@oracle/icons/GradientLogo';
 
 function NotFoundPage() {
   const [secondsToRedirect, setSecondsToRedirect] = useState(5);
@@ -28,10 +29,13 @@ function NotFoundPage() {
         marginTop: '16rem',
       }}
     >
-      <Spacing mb={1}>
+      <GradientLogoIcon height={64} />
+      <Spacing my={2}>
         <Headline level={1}>404 - Page Not Found</Headline>
       </Spacing>
-      <Text>You will be redirected to the Pipelines dashboard in {secondsToRedirect} seconds.</Text>
+      <Text large>
+        You will be redirected to the Pipelines dashboard in {secondsToRedirect} seconds.
+      </Text>
     </main>
   );
 }
--- a/mage_ai/frontend/pages/404.tsx
+++ b/mage_ai/frontend/pages/404.tsx
@@ -1,18 +1,38 @@
-import { useEffect } from 'react';
+import { useEffect, useState } from 'react';
 import { redirectToUrl } from '@utils/url';
-import { toast } from 'react-toastify';
+import Headline from '@oracle/elements/Headline';
+import Text from '@oracle/elements/Text';
+import Spacing from '@oracle/elements/Spacing';
 
 function NotFoundPage() {
+  const [secondsToRedirect, setSecondsToRedirect] = useState(5);
   useEffect(() => {
-    toast.error('The previous page does not exist.');
-    redirectToUrl('/pipelines');
-  }, []);
+    const i = setInterval(() => {
+      if (secondsToRedirect > 0) {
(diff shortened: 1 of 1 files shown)


edmondwinston:
@johnson-mage 

I left the testing section empty because I still have 2 concerns.

1. The toast is being displayed twice on localhost. I'm not super familiar with React, so although I found answers saying the useEffect behaves like that only in development, but I'd still like to have you confirm this on your side.
2. To be honest, the best behaviour would be setTimeout at the 404 page, display text saying something like 'This URL does not exist, you will be redirected to .. now'. This would reduce a ton of confusion, and give users a chance to navigate around on their own initiative. What do you think?

johnson-mage:
> @johnson-mage
> 
> I left the testing section empty because I still have 2 concerns.
> 
> 1. The toast is being displayed twice on localhost. I'm not super familiar with React, so although I found answers saying the useEffect behaves like that only in development, but I'd still like to have you confirm this on your side.
> 2. To be honest, the best behaviour would be setTimeout at the 404 page, display text saying something like 'This URL does not exist, you will be redirected to .. now'. This would reduce a ton of confusion, and give users a chance to navigate around on their own initiative. What do you think?

1. Actually not completely sure why the useEffect hook was executing twice, but you can use the `useRef` solution [here](https://stackoverflow.com/questions/73732474/handling-oauth-with-react-18-useeffect-hook-running-twice) to make sure the toast displays only once.
2. Good suggestion! On the 404 page, can you please center with larger text (you can use the `Headline` and `Text` components) the following? Also make sure the background color on the 404 page matches a similar color to the app (e.g. `#1E1F24`, which is the background.page color in the `mage_ai/frontend/oracle/styles/themes/dark.ts` file).
```
404 - Page not found
You will be redirected to the Pipelines dashboard in X seconds...
```
The `X` should countdown from 5 seconds, so after 5 seconds, the page redirects to the `/pipelines` page. Then we don't need the toast message anymore. Thank you!

edmondwinston:
> Also make sure the background color on the 404 page matches a similar color to the app (e.g. #1E1F24, which is the background.page color in the mage_ai/frontend/oracle/styles/themes/dark.ts file).

I'm not sure about this one - I found a bunch of utilities in the themes folder, only to figure out they have never been used. The colour is already the dark background one, because it is set in `_app.tsx` already.

mage_ai/frontend/pages/404.tsx
```diff
@@ -0,0 +1,39 @@
...
+        display: 'flex',
+        flexDirection: 'column',
+        height: '100vh',
+        marginTop: '16rem',
+      }}
+    >
+      <Spacing mb={1}>
+        <Headline level={1}>404 - Page Not Found</Headline>
+      </Spacing>
+      <Text>You will be redirected to the Pipelines dashboard in {secondsToRedirect} seconds.</Text>
```
johnson-mage, line 34: Can you add the `large` prop to the Text component please?

johnson-mage:
> > Also make sure the background color on the 404 page matches a similar color to the app (e.g. #1E1F24, which is the background.page color in the mage_ai/frontend/oracle/styles/themes/dark.ts file).
> 
> I'm not sure about this one - I found a bunch of utilities in the themes folder, only to figure out they have never been used. The colour is already the dark background one, because it is set in `_app.tsx` already.

No worries, looks like we don't need it.

mage_ai/frontend/pages/404.tsx
```diff
@@ -0,0 +1,39 @@
...
+    <main
+      style={{
+        alignItems: 'center',
+        display: 'flex',
+        flexDirection: 'column',
+        height: '100vh',
+        marginTop: '16rem',
+      }}
+    >
+      <Spacing mb={1}>
```
johnson-mage, line 31: Can you add the Mage logo here and change the Spacing margin prop to `my` with value of 2 (so it'd be 16px of padding above and below the "404 - Page Not Found")? You can import the logo with the following:
```
import GradientLogoIcon from '@oracle/icons/GradientLogo';
```

johnson-mage:
@edmondwinston Just a couple minor comments and it shud be good to go. The redirect was a nice touch!

johnson-mage (approved):

PR #4227, opened 2023-12-20
allow filtering of block runs using URL params

hjhdaniel:
# Description
- Issue: It is currently a little difficult to find failed blocks in `Block Runs` page as it does not allow filtering. Sorting is insufficient if there are many dynamic block runs (~1000 block runs)
- PR Solution: Allow filtering of block runs using URL params in `Block Runs` page. Currently only supports filtering by `status` but can potentially expand/refactor to extend filtering functionality
- Note: No UI changes for applying the filter for now (just allow filter by adding URL param manually). Ideally the block filtering should be separate from URL params (which i believe are more for page level fields). In the future `blockRunsFilter` can be a state in itself which is updated based on some UI components/handling instead of the URL params (similar to the side panel in logs page)

# How Has This Been Tested?

- [x] Local testing: base URL with no `status` param
- [x] Local testing: base URL with `status` param only
- [x] Local testing: base URL with `status` and other params (such as sorting and page)
- [x] Local testing: checked that linking to the block details page is unaffected for all 3 cases

# Checklist

cc:
@oonyoontong, @johnson-mage 

(P.S. not sure why I am unable to assign reviewers or labels to the PR)

### DIFF
--- a/mage_ai/frontend/pages/pipelines/[pipeline]/runs/[run]/index.tsx
+++ b/mage_ai/frontend/pages/pipelines/[pipeline]/runs/[run]/index.tsx
@@ -34,7 +34,9 @@ import buildTableSidekick, {
 import { PageNameEnum } from '@components/PipelineDetailPage/constants';
 import { PADDING_UNITS } from '@oracle/styles/units/spacing';
 import { SortDirectionEnum, SortQueryEnum } from '@components/shared/Table/constants';
+import { RunStatus as RunStatusEnum } from '@interfaces/BlockRunType';
 import { TabType } from '@oracle/components/Tabs/ButtonTabs';
+import _ from 'lodash';
 import { datetimeInLocalTimezone } from '@utils/date';
 import { onSuccess } from '@api/utils/response';
 import { pauseEvent } from '@utils/events';
@@ -47,6 +49,19 @@ type PipelineBlockRunsProps = {
   pipeline: PipelineType;
   pipelineRun: PipelineRunType;
 };
+
+type BlockRunsFilter = {
+  status?: RunStatusEnum
+}
+
+function FilterBlockRuns (blockRuns: BlockRunType[], blockRunsFilter: BlockRunsFilter) {
+  const filteredBlockRuns = blockRuns.filter((blockRun) => {
+    const blockRunProperties = _.pick(blockRun, Object.keys(blockRunsFilter))
+    const accepted = _.isEqual(blockRunProperties, blockRunsFilter)
+    return accepted
+  })
+  return filteredBlockRuns
+}
 
 function PipelineBlockRuns({
   pipeline: pipelineProp,
@@ -103,6 +118,7 @@ function PipelineBlockRuns({
   };
   const sortColumnIndexQuery = q?.[SortQueryEnum.SORT_COL_IDX];
   const sortDirectionQuery = q?.[SortQueryEnum.SORT_DIRECTION];
+  const blockRunsFilter: BlockRunsFilter = _.omitBy({status: q?.status}, value => value === undefined);
   if (sortColumnIndexQuery) {
     const blockRunSortColumn = COL_IDX_TO_BLOCK_RUN_ATTR_MAPPING[sortColumnIndexQuery];
     const sortDirection = sortDirectionQuery || SortDirectionEnum.ASC;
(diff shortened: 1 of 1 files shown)


oonyoontong (approved):

johnson-mage:
@hjhdaniel Thanks for putting up this PR, ill take a look

johnson-mage:
@hjhdaniel Sorry for the delay in reviewing this. I was on a break for the holidays so wasn't able to get to this until now. If we want to add a block run `status` filter to the Block Runs table, there are several things we should account for:

1) Filtering on the backend instead of on the frontend: This is for consistency with the rest of the app, but also due to issues that would come up if there are many block runs with pagination. For example if we want to view the "failed" block runs while on page 1, but those failed runs are on page 2, they wouldn't appear with your current implementation. You can refer to the [PipelineResource.py file](https://github.com/mage-ai/mage-ai/blob/master/mage_ai/api/resources/PipelineRunResource.py#L125-L126) for an example of how we filter on the backend. 
2) We should include a way to filter the block run statuses in the UI. Otherwise, users would never know they could filter the block runs. The simplest way would be to add a [dropdown menu](https://github.com/mage-ai/mage-ai/blob/master/mage_ai/frontend/pages/pipelines/%5Bpipeline%5D/runs/index.tsx#L558-L589) like in the `/pipelines/[pipeline_uuid]/runs` page, but we may eventually want to allow multiple filters (e.g. filtering by multiple statuses).
image
3) We should allow block run filtering on both the overall block runs page (`/pipelines/demo_pipeline_copy/runs?tab=block_runs`) for a pipeline, as well as the block runs page for an individual pipeline run (`/pipelines/[pipeline_uuid]/runs/[pipeline_run_id]`).
4) We may need to update the [BlockRunPolicy](https://github.com/mage-ai/mage-ai/blob/master/mage_ai/api/policies/BlockRunPolicy.py#L37-L41) as well to include the `status[]` query param. You should test with [user authentication](https://docs.mage.ai/production/authentication/overview#require-user-authentication) enabled since many users use that.

Are you still able to take this feature on? We do have block run sorting, which helps with grouping blocks run by a certain status, but there are other features the Mage team needs to prioritize at the moment. Thanks!

hjhdaniel:
> @hjhdaniel Sorry for the delay in reviewing this. I was on a break for the holidays so wasn't able to get to this until now. If we want to add a block run `status` filter to the Block Runs table, there are several things we should account for:
> 
> 1. Filtering on the backend instead of on the frontend: This is for consistency with the rest of the app, but also due to issues that would come up if there are many block runs with pagination. For example if we want to view the "failed" block runs while on page 1, but those failed runs are on page 2, they wouldn't appear with your current implementation. You can refer to the [PipelineResource.py file](https://github.com/mage-ai/mage-ai/blob/master/mage_ai/api/resources/PipelineRunResource.py#L125-L126) for an example of how we filter on the backend.
> 2. We should include a way to filter the block run statuses in the UI. Otherwise, users would never know they could filter the block runs. The simplest way would be to add a [dropdown menu](https://github.com/mage-ai/mage-ai/blob/master/mage_ai/frontend/pages/pipelines/%5Bpipeline%5D/runs/index.tsx#L558-L589) like in the `/pipelines/[pipeline_uuid]/runs` page, but we may eventually want to allow multiple filters (e.g. filtering by multiple statuses).
>    image
> 3. We should allow block run filtering on both the overall block runs page (`/pipelines/demo_pipeline_copy/runs?tab=block_runs`) for a pipeline, as well as the block runs page for an individual pipeline run (`/pipelines/[pipeline_uuid]/runs/[pipeline_run_id]`).
> 4. We may need to update the [BlockRunPolicy](https://github.com/mage-ai/mage-ai/blob/master/mage_ai/api/policies/BlockRunPolicy.py#L37-L41) as well to include the `status[]` query param. You should test with [user authentication](https://docs.mage.ai/production/authentication/overview#require-user-authentication) enabled since many users use that.
> 
> Are you still able to take this feature on? We do have block run sorting, which helps with grouping blocks run by a certain status, but there are other features the Mage team needs to prioritize at the moment. Thanks!

Hi @johnson-mage sorry in that case I might have misunderstood the variable `dataBlockRuns?.blockRuns`

(I thought `dataBlockRuns?.blockRuns` was an array of all the block runs in the given `pipelineRunId`, where pagination is only applied after the filtering of total blocks.)

``` javascript
  const totalBlockRuns = useMemo(() => dataBlockRuns?.metadata?.count || [], [dataBlockRuns]);
  const paginationEl = useMemo(() => (
    <Spacing p={2}>
      <Paginate
        maxPages={MAX_PAGES}
        onUpdate={(p) => {
          const newPage = Number(p);
          const updatedQuery = {
            ...q,
            page: newPage >= 0 ? newPage : 0,
          };
          setSelectedRun(null);
          router.push(
            '/pipelines/[pipeline]/runs/[run]',
            `/pipelines/${pipelineUUID}/runs/${pipelineRunId}?${queryString(updatedQuery)}`,
          );
        }}
        page={Number(page)}
        totalPages={Math.ceil(totalBlockRuns / ROW_LIMIT)}
      />
    </Spacing>
  ), [page, pipelineRunId, pipelineUUID, q, router, totalBlockRuns]);
```

I currently do not have the bandwidth to continue this, so let's archive this for now.

Thank you!

hjhdaniel:
See comments [here](https://github.com/mage-ai/mage-ai/pull/4227#issuecomment-1878124153)

hjhdaniel:
Refer to comment [here](https://github.com/mage-ai/mage-ai/pull/4227#issuecomment-1878124153)

Issue #4102, opened 2023-11-30
Unable to use Polars DataFrames


augcollet:
Hello,

I'm trying to use Polars (0.19.18) with mage-ai (0.9.46). I read that support for Polars was added in 0.8.71 (https://github.com/mage-ai/mage-ai/issues/2495).

However it seems like mage is not able deserialize Polars DataFrame correctly, as demonstrated with the 2 blocks below :

**To Reproduce**
In the first block, I just create a Polars DataFrame, and everything seems to work as expected : 
![image](https://github.com/mage-ai/mage-ai/assets/99130480/22e2fe0f-76c3-4d10-92f8-9190b1071241)

In the second block, this Polars DataFrame is by mistake interpreted as a Pandas DataFrame, so I cant directly apply Polars functions.
![image](https://github.com/mage-ai/mage-ai/assets/99130480/3d0035fb-8c55-4f2b-99c7-3555714982c7)

**Expected behavior**
The Polars DataFrame should be automatically interpreted as a Polars DataFrame in the second block.
I could cast back the Pandas DataFrame to a Polars DataFrame within the block, but I think it should not be a correct solution.

I thank you in advance for your help !

PR #4281, opened 2024-01-04
[dy] Support serialize Polars dataframe

dy46:
# Description

Resolves https://github.com/mage-ai/mage-ai/issues/4102.

In order to serialize a polars dataframe, we have to write some metadata about the output variable so that we can determine whether or not to use polars or pandas when serializing. This PR adds a `type.json` file to the variable directory that currently just stores the type of the variable which will be used when reading the output data. If there is no file, we will fallback to the previous logic.

# How Has This Been Tested?

- [x] Tested locally with the example in the issue
Screenshot 2024-01-03 at 5 33 36 PM
Screenshot 2024-01-03 at 5 33 32 PM
- [x] Tested with S3 remote variable directory
- [x] Unit test

# Checklist

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/data_preparation/models/block/__init__.py
+++ b/mage_ai/data_preparation/models/block/__init__.py
@@ -1886,7 +1886,6 @@ class Block(DataIntegrationMixin, SparkBlock, ProjectPlatformAccessible):
 
         data_products = []
         outputs = []
-        variable_manager = self.pipeline.variable_manager
 
         all_variables = self.get_variables_by_block(
             block_uuid=block_uuid,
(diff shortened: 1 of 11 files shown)


requirements.txt
```diff
@@ -25,7 +25,7 @@ ldap3==2.9.1
 newrelic==8.8.0
 numpy>=1.21.0
 pandas>=1.3.0
-polars<=0.19.2
```
wangxiaoyou1993, line 28: did this version cause any issues?

mage_ai/data_preparation/storage/local_storage.py
```diff
@@ -93,9 +93,12 @@ async def write_json_file_async(self, file_path: str, data) -> None:
             )
             await file.write(fcontent)
 
-    def read_parquet(self, file_path: str, **kwargs) -> pd.DataFrame:
```
wangxiaoyou1993, line 96: Let's not remove `**kwargs`. It'll cause errors.
The S3Storage and GCSStorage also have the `**kwargs` param https://github.com/mage-ai/mage-ai/blob/master/mage_ai/data_preparation/storage/s3_storage.py#L96

mage_ai/data_preparation/models/variable.py
```diff
@@ -273,6 +302,20 @@ async def write_data_async(self, data: Any) -> None:
         else:
             await self.__write_json_async(data)
 
+        self.write_metadata()
+
+    def write_metadata(self) -> None:
```
wangxiaoyou1993, line 307: we use another [`metadata.json`](https://github.com/mage-ai/mage-ai/blob/master/mage_ai/data_preparation/models/variable.py#L607) to store dataframe stats. can we clarify the difference here?

requirements.txt
dy46: the `.cast` function doesn't exist in 0.17.x, and the polars documentation doesn't even have a page for under 0.18, so i just set it as a lower bound

mage_ai/data_preparation/models/variable.py
dy46: ok i wasn't sure if that was still being used. i can change the name of the new file

dy46:
@wangxiaoyou1993 Thanks for catching those issues. I tested it more with S3 and found some issues, so I fixed those as well as addressed your comments. I also changed the name of the file to `type.json` to differentiate it.

mage_ai/data_preparation/models/block/__init__.py
```diff
@@ -2074,6 +2075,24 @@ async def __get_outputs_async(
                 )
                 data_products.append(data)
                 continue
+            if type(data) is pl.DataFrame:
+                row_count, column_count = data.shape
```
wangxiaoyou1993, line 2079: the row_count and column_count are sample data's row and column count.
if we want to get accurate values, we need store them.

mage_ai/data_preparation/models/block/__init__.py
```diff
@@ -2074,6 +2075,24 @@ async def __get_outputs_async(
                 )
                 data_products.append(data)
                 continue
+            if type(data) is pl.DataFrame:
```
wangxiaoyou1993, line 2078: you should use `elif`

requirements.txt
wangxiaoyou1993: how about changing `<=0.19.2` to `<0.19.2`?

dy46:

wangxiaoyou1993 (approved):

PR #4291, opened 2024-01-04
[jk] Prevent unnecessary initial pipeline run

johnson-mage:
# Description
- The scheduler was immediately creating a pipeline run before the current time for a trigger with a start date in the past, even though the "next run date" was correct. Users would expect the initial pipeline run for a trigger to be the "Next run date" as indicated in the UI, but there would be an initial pipeline run created (unexpectedly) for the current execution date period if the current execution date was after the start date but before the current time.
- For example, if the current time is `1/4/24 12:00 UTC` and you create a daily trigger with a start date of `1/3/24 12:00 UTC` and then enable it, a pipeline run with execution date `2024-01-04 00:00` gets immediately created, which can be unintuitive for users since they don’t expect the first pipeline run to be created until the execution date of `2024-01-05 00:00` (the next run date in the future after `1/4/24 12:00`).
- This PR prevents this unnecessary initial pipeline run from being created. It checks if the pipeline schedule's `current_execution_date` is earlier than the `last_enabled_at` date for the pipeline schedule (the `last_enabled_at` date is a new attribute on the PipelineSchedule model). If it is, the pipeline scheduler does not schedule a pipeline run. This can be bypassed by enabling/checking the `Create initial pipeline run` setting when editing a trigger.

# How Has This Been Tested?
- Added unit tests and tested the following scenarios locally:
- [x] Made sure initial pipeline run was not created when creating a new hourly/daily/weekly trigger with a past start date through the UI and then enabling it through the UI
- [X] Made sure initial pipeline run was not created when creating a new daily trigger via code (the `triggers.yaml` file), where the trigger config had an `active` status, past start date, and NO `last_enabled_at` property.
No run created for trigger via code
- [X] Made sure initial pipeline run was not created when creating a new daily trigger via API POST request, where the trigger had an `active` status, past start date, and NO `last_enabled_at` property.
No run created after POST request

BEFORE - Initial pipeline run unnecessarily created before the current time (now):
BEFORE - Initial pipeline run before current time

AFTER - No initial pipeline run created:
No initial pipeline run

AFTER - Initial pipeline run created if setting enabled
Initial pipeline run created with setting enabled

"Create initial pipeline run" setting when editing a Trigger:
image

# Checklist

cc:
@dy46


mage_ai/orchestration/db/models/schedules.py
```diff
@@ -263,20 +265,32 @@ def create_or_update_batch(self, trigger_configs: List[Trigger]):
                 traceback.print_exc()
                 continue
 
+            last_enabled_at = trigger_config.last_enabled_at
+            if trigger_config.status == ScheduleStatus.ACTIVE and \
+                    trigger_config.last_enabled_at is None:
+                last_enabled_at = datetime.now(tz=pytz.UTC)
+                trigger_config.last_enabled_at = last_enabled_at
+                add_or_update_trigger_for_pipeline_and_persist(
```
johnson-mage, line 273: This was added for the case where a trigger was created via code and that trigger had an `active` status, but no `last_enabled_at` property. This would automatically update the `triggers.yaml` file with a `last_enabled_at` property with the current datetime as its value.

Without this line of code, the `last_enabled_at` property is still updated in the metadata db, but not written to the `triggers.yaml` config file, causing the trigger in the metadata db to be inconsistent with the trigger via code in the `triggers.yaml` file.

docs/guides/triggering-pipelines.mdx
```diff
@@ -62,6 +62,13 @@ payload.
 * (Schedule triggers only) Skip run if previous run still in progress
   * Enabling this setting will tell the scheduler to not initialize the current run if any
     previous run for this trigger is still in progress.
+* (Schedule triggers only) [available starting in v0.9.60] Create initial pipeline run if start date is before current execution period
```
johnson-mage, line 65: This might be changed to v0.9.61 since this feature may not be included in the very next release.

dy46 (commented):
i think you may need to update the code here too: https://github.com/mage-ai/mage-ai/blob/master/mage_ai/orchestration/db/models/schedules.py#L238

mage_ai/orchestration/db/models/schedules.py
```diff
@@ -287,20 +289,32 @@ def create_or_update_batch(self, trigger_configs: List[Trigger]):
...
-                schedule_interval=trigger_config.schedule_interval,
                 status=trigger_config.status,
                 variables=trigger_config.variables,
-                sla=trigger_config.sla,
-                settings=trigger_config.settings,
             )
 
             if existing_trigger:
                 if any([
+                    existing_trigger.last_enabled_at != kwargs.get('last_enabled_at'),
```
johnson-mage, line 317: > i think you may need to update the code here too: https://github.com/mage-ai/mage-ai/blob/master/mage_ai/orchestration/db/models/schedules.py#L238
> 
> ```
> def create_or_update_batch(self, trigger_configs: List[Trigger]):
> ```

@dy46 I think the link to the code line may have changed since it was added, but i think you were referring to the line above. Which part were you referring to specifically for updating the code? Not the parts updated here?

dy46 (approved):

PR #4338, opened 2024-01-10
[jk] Fix caching issues with block cache and shared pipelines

johnson-mage:
# Description
- Fixed bugs related to block cache and shared pipelines displayed for blocks in the Pipeline Editor.

# How Has This Been Tested?
- Tested locally to make sure "Shared pipelines" were being displayed properly in block settings and block cache was getting updated correctly when removing or adding a block to a pipeline.

# Checklist

cc:

### DIFF
--- a/mage_ai/cache/block.py
+++ b/mage_ai/cache/block.py
@@ -150,7 +150,7 @@ class BlockCache(BaseCache):
                 pipeline_dict = cache_dict.get('pipeline') or {}
 
                 return pipeline_dict.get('uuid') != pipeline_uuid and \
-                    pipeline_dict.get('repo_path') != repo_path
+                    pipeline_dict.get('repo_path') == repo_path
 
             pipelines_arr = list(filter(__filter, pipelines_arr))
             pipelines_arr.append(build_pipeline_dict(
@@ -195,7 +195,7 @@ class BlockCache(BaseCache):
         def __filter(cache_dict: Dict, pipeline_uuid=pipeline_uuid, repo_path=repo_path) -> bool:
             pipeline_dict = cache_dict.get('pipeline') or {}
             return pipeline_dict.get('uuid') != pipeline_uuid and \
-                pipeline_dict.get('repo_path') != repo_path
+                pipeline_dict.get('repo_path') == repo_path
 
         pipelines = (mapping.get(key) or {}).get('pipelines', [])
 
--- a/mage_ai/cache/block.py
+++ b/mage_ai/cache/block.py
@@ -51,7 +51,7 @@ class BlockCache(BaseCache):
         if isinstance(block, dict):
             block_type = block.get('type')
             block_uuid = block.get('uuid')
-            block_language = block.get('language', '')
+            block_language = block.get('language')
             configuration = block.get('configuration') or {}
             file_path = os.path.join(
                 repo_path,
--- a/mage_ai/cache/block.py
+++ b/mage_ai/cache/block.py
@@ -8,6 +8,7 @@ import inflection
 from mage_ai.cache.base import BaseCache
 from mage_ai.cache.constants import CACHE_KEY_BLOCKS_TO_PIPELINE_MAPPING
 from mage_ai.cache.utils import build_block_dict, build_pipeline_dict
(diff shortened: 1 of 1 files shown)


mage_ai/cache/block.py
```diff
@@ -150,7 +162,7 @@ def __filter(
                 pipeline_dict = cache_dict.get('pipeline') or {}
 
                 return pipeline_dict.get('uuid') != pipeline_uuid and \
-                    pipeline_dict.get('repo_path') != repo_path
+                    pipeline_dict.get('repo_path') == repo_path
```
johnson-mage, line 165: The `!=` was filtering out all pipelines, resulting in no "shared pipelines" being shown for the blocks.

mage_ai/cache/block.py
```diff
@@ -62,10 +71,13 @@ def build_key(self, block: Union[Dict], repo_path: str = None) -> str:
...
             file_path = block.file_path
 
-        if configuration:
-            file_source = (configuration or {}).get('file_source') or {}
-            if file_source and (file_source or {}).get('path'):
+        file_source = (configuration or {}).get('file_source') or {}
+        if file_source:
+            if (file_source or {}).get('path'):
                 return remove_base_repo_path_or_name((file_source or {}).get('path'))
+        elif block_language:
```
johnson-mage, line 78: This is to make the cache key consistent and always include the file extensions. Previously, some keys included the file extension and some did not.

mage_ai/cache/block.py
```diff
@@ -43,17 +47,22 @@ def build_key(self, block: Union[Dict], repo_path: str = None) -> str:
...
         file_path = None
         repo_path = repo_path or get_repo_path(root_project=False)
 
         if isinstance(block, dict):
             block_type = block.get('type')
             block_uuid = block.get('uuid')
+            block_language = block.get('language')
             configuration = block.get('configuration') or {}
+            block_type_plural = inflection.pluralize(block_type) \
+                if block_type != BlockType.CUSTOM \
```
johnson-mage, line 61: Custom blocks specifically should not be pluralized because the folder containing the custom blocks is `custom`, not `customs`.

mage_ai/cache/block.py
```diff
@@ -62,10 +71,13 @@ def build_key(self, block: Union[Dict], repo_path: str = None) -> str:
             configuration = block.configuration or {}
             file_path = block.file_path
 
-        if configuration:
-            file_source = (configuration or {}).get('file_source') or {}
-            if file_source and (file_source or {}).get('path'):
+        file_source = (configuration or {}).get('file_source') or {}
+        if file_source:
+            if (file_source or {}).get('path'):
                 return remove_base_repo_path_or_name((file_source or {}).get('path'))
```
dy46, line 77: this looks a bit messy and confusing, i think you can just do this:

```suggestion
        if file_source.get('path'):
            return remove_base_repo_path_or_name(file_source.get('path'))
```

mage_ai/cache/block.py
```diff
@@ -62,10 +71,13 @@ def build_key(self, block: Union[Dict], repo_path: str = None) -> str:
...
+        elif block_language:
+            block_file_extension = f'.{BLOCK_LANGUAGE_TO_FILE_EXTENSION[block_language]}'
+            file_path = f'{file_path}{block_file_extension}'
```
dy46, line 80: i don't fully understand what's going on here, but is there a chance the original `file_path` has the file extension already?

johnson-mage: When the block is a dictionary (which when testing was when the block cache is initialized), the `file_path` does not include the file extension, but when the block is a `Block` model, the `file_path` does include the file extension already.

So that's why the `block_language` variable is only reassigned in the `if isinstance(block, dict)` section. I actually initially included the `block_language` in the `else` block, but then the block cache key had a double file extension on the end.

johnson-mage: If the block is a dictionary, I don't think the `file_path` would ever have the file extension since it's using this code to build it and the `block_uuid` doesn't have a file extension: 
```
file_path = os.path.join(
                repo_path,
                block_type_plural,
                block_uuid,
            )
```

dy46 (approved):

PR #4402, opened 2024-01-19
Fixed Oauth connection on Salesforce Source

Luishfs:
# Description

This PR fixes OAuth connection inside Salesforce Source

These problems originated by a timer set on the login method inside SalesforceOAuth connection. This timer would break the 10 seconds rule for Test Connection and block discover time.

Also, the Login URL was modified to match Salesforce Destination connection, and support specific domains instead of the usual login/test param.

Finally, the `__finally_clean_up()` method was removed from the discover phase. This method would only delivery such information outside of the main pipelines logs, and the standard logs from pipeline output already deliver the exact number of requests versus the max cap. This should allow for less discovery time.

# How Has This Been Tested?

Tested using a Salesforce Source account and a PostgreSQL DB as destination.

## FULL TABLE Sync
Screenshot from 2024-01-19 07-01-16

## INCREMENTAL Sync
Screenshot from 2024-01-19 07-01-32

## Data Example
( Account stream example )
Screenshot from 2024-01-19 07-35-17

# Checklist

cc:

@wangxiaoyou1993

### DIFF
--- a/mage_integrations/mage_integrations/sources/salesforce/client/tap_salesforce/salesforce/credentials.py
+++ b/mage_integrations/mage_integrations/sources/salesforce/client/tap_salesforce/salesforce/credentials.py
@@ -31,8 +31,8 @@ def parse_credentials(config):
 
 
 class SalesforceAuth():
-    def __init__(self, credentials, is_sandbox='login'):
-        self.is_sandbox = is_sandbox
+    def __init__(self, credentials, domain='login'):
+        self.domain = domain
         self._credentials = credentials
         self._access_token = None
         self._instance_url = None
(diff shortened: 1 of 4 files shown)


mage_integrations/mage_integrations/sources/salesforce/client/tap_salesforce/salesforce/credentials.py
```diff
@@ -31,7 +31,7 @@ def parse_credentials(config):
 
 
 class SalesforceAuth():
-    def __init__(self, credentials, is_sandbox=False):
+    def __init__(self, credentials, is_sandbox='login'):
```
wangxiaoyou1993, line 34: We can just rename the argument to domain. The `is_sandbox` name is confusing.

wangxiaoyou1993 (approved):

PR #4493, opened 2024-01-30
Fix alter table column names cleaning (postgresql integration exporter)

arnetkachev:
# Description
Properly cleaning column names for existing columns when altering table in postgresql destination mage_integrations.

# How Has This Been Tested?
1. Create data_loader block that returns any uppercase column name DataFrame e.g. `[{"ID": 1,},]`
2. Create data_exporter integration destination block (PostgreSQL). Set option `lower_case: false`
3. Run pipeline. This will create a new table, columns and data. Correctly.
4. Run this pipeline again. As a result the following will happen:
- Without this PR, runner will raise exception: `Error while executing query: column "ID" of relation ... already exists ... Query string: ALTER TABLE "schemaname"."tablename" ADD COLUMN "ID" BIGINT ...`
- With this PR, repeated integration runs are performed correctly.

# Checklist


wangxiaoyou1993 (approved):
Thanks for fixing the issue

PR #4511, opened 2024-02-01
[jk] Reset old block cache if it is older than 2 weeks

johnson-mage:
# Description
- We recently started including file extensions (e.g. `.py`) in the block cache keys to determine which pipelines are currently being used by a block. However, existing Mage projects might have a `blocks_to_pipeline_mapping` cache with existing pipelines and blocks that use the previous cache key convention of NOT including the file extensions.
- This could result in the pipelines using a block (i.e. "shared pipelines") not being displayed properly in the UI if these blocks were already in the existing pipeline before the file extension in cache keys update. For example, a block that is clearly in a pipeline might not even show the current pipeline under the "Pipelines using this block" section in the Block Settings:
missing shared pipeline
- In order to prevent this issue from happening, this PR checks if the block cache file (e.g. `default_repo/.cache/blocks_to_pipeline_mapping`) is older than 2 weeks (by default, this value can be changed), and if it is, we remove the block cache file so it can be replaced with a new cache file that has the updated block cache keys.

# How Has This Been Tested?
- Tested locally:
1) Replaced a block's cache key with one formatted without the file extension so that no shared pipelines would appear in the UI for it
2) Reinitialized the block cache so the existing block cache would be deleted and replaced with a new one.
3) Confirmed that new block cache correctly showed shared pipelines using the block in both the file browser and block settings.

# Checklist

### DIFF
--- a/mage_ai/cache/block.py
+++ b/mage_ai/cache/block.py
@@ -31,9 +31,15 @@ class BlockCache(BaseCache):
         repo_path = repo_path or get_repo_path(root_project=root_project)
         cache = self(repo_path=repo_path)
 
+        # Replace the cache if it is older than 2 weeks (the default if days_ago not passed in).
         cache.remove_old_cache()
 
-        if replace or not cache.exists():
+        """
+        We also check if the cache file does not exist, and if it does not,
+        we create a new cache file, since it is possible for the cache to exist
+        (e.g. as an empty dictionary) but the cache file to not exist at the same time.
+        """
+        if replace or not cache.exists() or not cache.cache_file_exists():
             await cache.initialize_cache_for_all_pipelines(caches=caches, file_path=file_path)
 
         return cache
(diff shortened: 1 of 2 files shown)


mage_ai/cache/block.py
```diff
@@ -30,7 +30,16 @@ async def initialize_cache(
     ) -> 'BlockCache':
         repo_path = repo_path or get_repo_path(root_project=root_project)
         cache = self(repo_path=repo_path)
-        if replace or not cache.exists():
+
+        # Replace the cache if it is older than 2 weeks (the default if days_ago not passed in).
+        cache.remove_old_cache()
```
wangxiaoyou1993, line 35: Can you only reset cache when replace is True?

johnson-mage: updated

johnson-mage:
Closing this PR as it is no longer needed.

PR #4629, opened 2024-02-23
[dy] Add DB locking

dy46:
# Description

Add DB locking for postgres and mysql. Mage doesn't work properly with mysql yet, but I added it in case we support it in the future. Each instance of mage will attempt to acquire the lock for 30 seconds after starting, and if it cannot acquire it an error will be thrown.

The code for this is based off of airflow's db locking implementation.

# How Has This Been Tested?

- [x] Tested locally with separate webserver and scheduler using postgres and mysql db
Screenshot 2024-02-22 at 3 59 57 PM

# Checklist

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/orchestration/db/database_manager.py
+++ b/mage_ai/orchestration/db/database_manager.py
@@ -1,10 +1,75 @@
+import contextlib
+import enum
 import logging
 import os
+from typing import Generator
 
 from alembic import command
 from alembic.config import Config
+from sqlalchemy import text
 
-from mage_ai.orchestration.db import db_connection_url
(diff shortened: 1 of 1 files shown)


wangxiaoyou1993:
One customer's deadlock is like this 
image
When Mage runs DB migration on pipeline schedule, the pipeline_schedules table is locked.
Can this PR solve this problem?

dy46:
why is the `pipeline_schedules` table locked when the DB migration is ran?

dy46:
this should help if there are issues because multiple pods are running migrations since only 1 instance can hold the lock at any time. i'm not sure why there's a specific issue with the pipeline_schedule table though

PR #4687, opened 2024-03-01
[dy] Fix elasticsearch destination

dy46:
# Description

Fix a couple of issues with the Elasticsearch destination:
* Test connection will fail if the `properties` field doesn't exist, so adding a safe get there.
* The `logger.debug` command was causing an error when using the `metadata_fields` config value.

# How Has This Been Tested?

- [x] Tested with local elasticsearch cluster

# Checklist

cc: @wangxiaoyou1993


mage_integrations/mage_integrations/destinations/elasticsearch/__init__.py
```diff
@@ -13,18 +13,18 @@
...
-                             stream_name='test',
-                             schema={},
-                             key_properties=None,).client
+        target = TargetElasticsearch(config=self.config, logger=self.logger)
+        client = ElasticSink(
+            target=target,
+            stream_name='test',
+            schema={},
+            key_properties=None,
+        ).client
```
dy46, line 27: these are just formatting changes

Issue #4681, opened 2024-02-29
[BUG] Mage.ai doesn't work when sql server used for MAGE_DATABASE_CONNECTION_URL


tuokorTW:
### Mage version

v0.9.65

### Describe the bug

If I set MAGE_DATABASE_CONNECTION_URL to use sql server as per documentation here https://docs.mage.ai/guides/docker/connecting-a-database 
On startup mage will throw errors with stacktrace:
```
magic-1         | INFO:alembic.runtime.migration:Running upgrade dfe49d040487 -> 90d978a8aef8, Update unique constraint for secret
magic-1         | INFO:sqlalchemy.engine.Engine:ROLLBACK
magic-1         | Traceback (most recent call last):
magic-1         |   File "/usr/local/lib/python3.10/site-packages/mage_ai/server/scheduler_manager.py", line 40, in run_scheduler
magic-1         |     database_manager.run_migrations()
magic-1         |   File "/usr/local/lib/python3.10/site-packages/mage_ai/orchestration/db/database_manager.py", line 114, in run_migrations
magic-1         |     command.upgrade(alembic_cfg, revision='head')
magic-1         |   File "/usr/local/lib/python3.10/site-packages/alembic/command.py", line 403, in upgrade
magic-1         |     script.run_env()
magic-1         |   File "/usr/local/lib/python3.10/site-packages/alembic/script/base.py", line 583, in run_env
magic-1         |     util.load_python_file(self.dir, "env.py")
magic-1         |   File "/usr/local/lib/python3.10/site-packages/alembic/util/pyfiles.py", line 95, in load_python_file
magic-1         |     module = load_module_py(module_id, path)
magic-1         |   File "/usr/local/lib/python3.10/site-packages/alembic/util/pyfiles.py", line 113, in load_module_py
magic-1         |     spec.loader.exec_module(module)  # type: ignore
magic-1         |   File "<frozen importlib._bootstrap_external>", line 883, in exec_module
magic-1         |   File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
magic-1         |   File "/usr/local/lib/python3.10/site-packages/mage_ai/orchestration/db/migrations/env.py", line 103, in <module>
magic-1         |     run_migrations_online()
magic-1         |   File "/usr/local/lib/python3.10/site-packages/mage_ai/orchestration/db/migrations/env.py", line 97, in run_migrations_online
magic-1         |     context.run_migrations()
[...]

PR #4695, opened 2024-03-01
Bugfix: Fix migration crash when using MSSQL as mage database

tuokor:
# Description
This is a fix for issue https://github.com/mage-ai/mage-ai/issues/4681 where migration file 90d978a8aef8_update_unique_constraint_for_secret.py crashed due to alembic not supporting constraint retrieval for MSSQL as constraint are implemented as indexes that need to fetched using get_indexes() and dropped via alter table sql command.

# How Has This Been Tested?
I verified the changes by setting mage.ai to use sql server as the database and also tested with postgres to make sure changes didn't cause side effects to existing migration code.

- [x] Migration works with ms sql server 
- [x] Migration still works with postgres

# Checklist
- [x] If new documentation has been added, relative paths have been added to the appropriate section of `docs/mint.json`

cc:
@dy46


mage_ai/orchestration/db/migrations/versions/90d978a8aef8_update_unique_constraint_for_secret.py
```diff
@@ -20,11 +20,17 @@ def upgrade() -> None:
     # ### commands auto generated by Alembic - please adjust! ###
     try:
         bind = op.get_bind()
-        insp = sa.inspect(bind.engine)
-        with op.batch_alter_table('secret', schema=None) as batch_op:
-            unique_constraints = insp.get_unique_constraints('secret')
-            for constraint in unique_constraints:
-                batch_op.drop_constraint(constraint['name'], type_='unique')
+        insp = sa.inspect(bind)
```
dy46, line 23: Changing this line affects the `insp` in the `else` block as well.

tuokor: Yes, but it did not affect the results when I tested with different sql servers. 

SQL server inspection did not work if I had the original line `sa.inspect(bind.engine)`. All inspections calls freeze the migration if bind.engine is used with sql server. I read the documentation the inspect method does accept either an engine or a connection which the bind variable is https://docs.sqlalchemy.org/en/20/core/inspection.html#available-inspection-targets .

tuokor: Here is also documentation about the get_bind() function that states that it returns a connection when "online" mode https://alembic.sqlalchemy.org/en/latest/api/runtime.html#alembic.runtime.environment.EnvironmentContext.get_bind

dy46 (approved):

PR #4758, opened 2024-03-14
[jk] Limit number of block output files read and included in pipeline response

johnson-mage:
# Description
- If a user's block returned a list of many values (versus a list of a list with many values), it could cause significant slowdowns in the browser when loading the pipeline because a block output file is created for each value in the output list. For example, a block returning a list of 1000 items creates 1000 output files, which takes a long time to read so many files and resulting in an unintended large pipeline response.
- This PR limits the number of output files' contents included in the pipeline response to 10 files (previously all output files' contents were included).

# How Has This Been Tested?
- Tested using the same block that returned a list with 20k items
- [x] Tested with local variable manager (local storage)
- [x] Tested with AWS S3 variable manager (amazon s3 storage)
- [x] Tested with GCS variable manager (google cloud storage)

Before (took ~34 seconds to load the pipeline/page):
image

After (took <1 second to load the pipeline/page):
image

# Checklist

### DIFF
--- a/mage_ai/services/aws/s3/s3.py
+++ b/mage_ai/services/aws/s3/s3.py
@@ -45,11 +45,17 @@ class Client:
             },
         )
 
-    def listdir(self, prefix: str, delimiter: str = '/', suffix: str = None):
+    def listdir(
+        self,
+        prefix: str,
+        delimiter: str = '/',
+        suffix: str = None,
+        max_results: int = None,
+    ):
         keys = []
         response = self.client.list_objects_v2(
             Bucket=self.bucket,
-            MaxKeys=MAX_KEYS,
+            MaxKeys=max_results or MAX_KEYS,
             Prefix=prefix,
             Delimiter=delimiter,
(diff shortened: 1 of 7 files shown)


mage_ai/services/aws/s3/s3.py
```diff
@@ -45,11 +45,17 @@ def delete_objects(self, prefix: str):
...
+        prefix: str,
+        delimiter: str = '/',
+        suffix: str = None,
+        max_results: int = None,
+    ):
         keys = []
         response = self.client.list_objects_v2(
             Bucket=self.bucket,
-            MaxKeys=MAX_KEYS,
+            MaxKeys=max_results or MAX_KEYS,
```
johnson-mage, line 58: Per AWS [docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/list_objects_v2.html)

dy46 (approved):

PR #4761, opened 2024-03-14
[dy] Record memory and cpu usage for pipeline runs

dy46:
# Description

This PR adds memory and cpu tracking for pipeline runs in the backend. There will be a follow up PR to surface the metrics in the frontend.

Calculating memory is relatively straightforward, we can get the memory usage for all processes for a pipeline run. To get the memory usage per pipeline, we can use `psutil.Process.memory_percent()`. CPU can be done in a similar way, but getting the cpu usage percentage per process is more complicated. The comments in the code explains more in depth how it's implemented, but basically we need to track the cpu usage at various points in time and compare the numbers to get the cpu usage percent. 

Right now, we get memory and cpu usage every time a heartbeat is run for the pipeline run, so roughly every 10 seconds. The usage is stored in the `PipelineRun.metrics` field. The pipeline run cpu and memory usage will also be included in the tags for the heartbeat log.

Example:
Screenshot 2024-03-14 at 3 29 37 PM

Screenshot 2024-03-14 at 3 41 28 PM

# How Has This Been Tested?

- [x] Tested locally with various pipelines

# Checklist

cc: @wangxiaoyou1993

### DIFF
--- a/mage_ai/orchestration/pipeline_scheduler_original.py
+++ b/mage_ai/orchestration/pipeline_scheduler_original.py
@@ -1,10 +1,12 @@
 import asyncio
 import collections
 import os
+import time
 import traceback
 from datetime import datetime, timedelta
 from typing import Any, Dict, Generator, List, Set, Tuple
 
+import psutil
 import pytz
 from dateutil.relativedelta import relativedelta
 from sqlalchemy import desc, func
@@ -120,8 +122,6 @@ class PipelineScheduler:
             self.pipeline_schedule.get_settings().allow_blocks_to_fail
             if self.pipeline_schedule else False
         )
-
-        self.processes_to_track = dict()
 
     @safe_db_query
     def start(self, should_schedule: bool = True) -> bool:
@@ -841,9 +841,6 @@ class PipelineScheduler:
         return crashed_runs
 
     def __run_heartbeat(self) -> None:
-        import time
-
-        import psutil
         load1, load5, load15, cpu_count = get_compute()
         cpu_count = cpu_count or 1
         cpu_usage = load15 / cpu_count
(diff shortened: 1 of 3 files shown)


mage_ai/orchestration/pipeline_scheduler_original.py
```diff
@@ -854,14 +857,124 @@ def __run_heartbeat(self) -> None:
...
+
         self.logger.info(
             f'Pipeline {self.pipeline.uuid} for run {self.pipeline_run.id} '
             f'in schedule {self.pipeline_run.pipeline_schedule_id} is alive.',
             **tags,
         )
 
-        if memory_usage and memory_usage >= MEMORY_USAGE_MAXIMUM:
-            self.memory_usage_failure(tags=tags)
+    def __measure_and_record_usage(self) -> Tuple[float, float]:
```
wangxiaoyou1993, line 876: This will work with local_python executor but not other executors.

mage_ai/orchestration/pipeline_scheduler_original.py
```diff
@@ -854,14 +857,124 @@ def __run_heartbeat(self) -> None:
...
+        Returns:
+            Tuple[float, float]: A tuple of pipeline run cpu and memory usage percent.
+        """
+        # Get all job_ids for the pipeline run so that we can get the pids for the jobs
+        job_ids = []
+        if self.pipeline and (
+            self.pipeline.type in [PipelineType.INTEGRATION, PipelineType.STREAMING]
+            or self.pipeline.run_pipeline_in_one_process
+        ):
+            if job_manager.has_pipeline_run_job(self.pipeline_run.id):
```
wangxiaoyou1993, line 889: this will not work when you have multiple schedulers

PR #4812, opened 2024-03-22
[jk] Avoid N dbt config option requests for N blocks on Pipeline Editor

johnson-mage:
# Description
- The Pipeline Editor page was making a `configuration_options` request for each dbt block in the pipeline upon initial load (e.g. 50 dbt blocks would make 50 requests). This PR mitigates the number of requests made by making a single `configuration_options` request upon initial load.

# How Has This Been Tested?
Load pipeline with dbt blocks (2 sql and 1 yaml):
load dbt pipeline

# Checklist

cc:
@wangxiaoyou1993

### DIFF
--- a/mage_ai/frontend/components/CodeBlock/index.tsx
+++ b/mage_ai/frontend/components/CodeBlock/index.tsx
@@ -217,6 +217,7 @@ type CodeBlockProps = {
   cursorHeight2?: number;
   cursorHeight3?: number;
   dataProviders?: DataProviderType[];
+  dbtConfigurationOptions?: ConfigurationOptionType[]
   defaultValue?: string;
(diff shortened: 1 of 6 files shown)


wangxiaoyou1993:
do we need the initial request without block_uuid? i guess we only need to make the request when the block is selected.

wangxiaoyou1993:
also want to make sure, we don't send the request again when the block is selected multiple times, right?

johnson-mage:
> do we need the initial request without block_uuid? i guess we only need to make the request when the block is selected.

Not necessarily, if the dbt blocks were all yaml blocks, then the page would only make that initial request. But I made a change so the request is only made when a block is selected.

PR #4841, opened 2024-03-27
[jk] Bugfix pipeline tag cache

johnson-mage:
# Description
- Remove pipeline tag from filter dropdown on Pipelines Dashboard if that tag is no longer used by any pipelines.
- Update tag cache when removing tags from a pipeline (previously the tag cache was not getting updated in this scenario).

# How Has This Been Tested?
Tag is removed from pipeline filters after the only pipeline using it is deleted:
unused tag

The pipelines dashboard reflects a pipeline's updated tags after removing one of its tags:
pipeline tags updated when removed

# Checklist

### DIFF
--- a/mage_ai/cache/tag.py
+++ b/mage_ai/cache/tag.py
@@ -144,6 +144,9 @@ class TagCache(BaseCache):
 
         mapping[key][KEY_FOR_PIPELINES] = pipelines_dict
 
+        if not pipelines_dict and mapping[key].keys() == set([KEY_FOR_PIPELINES]):
+            mapping.pop(key, None)
+
         self.set(self.cache_key, mapping)
 
     async def initialize_cache_for_all_objects(self) -> None:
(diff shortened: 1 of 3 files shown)


mage_ai/cache/tag.py
```diff
@@ -144,6 +144,9 @@ def remove_pipeline(self, tag_uuid: str, pipeline_uuid: str, repo_path: str = No
 
         mapping[key][KEY_FOR_PIPELINES] = pipelines_dict
 
+        if not pipelines_dict and mapping[key].keys() == set([KEY_FOR_PIPELINES]):
```
johnson-mage, line 147: This checks if no pipelines are still using the tag, and if there aren't any, it'll purge it from the cache (so it doesn't still appear in the pipeline filters dropdown).

dy46: this comparison does look weird to me, but i can't think of a better way to do it right now

dy46 (approved):

Issue #4898, opened 2024-04-08
OIDC Auth does not support user_roles


mihaivint:
### Mage version

0.9.68 or newer

### Describe the bug

OIDC auth only returns email, username and ignores user_roles if provided see mage_ai/authentication/providers/oidc.py at the end

wangxiaoyou1993:
This is supported in this PR https://github.com/mage-ai/mage-ai/pull/5053 (version 0.9.71)

PR #4899, opened 2024-04-08
Add support for user_roles when using OIDC auth

mihaivint:
OIDC auth ignores user_roles , this pr fixes reported bug #4898 

# Checklist
- [ x] The PR is tagged with proper labels (bug, enhancement, feature, documentation)
- [ x] I have performed a self-review of my own code

cc:

### DIFF
--- a/mage_ai/authentication/providers/oidc.py
+++ b/mage_ai/authentication/providers/oidc.py
@@ -131,4 +131,5 @@ class OidcProvider(OauthProvider, SsoProvider):
         return dict(
             email=email,
             username=userinfo_resp.get('preferred_username', email),
+            user_roles = userinfo_resp.get('user_roles'),
         )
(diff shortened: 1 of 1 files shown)


dy46:
what does the `user_roles` field look like in your use case?

mihaivint:
I actually tested against Vault, But the user_roles should be a list 
SCOPE_TEMPLATE='{

    "username": {{identity.entity.name}},

    "email": {{identity.entity.metadata.email}},

    "user_roles": ["Admin","Editor","Owner","Viewer"]

}'

TheGthr:
One would need to implement for every OIDC-based provider no? And is this field used to build the user?

mihaivint:
Actually no as i've set it to default to empty current behaviour. So if you don't define it, nothing is changed, meaning after you login you must assign as admin a role to that user.

dy46 (approved):

PR #4926, opened 2024-04-11
[jk] Add "Last run failed" filter to Pipeline Runs table

johnson-mage:
# Description
- Add a "Last run failed" filter to the Pipeline Runs Tables on pages where the filter is a dropdown (not where the filter is a toggle menu and multiple filters can be applied, such as on the project-wide Pipeline Runs page). The "Last run failed" filter displays failed pipeline runs that are the last pipeline run retry (or runs without any retries). This can be helpful if the user does not want to see failed pipeline runs that have already been retried successfully.

# How Has This Been Tested?
- [ ] Added unit test
- [ ] Tested locally (see gif below):

# Checklist

cc:

### DIFF
--- a/mage_ai/api/resources/PipelineRunResource.py
+++ b/mage_ai/api/resources/PipelineRunResource.py
@@ -1,3 +1,4 @@
+from sqlalchemy import and_, desc, func
 from sqlalchemy.orm import selectinload
 
 from mage_ai.api.operations.constants import META_KEY_LIMIT, META_KEY_OFFSET
@@ -5,7 +6,10 @@ from mage_ai.api.resources.DatabaseResource import DatabaseResource
 from mage_ai.api.utils import get_query_timestamps
 from mage_ai.cache.tag import TagCache
 from mage_ai.data_preparation.models.block.utils import get_all_descendants
-from mage_ai.data_preparation.models.constants import PipelineType
+from mage_ai.data_preparation.models.constants import (
+    PIPELINE_RUN_STATUS_LAST_RUN_FAILED,
+    PipelineType,
+)
 from mage_ai.data_preparation.models.pipeline import Pipeline
 from mage_ai.data_preparation.models.triggers import (
     ScheduleInterval,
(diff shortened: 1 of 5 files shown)


mage_ai/api/resources/PipelineRunResource.py
```diff
@@ -94,9 +98,38 @@ async def collection(self, query_arg, meta, user, **kwargs):
...
+                        order_by=desc(PipelineRun.id))
+                    .label('row_number')
+            ).cte(name='latest_pipeline_runs')
+            query = (PipelineRun.select(
+                    PipelineRun,
+                )
+                .join(latest_pipeline_runs, and_(
+                    PipelineRun.id == latest_pipeline_runs.c.id,
+                    latest_pipeline_runs.c.row_number == 1,
+                ))
```
dy46, line 126: what's the point of this join?

johnson-mage: @dy46 It joins the `latest_pipeline_runs` cte, which has the pipeline_run ids and row numbers to determine the latest pipeline run by `execution_date` (`pipeline_schedule_id` is used to sort the runs because there can be runs with the exact same `created_at` dates, which I tried initially), with the regular PipelineRun table and also filters out any rows that do NOT have row_number of `1`. The rows with row_number of 1 are the latest pipeline run.

Maybe a little confusing combining the filter in the join. Another way of writing the query, would this be clearer?
```
query = (PipelineRun.select(
                    PipelineRun,
                )
                # The `and_` is needed even though there's only 1 condition or the query doesn't work properly.
                .join(latest_pipeline_runs, and_(PipelineRun.id == latest_pipeline_runs.c.id))
                .filter(latest_pipeline_runs.c.row_number == 1)
            )
```

dy46: ah yeah i get it - i forgot you have to get the actual pipeline run object. i think the original way you did it is ok

dy46 (approved):

PR #4927, opened 2024-04-11
Update integrate-mage-airflow.mdx

Wesseldr:
The command did not work with docker, think this fixed it.

# Description
Example did not work.

# How Has This Been Tested?
running the command did create the test project inside the dag folder


wangxiaoyou1993 (approved):
Thanks for fixing the doc!

PR #5110, opened 2024-05-24
Update utils.py

hanslemm:
# Description
This PR proposes a better column type for OBJECT column types.

The original VARCHAR(65535) is very limited in the volume of data that it can handle (big texts/JSON won't be able to be ingested, creating a problem for ingestion). SUPER has a more generous data volume.

cc:
@wangxiaoyou1993


hanslemm:
@wangxiaoyou1993 simple change that can improve every Redshift user life.

wangxiaoyou1993:
@hanslemm did you test this change?

Issue #5360, opened 2024-08-23
[BUG] OAUTH API Key Not working -  Main Project > Workspaces setup


Arthidon:
### Mage version

9.73

### Describe the bug

API calls in "test" environment for workspaces are receiving OAUTH error.
Tested endpoints "api/syncs" and api/files
We can successfully get the token in the authentication call
This is happening in the pipeline and with postman
there are no issues in the "develop" environment

### Screenshots

ENV variables for working  Workspaces in Develop
![image](https://github.com/user-attachments/assets/78408c62-cbfc-43b2-a4ba-07e25f78ef8e)

ENV variables for non working  Workspaces in Test
![image](https://github.com/user-attachments/assets/3fdec5e0-63c8-4a71-8509-f6f75e25b9e4)

Api Error:
![image](https://github.com/user-attachments/assets/ff1951a3-1514-4c6e-9bba-c7350029ce4b)

### Additional context

Current Setup:
Gitlab CI/CD pipeline using Teraform / Ansible / Helm / EKS 
We have a Main Project on 9.71 to create workspaces
We have 2 Environments, "develop"and "test". We will be deploying Prod soon, this issue is a blocker at the moment.
We recently starting upgrading workspaces to 9.73
We use API calls to configure our workspaces  in our gitlab pipelines

Arthidon:
It looks like you are using the test environment variable to run the local database , which makes it so it can't talk to the RDS database
Source code where we think issue is: https://github.com/mage-ai/mage-ai/blob/master/mage_ai/orchestration/db/__init__.py
To reproduce

johnson-mage:
Added a fix for this. Will be in the next version. A temporary workaround for now if your test environment needs to be used would be to change the `ENV` value in your test environment to something other than `test` (also avoid using `test_mage` since that's what we're using for the environment to run our unit tests).

PR #5372, opened 2024-08-29
[jk] Update test ENV value

johnson-mage:
# Description
- Using `test` as the value for the `ENV` environment variable for running unit tests could cause the test db to conflict with the metadata db used by users in their own test environments. This PR updates the value used for `ENV` in the environment for running mage's unit tests.

# How Has This Been Tested?
Before:
- Using `test` for the `ENV` env var resulted in using the `test.db` file as the metadata db, even if postgres was set as the metadata db using the `MAGE_DATABASE_CONNECTION_URL` env var.

After:
Using `test` for the `ENV` env var did not result in using the `test.db` file as the metadata db. If postgres was set as the metadata db using the `MAGE_DATABASE_CONNECTION_URL` env var, postgres would be properly used as the metadata db. Confirmed by creating new users and verifying they were stored in the postgres db.

# Checklist

cc:

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -16,6 +16,9 @@ on:
       - "requirements.txt"
       - "setup.py"
 
+env:
+  ENV: test_mage
+
 jobs:
   check-code-quality:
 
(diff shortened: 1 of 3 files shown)


johnson-mage:
@wangxiaoyou1993 This is a simple fix to prevent this [github issue](https://github.com/mage-ai/mage-ai/issues/5360). The unit tests ran fine after updating the `ENV` test value to be more unique. This shouldn't have any side effects on the app afaik, but can you please take a look to see if I might've missed something?

wangxiaoyou1993 (approved):

PR #5393, opened 2024-09-07
Enable build and test for Python 3.11 and 3.12

csharplus:
# Description

The build and test for Python 3.12 was skipped so far for Mage, which has been enabled.

# How Has This Been Tested?

- [X] Will check to see if there are any issues through GitHub's build and test workflow.

# Checklist

cc:

@wangxiaoyou1993


setup.py
```diff
@@ -117,9 +122,12 @@ def readme():
             'confluent-avro~=1.8.0',
             'elasticsearch==8.9.0',
             'influxdb_client==1.36.1',
-            'kafka-python==2.0.2',
+            "kafka-python==2.0.2; python_version < '3.12'",
+            "kafka-python @ git+https://github.com/dpkp/kafka-python.git ; "
```
wangxiaoyou1993, line 126: we can't use library from github repo in pypi package

csharplus: It is an issue of `kafka-python` tracked at https://github.com/dpkp/kafka-python/issues/2412, without a good solution so far.

csharplus: Found another work-around about this issue, given in https://stackoverflow.com/questions/77287622/modulenotfounderror-no-module-named-kafka-vendor-six-moves-in-dockerized-djan

wangxiaoyou1993 (approved):

Issue #5417, opened 2024-09-14
[BUG] Can't use git with Bitbucket


kennynguyeenx:
### Mage version

0.9.73

### Describe the bug

After authentication with Bitbucket successfully, I tried to fetch code from Bitbucket and got this error
`Cmd('git') failed due to: exit code(128)
  cmdline: git fetch -v --progress -- test
  stderr: 'fatal: Authentication failed for 'https://bitbucket.org/finalto360/mage.git/''`

I c

### To reproduce

1. Set up Oauth with Bitbucket using this instruction https://docs.mage.ai/production/data-sync/version-control#authenticate-with-bitbucket
2. Authenticate with Bitbucket successfully
3. Go to Remotes section and add one more remote repo using repo in Bitbucket. For example: https://bitbucket.org/mage-ai/mage-ai.git
4. Go to Actions section and choose the remote repo that I have just created in Step 3 and click fetch
5. Got authentication error

### Expected behavior

Can fetch code from the remote repo in Bitbucket successfully

### Screenshots

Authenticate successfully with Bitbucket
![Screenshot from 2024-09-14 11-17-59](https://github.com/user-attachments/assets/883a25de-2b34-4c4f-979f-428344a233fa)

Add the remote repo from Bitbucket

![Screenshot from 2024-09-14 11-18-06](https://github.com/user-attachments/assets/c562554d-da26-4e78-bb1b-f29394f75926)
Try to fetch code from the remote repo in Bitbucket and got error

![Screenshot from 2024-09-14 11-18-13](https://github.com/user-attachments/assets/2137bdb1-3821-4871-9564-586bb3784abd)

### Operating system

- OS: Amazon LInux

### Additional context

I checked documentation from Bitbucket and found this note https://support.atlassian.com/bitbucket-cloud/docs/use-oauth-on-bitbucket-cloud/
![Screenshot from 2024-09-14 11-05-54](https://github.com/user-attachments/assets/23eb559b-27e1-45b1-9eab-0410c0d55b19)
It means that we have to use x-token-auth instead of username
But When I go to the code of mage-ai, I see we still use username like the other providers in function build_authenticated_remote_url‎
https://github.com/mage-ai/mage-ai/blob/0.9.73/mage_ai/data_preparation/git/utils.py#L172

wangxiaoyou1993:
This is published in 0.9.74

PR #5418, opened 2024-09-14
#5417 Fix bug: Can't use git with Bitbucket

kennynguyeenx:
# Description
Based on this documentation from Bitbucket, there is a difference between Bitbucket and the other providers like Github in using auth-token

> The literal string x-token-auth as a substitute for username is required.
Screenshot (86)

So I changed a little bit code in function build_authenticated_remote_url‎ to customize for bitbucket only.
I don't want to change too much in the other function because it should be isolated for only this function, the other functions are the same with the other providers
# How Has This Been Tested?
Screenshot (84)

- Tested this code by updating code directly in lib file: /usr/local/lib/python3.10/site-packages/mage_ai/data_preparation/git/utils.py and it worked

Screenshot (85)

# Checklist

cc:

### DIFF
--- a/mage_ai/data_preparation/git/utils.py
+++ b/mage_ai/data_preparation/git/utils.py
@@ -172,7 +172,10 @@ def get_access_token(git_config, repo_path: str) -> str:
 def build_authenticated_remote_url(remote_url: str, username: str, token: str) -> str:
     # https://[username]:[token]@github.com/[remote_url]
     url = urlsplit(remote_url)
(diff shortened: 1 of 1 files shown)


wangxiaoyou1993:
I guess for bitbucket cloud, you can use `x-token-auth:{token}`
and for bitbucket server, you will use `{username}:{token}`

wangxiaoyou1993 (approved):

PR #5485, opened 2024-10-09
[jk] Allow user to set number of lines displayed for block output sample preview

johnson-mage:
# Description
- The block output sample preview in the Pipeline Editor defaults to 10 lines, but users may want to decrease this amount to improve browser performance, especially if they have many blocks in their pipeline or blocks that returns dataframes with many columns / larges values in those columns.
- The environment variable `DATAFRAME_SAMPLE_COUNT_PREVIEW` is used to set the number of lines for the block output sample preview.
- Make the "Disable output preview" setting wording clearer. It says "Disable output preview on page load" now since the block's sample output is still visible after running a block, but users might think that the output is disabled at all times. This setting prevents the block sample output from appearing when the page initially loads. Also added `disable_output_preview` as a reserved key so it does not appear in the "Block variables" section of the block settings, since it is part of a block's config.
image

# How Has This Been Tested?
- Confirmed that the number of rows in the block sample output matches the value used for the env var `DATAFRAME_SAMPLE_COUNT_PREVIEW`
- Value of `5` used below:
image

# Checklist

cc:


docs/development/variables/environment-variables.mdx
```diff
@@ -49,50 +49,51 @@ the `MAGE_REQUESTS_BASE_PATH` and `MAGE_ROUTES_BASE_PATH` environment variables
...
+| -------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------ |
+| `ACTIVE_DIRECTORY_DIRECTORY_ID`  | Set this value to your organization's tenant id to enable signing in with Microsoft account.                                                                    | [More info](/production/authentication/microsoft)                        |
+| `AUTHENTICATION_MODE`            | Set the authentication mode in Mage. Omitting the variable will default to basic password authentication.                                                       | `ldap`                                                                   |
+| `DATAFRAME_SAMPLE_COUNT_PREVIEW` | Number of lines to display in block sample output preview for dataframes. Not suggested to go over 10 lines (defaults to 10).                                   | 5                                                                        |
```
johnson-mage, line 56: Added this line to the docs to describe the `DATAFRAME_SAMPLE_COUNT_PREVIEW` env var. There were no other changes to the docs. The rest of the diff in this file was just formatting.

wangxiaoyou1993 (approved):

PR #5499, opened 2024-10-14
Add time delay to avoid getting the same file timestamp for multiple changes

csharplus:
# Description

The root cause of some flaky unit tests was found: [the cached reading of YAML file](https://github.com/mage-ai/mage-ai/blob/master/mage_ai/cache/mem_lru.py) could not detect file changes if two changes are too close together (e.g., calling set_global_variable twice in a row as done in unit tests). This issue was explained in more detail in: https://stackoverflow.com/questions/19059877/python-os-path-getmtime-time-not-changing.

It could be an issue with Mage system overall, i.e., whenever pipeline state changes are too close to each other, the cached version of the pipeline YAML file in memory would be stale, and could lead to data loss due to wrong overwrites, e.g., the earlier set value is over-written.

I've tried the time delay value of `0.0001` and `0.0002`, which didn't work. The final value chosen was after some trial-and-errors.

# How Has This Been Tested?

- [x] Tested it in a local development environment

# Checklist

cc:

@wangxiaoyou1993

### DIFF
--- a/mage_ai/data_preparation/models/pipeline.py
+++ b/mage_ai/data_preparation/models/pipeline.py
@@ -2329,6 +2329,10 @@ class Pipeline:
         extension_uuid: str = None,
         widget: bool = False,
     ):
+        # Introduce a small delay to prevent multiple changes from generating
+        # identical timestamps for the pipeline YAML file
+        time.sleep(0.001)
+
         blocks_current = sorted([b.uuid for b in self.blocks_by_uuid.values()])
 
         if block_uuid is not None:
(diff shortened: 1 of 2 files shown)


mage_ai/data_preparation/models/pipeline.py
```diff
@@ -2328,6 +2329,10 @@ def save(
         extension_uuid: str = None,
         widget: bool = False,
     ):
+        # Introduce a small delay to prevent multiple changes from generating
+        # identical timestamps for the pipeline YAML file
+        time.sleep(0.0003)
```
wangxiaoyou1993, line 2334: is this just the fix to pass unit tests?

csharplus: This could be a significant issue with Mage production pipelines. When two state changes (e.g., setting two global variables in consecutive statements) happen in quick succession, the generated YAML file might receive the same timestamp. Consequently, the online cache retains only the earlier version of the changes, causing data loss as subsequent writes overwrite each other.

wangxiaoyou1993 (approved):

PR #5516, opened 2024-10-22
BUG MongoDB config.json needs username instead of user

Wesseldr:
When the template is presented in a new dataintegration pipeline it uses the wrong parameter for the username.

When using user, login is not possible. using username it works

# Description
I noticed login does not work on MongoDB, using username instead of user made the difference.

# How Has This Been Tested?
Create new dataintegration pipeline and setup mongodb as a source.

# Checklist


Wesseldr:
Not sure what it is, authentication that works in a normal pipeline seems to fail in a dataintegration pipeline. Closing this issue not sure if this is the problem... Sorry!

Issue #5427, opened 2024-09-18
[BUG] Data files not being properly passed from block to block in R "Cannot open file" 


gpierard:
### Mage version

0.9.73

### Describe the bug

after running an initial R data loader block I get this error in the following transformer (in R) cannot open file '/home/*********/.mage_data/
This has been reported by others on slack at https://mageai.slack.com/archives/C03HTTWFEKE/p1719534700823339, and the bug is still present in the latest version.

### To reproduce

1. Create a data loader block in R outputting any data
2. create a downstream R block and try to run

### Operating system

Distributor ID: Ubuntu
Description:    Ubuntu 20.04.6 LTS
Release:        20.04
Codename:       focal

noahsalvi:
Duplicate of #5336

**Some additional remarks:**
We're facing the same problem and found that this wasn't an issue in version 0.9.70 for the following setup:
(Python block outputting Pandas DataFrame) -> (R transformer)

However, the following setup using Polars still produced the same error as here in that version.
(Python block outputting **Polars** DataFrame) -> (R transformer).

csharplus:
This issue has been resolved by https://github.com/mage-ai/mage-ai/pull/5606.

Issue #5336, opened 2024-08-12
[BUG] Connecting two R Blocks fails with 'No such file or directory' error


mkabachia:
### Mage version

0.9.73

### Describe the bug

When connecting two R blocks, the loader saves data in /home/src/mage_data/............../sample_data/output_0/data.csv but the second(transformer) block tries to read /home/src/mage_data/............../sample_data/data.csv leading to 'No such file or directory' error.

### To reproduce

1. Create a data loader using R language template. 
2. Connect it an R language transformer. 

### Expected behavior

The transformer should access the data from the loader

### Screenshots

![mage_ai_R_bug](https://github.com/user-attachments/assets/1b5162e2-c234-4d49-8f0c-f25522e7511f)

### Operating system

Docker  - mageai/mageai:latest

csharplus:
This issue has been resolved by https://github.com/mage-ai/mage-ai/pull/5606.

PR #5606, opened 2024-12-08
[hw] Resolve the input path issue with R transformer blocks

csharplus:
# Description

Address the file reading issue with R transformer blocks reported in https://github.com/mage-ai/mage-ai/issues/5336 and https://github.com/mage-ai/mage-ai/issues/5427. The fix has been tested with a data pipeline of three R blocks: `R data loader + R Transformer + R Transformer`, as shown in the screenshot below:

Screenshot 2024-12-28 at 11 51 49 PM

# How Has This Been Tested?

- [x] Tested with a local development environment. 

# Checklist

cc:

@wangxiaoyou1993

### DIFF
--- a/mage_ai/data_preparation/models/block/r/__init__.py
+++ b/mage_ai/data_preparation/models/block/r/__init__.py
@@ -8,8 +8,11 @@ import pandas as pd
 import simplejson
 
 from mage_ai.data_preparation.models.block import Block
-from mage_ai.data_preparation.models.block.constants import CHILD_DATA_VARIABLE_UUID
-from mage_ai.data_preparation.models.constants import VARIABLE_DIR, BlockType
+from mage_ai.data_preparation.models.constants import (
+    CHILD_DATA_VARIABLE_UUID,
+    VARIABLE_DIR,
+    BlockType,
+)
 from mage_ai.data_preparation.models.variables.constants import (
     DATAFRAME_CSV_FILE,
     VariableType,
--- a/mage_ai/data_preparation/models/block/r/__init__.py
+++ b/mage_ai/data_preparation/models/block/r/__init__.py
@@ -8,7 +8,10 @@ import pandas as pd
 import simplejson
 
 from mage_ai.data_preparation.models.block import Block
-from mage_ai.data_preparation.models.constants import BlockType
+from mage_ai.data_preparation.models.block.dynamic.constants import (
+    CHILD_DATA_VARIABLE_UUID,
+)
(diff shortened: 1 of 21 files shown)


mage_ai/data_preparation/models/constants.py
```diff
@@ -27,6 +27,8 @@
 PREFERENCES_FILE = '.preferences.yaml'
 REPO_CONFIG_FILE = 'metadata.yaml'
 VARIABLE_DIR = '.variables'
+CHILD_DATA_VARIABLE_UUID = 'output_0'
+METADATA_VARIABLE_UUID = 'output_1'
```
wangxiaoyou1993, line 31: where is it used?

mage_ai/data_preparation/models/block/r/__init__.py
```diff
@@ -145,7 +157,13 @@ def __render_r_script(
             code=code,
             global_vars=global_vars_str,
             input_paths=[
-                os.path.join(v.variable_path, DATAFRAME_CSV_FILE)
+                os.path.join(
```
wangxiaoyou1993, line 160: why do we change to this path join?

mage_ai/data_preparation/models/constants.py
csharplus: Good catch: it has been removed.

mage_ai/data_preparation/models/block/r/__init__.py
csharplus: The expression `os.path.join(v.variable_path, DATAFRAME_CSV_FILE)` was reading from a path `/home/src/mage_data/............../sample_data/data.csv`, which missed the required `output_0/` in the subpath where the file was stored by the previous data loader block, i.e.,  `/home/src/mage_data/............../sample_data/output_0/data.csv`. Thus, the file path has to be updated to read from the right place.

csharplus: Please see if it needs to be changed to the following code, which would be easier to understand:
```
                os.path.join(
                    v.variable_path,
                    CHILD_DATA_VARIABLE_UUID,
                    DATAFRAME_CSV_FILE
                )
```

csharplus: @wangxiaoyou1993 identified and fixed the root cause (i.e., a much cleaner solution), please review the updated PR, thanks.

wangxiaoyou1993 (approved):

PR #5920, opened 2025-10-21
[xy] Escape characters when interpolating code

wangxiaoyou1993:
# Description

Escape characters when interpolating code
This pull request focuses on improving the handling of file paths, specifically Windows-style paths with backslashes, in the output display utilities. The main changes ensure that backslashes in repository paths are properly escaped before being interpolated into code content, which helps prevent issues with string formatting and regular expressions.

**Improvements to path handling and code interpolation:**

* Escaped backslashes in `repo_path` by replacing single backslashes with double backslashes before using the value in replacements in `add_internal_output_info` (`mage_ai/server/utils/output_display.py`).
* Updated the replacement value for `repo_path` in the replacements list to use the newly escaped path (`mage_ai/server/utils/output_display.py`).
* Modified the code interpolation logic to escape all backslashes in replacement values, ensuring correct substitution when using regular expressions in `__interpolate_code_content` (`mage_ai/server/utils/output_display.py`).
# How Has This Been Tested?

- [x] tested on windows

# Checklist

cc:

### DIFF
--- a/mage_ai/server/utils/output_display.py
+++ b/mage_ai/server/utils/output_display.py
@@ -199,8 +199,8 @@ def __interpolate_code_content(
 
         for placeholder, replacement in replacements:
             placeholder_pattern = f"'{{{placeholder}}}'"
-            # Escape backslashes and other regex special characters in the replacement string
-            escaped_replacement = re.escape(str(replacement))
+            # Escape all regex special characters including backslashes
+            escaped_replacement = str(replacement).replace('\\', '\\\\')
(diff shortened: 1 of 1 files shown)


ZhenoSan (approved):
Code Blocks would run indefinitely before the change. This was due to bad escape character being caused by the program running in C:\Users directory and \U being picked up as unicode escape. Repo path and code were not being escaped properly.

image
 
 After change:
 
image

Code block finishes completion. Although extra message "DBT library not installed." is printed along.

PR #5928, opened 2025-11-01
feat: add search functionality to triggers page

### LINKED ISSUE #5697: Search trigger by name in the UI
It would be very useful if we could search for pipeline name in the Triggers page, specially when you have pages of  trigger, like in the example below.

Image

gbretas:
# Description

This PR adds search functionality to the Triggers page, allowing users to filter triggers by pipeline name or trigger name. This is especially useful when managing multiple pages of triggers.

## Changes

- Add search input field to filter triggers by name or pipeline
- Implement backend filtering with SQLAlchemy OR query
- Filter by both trigger name and pipeline UUID
- Add responsive UI layout with search on left, sort on right

Closes #5697

## Motivation

Currently, users have to manually navigate through pages of triggers to find a specific one. With many triggers in the system, this becomes cumbersome. This change adds a real-time search field that filters triggers server-side, improving UX significantly.

# How Has This Been Tested?

## Manual Testing

- [x] Search by full pipeline name - filters correctly
- [x] Search by partial pipeline name - shows matching results
- [x] Search by trigger name - filters correctly  
- [x] Search with empty field - shows all triggers
- [x] Pagination works correctly with filtered results
- [x] Sort functionality works alongside search
- [x] Backend returns filtered results from all pages (not just current page)

## Test Steps

1. Navigate to `/triggers` page
2. Create multiple pipelines with triggers (or use existing)
3. Type in the search field
4. Verify triggers filter in real-time
5. Test with partial names, full names, and mixed case
6. Verify pagination updates correctly

# Checklist

### DIFF
--- a/mage_ai/frontend/pages/triggers/index.tsx
+++ b/mage_ai/frontend/pages/triggers/index.tsx
@@ -1,6 +1,7 @@
-import { useMemo, useState } from 'react';
+import { useEffect, useMemo, useRef, useState } from 'react';
 import { useRouter } from 'next/router';
 
+import Button from '@oracle/elements/Button';
 import Dashboard from '@components/Dashboard';
 import FlexContainer from '@oracle/components/FlexContainer';
 import Paginate, { ROW_LIMIT } from '@components/shared/Paginate';
@@ -27,6 +28,8 @@ function TriggerListPage() {
   const router = useRouter();
   const [errors, setErrors] = useState(null);
   const [searchText, setSearchText] = useState('');
+  const [searchQuery, setSearchQuery] = useState('');
+  const debounceTimeout = useRef(null);
 
   const q = queryFromUrl();
   const page = q?.page ? q.page : 0;
@@ -39,11 +42,35 @@ function TriggerListPage() {
     [project?.features],
   );
 
+  // Debounce search input
+  useEffect(() => {
+    if (debounceTimeout.current) {
+      clearTimeout(debounceTimeout.current);
+    }
+
+    debounceTimeout.current = setTimeout(() => {
+      setSearchQuery(searchText);
+    }, 500);
+
+    return () => {
+      if (debounceTimeout.current) {
+        clearTimeout(debounceTimeout.current);
+      }
+    };
+  }, [searchText]);
+
+  // Reset page to 0 when search query changes
+  useEffect(() => {
+    if (searchQuery && page !== 0) {
+      goToWithQuery({ page: 0 });
+    }
+  }, [searchQuery]);
+
   const pipelineSchedulesRequestQuery: PipelineScheduleReqQueryParamsType = {
     _limit: ROW_LIMIT,
     _offset: page * ROW_LIMIT,
     order_by: orderByQuery,
-    ...(searchText && { search: searchText }),
+    ...(searchQuery && { search: searchQuery }),
   };
 
   const {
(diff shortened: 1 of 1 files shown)


johnson-mage:
@gbretas Thank you for adding this feature! I have a couple comments after testing it out.
1. Can you please add a "Search" button after the search input that makes the fetch request for the updated triggers (i.e. pipeline schedules) when users click on it or press "Enter" on the keyboard? It currently makes a new request every time a key is pressed, making excessive repeated requests if a user types out a word. You can remove the `Search:` text before the input, since it'd be redundant to have that there and the "Search" button as well. Adding the "Enter" keyboard shortcut isn't absolutely necessary, but a nice enhancement.
2. When users press the "Search" button, the pagination page should reset to the first page. Right now if a user is on page 5 for example, but the number of search results only results in 2 pages (less than 5 pages), it says "No triggers available" even though triggers ARE available--they're just on another page.

I've attached a gif demonstrating both of these. Note how a new request is made every time I enter a character in the search. We want to minimize the number of requests. And I start on page 7 but after entering my search phrase, there are only 4 pages of results, so I have to click on page 1-4 in order to see the results.

trigger search

mage_ai/frontend/pages/triggers/index.tsx
```diff
@@ -71,32 +101,63 @@ function TriggerListPage() {
...
+              onChange={e => {
+                setSearchText(e.target.value);
+              }}
+              onKeyPress={e => {
+                if (e.key === 'Enter') {
+                  setSearchQuery(searchText);
+                  goToWithQuery({ page: 0 });
+                }
+              }}
+              placeholder="Search by pipeline or trigger name..."
```
johnson-mage, line 117: nit: Remove the `...`, since it looks like a single period in the placeholder text. Would look better with no punctuation at the end at all

Image

mage_ai/frontend/pages/triggers/index.tsx
```diff
@@ -37,10 +42,35 @@ function TriggerListPage() {
     [project?.features],
   );
 
+  // Debounce search input
```
johnson-mage, line 45: If you include the `Search` button, including the auto-fetch after user types in the keyboard with debounce is unnecessary. Use one or the other. If you use the explicit search when user presses "enter" or clicks "search", you don't need this useEffect.

mage_ai/frontend/pages/triggers/index.tsx
```diff
@@ -37,10 +42,35 @@ function TriggerListPage() {
...
+
+    return () => {
+      if (debounceTimeout.current) {
+        clearTimeout(debounceTimeout.current);
+      }
+    };
+  }, [searchText]);
+
+  // Reset page to 0 when search query changes
+  useEffect(() => {
```
johnson-mage, line 63: If you remove the debounced search / auto-fetching when user types in the keyboard, this useEffect shouldn't be necessary.

mage_ai/frontend/pages/triggers/index.tsx
johnson-mage: If you use the debounced search, you can use the utility hook at `mage_ai/frontend/utils/hooks/useDebounce.ts`.

mage_ai/frontend/pages/triggers/index.tsx
```diff
@@ -37,10 +42,35 @@ function TriggerListPage() {
...
+    if (searchQuery && page !== 0) {
+      goToWithQuery({ page: 0 });
+    }
+  }, [searchQuery]);
+
   const pipelineSchedulesRequestQuery: PipelineScheduleReqQueryParamsType = {
     _limit: ROW_LIMIT,
     _offset: page * ROW_LIMIT,
     order_by: orderByQuery,
+    ...(searchQuery && { search: searchQuery }),
```
johnson-mage, line 73: Include the `search` in the url query parameters as well (e.g. like how `page` is included and updated in the url when changing pages).

Issue #5642, opened 2025-01-10
[Feature Request] Button(s) for moving blocks up/down in pipeline


oscarlofwenhamn:
**Is your feature request related to a problem? Please describe.**
Sometimes when adding a new block to a pipeline, it's actually meant to be added to the top of the pipeline. Moving blocks using the mouse up/down can be slow and sometimes tricky, since grabbing it and scrolling isn't supported(?).

**Describe the solution you'd like**
One or two buttons added to the block menu:
- "Move to top" (priority)
- "Move up" (nice-to-have)

"Move to top" would move the block in question to the top of the pipeline, and "Move up" would move it up one position, swapping with the block above. 

I think it would be good UX if visual focus remained on the block, so you 1. don't have to scroll up to find it and 2. possibly could quickly re-perform a "Move up" action multiple times, thus achieving something like a "Move up X times" functionality without having to add a bunch of extra buttons.

**Describe alternatives you've considered**
Support for mouse scroll when picking up and moving blocks could be an option, since that would make moving them far more dynamic than what's presently enabled.

wangxiaoyou1993:
This ticket is resolved in this PR: https://github.com/mage-ai/mage-ai/pull/5941
Will be released in version 0.9.79

PR #5933, opened 2025-11-06
feat(frontend): move node to top + auto-scroll (T-5642)

DevDevvy:
# Description
This PR implements the “Move to Top” feature for nodes in the pipeline dependency graph (ticket #5642).

## What was added:

Added a new context menu action labeled “Move to top” for each block/node in the pipeline graph UI and added node ID to block for automatic scrolling.

**Implemented a moveBlockToTop function that:**

- Detaches the selected node from its current position by removing all existing upstream and downstream connections. Makes node with no connections so user can easily create the connection needed.
- Bridges connections intelligently when possible — if the node had one child, the parent(s) now connect directly to that child, preserving logical flow.
- Refreshes the pipeline graph and scrolls the moved node into view automatically at the top of the canvas, providing instant visual feedback.

## What was not implemented and why:

The requested “Move Up” button was not added. Moving a node up one level within a complex DAG (directed acyclic graph) can easily break data dependencies or introduce ambiguous states in the pipeline. Instead, this PR focuses on a safe and predictable “Move to top” action that resets connections and allows users to rewire intentionally.

### Why it’s implemented this way:

Ensures the graph remains valid and consistent while giving users flexibility to reorganize pipelines visually.

The “move to top” logic operates through existing API mutation calls (updateBlockByDragAndDrop), ensuring backend compatibility and data integrity.

The smooth scrolling behavior enhances UX by automatically centering the updated node after re-layout via node ID tracking.

**Dependencies:**
No new external dependencies added.
Relies only on existing APIs and reaflow’s layout re-render cycle.

# How Has This Been Tested?
1. Created multiple test pipelines with varying depths (linear and branched).
2. Used context menu → “Move to top” on nodes with:
 - A single child
 - Multiple children
 - Leaf nodes
3. Verified that:
 - The moved node detached correctly and appeared at the top.
 - Upstream parents reconnected correctly when applicable.
 - The viewport scrolled smoothly to show the new top node.
 - No broken edges or orphaned references occurred after refresh.

- [ ] Test A: Node at bottom with no bottom connections moves to top, UI scrolls to top with node.
- [ ] Test B: Node between two nodes moves to top, passing connections it had to the node below it and moves to top with UI scrolling to top with node.

# Checklist

cc:

@wangxiaoyou1993

### DIFF
--- a/mage_ai/frontend/components/DependencyGraph/index.tsx
+++ b/mage_ai/frontend/components/DependencyGraph/index.tsx
@@ -1042,6 +1042,88 @@ function DependencyGraph({
     upstreamBlocksEditing,
   ]);
 
+    const moveBlockToTop = useCallback(
+    async function moveBlockToTop(blockB: BlockType) {
+      if (!blockB) return;
+
+      // Snapshot neighbors BEFORE any writes
+      const parentUUIDs = [...(blockB.upstream_blocks || [])];
+      const childUUIDs = [...(blockB.downstream_blocks || [])];
+
+      const uniq = <T,>(arr: T[]) => Array.from(new Set(arr));
+
+      // 0) HARD DETACH B FIRST (break mirror-sync loop)
+      await Promise.all([
+        blockB.upstream_blocks?.length || 0
+          ? updateBlockByDragAndDrop({ block: blockB, upstreamBlocks: [] })
+          : Promise.resolve(),
+        blockB.downstream_blocks?.length || 0
+          ? updateBlockByDragAndDrop({ block: blockB, downstreamBlocks: [] })
+          : Promise.resolve(),
+      ]);
+
+      if (childUUIDs.length === 1) {
+        // B had exactly one child → bridge parents to that child
+        const child = blockUUIDMapping[childUUIDs[0]];
+        if (child) {
+          // 1) Parents: replace B with C (or just remove B if not present)
+          for (const pUuid of parentUUIDs) {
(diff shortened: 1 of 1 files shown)


johnson-mage:
@DevDevvy The feature request from [github issue 5642](https://github.com/mage-ai/mage-ai/issues/5642) wasn't actually referring to the dependency graph, but the notebook portion of the Pipeline Editor. The dependencies of the pipeline's blocks shouldn't be affected, just the positioning of the blocks within the notebook. This can be useful if users want to position blocks within the notebook that makes sense with their position within the dependency graph.

Right now, users can reposition the blocks within the pipeline notebook section by dragging the block via the block header and dropping it over another block, but it can be easier and quicker to do if a user simply had an option to "Move up" or "Move to top" from the block header menu.
image

Here's an example. The `falling_cedar` sensor block is towards the bottom of the pipeline, but it makes sense to move it to the top of the pipeline since it is the root node in the dep graph. A user might have to do something like this:
move block one by one

Users can actually move the block near the top and the notebook section does scroll, but it might not be obvious:
move block scroll

Feel free to ask clarifying questions if needed.

johnson-mage:
> @johnson-mage Ah thank you for the clarification - I read block and just assumed node! Should the button(s) be added to the top of the block near the play button or added into the "more actions" list?

@DevDevvy The "Move up" and "Move to top" options can be added to the "More actions" dropdown menu.

Issue #5785, opened 2025-05-30
Mage 0.9.76 Exposed to CVE-2023-46298 via NextJS v 12.3.4


mfaizmzaki:
The title is self explanatory. I'm using the latest version of Mage AI, and upon checking the frontend/package.json, MageAI is using Next version 12.3.4. This version is vulnerable to CVE-2023-46298 (https://nvd.nist.gov/vuln/detail/CVE-2023-46298) relating to Cache-Control header. Need assistance on how to address this.

Issue #5226, opened 2024-06-24
Next.js 12 End of Life


the-archbishop:
Next.js 12 reached EoL on 21 Nov 2022. Security scans are showing that mage-ai is using v12.3.4 still. I attempted to verify, but could not find it in the package.json. 

Version should be upgraded.

johnson-mage:
We are using NextJS v12 right now, but do plan to upgrade the version later, just not right this moment.

ajcasarez:
Are there any updates on plans to upgrade Next.js to the latest version?

wangxiaoyou1993:
We will upgrade to NextJS v15.

ajcreswell:
Update on timing?

PR #5942, opened 2025-11-10
[dd] update the nextjs and axios package to the most stable version

dacdao1:
# Description

With Mage-AI's current version (0.9.78), it utilizes the Next.js package version 12.3.4, which has some vulnerability issues, primarily https://github.com/advisories/GHSA-c59h-r6p8-q9wc (https://nvd.nist.gov/vuln/detail/CVE-2023-46298), #5785. I updated the package to 14.2.33, which has solved the CE-2023-46298 issue, and it is also backward compatible. Unlike version 13 of Next.js, which has a different routing method (using /api instead of /pages) and other issues. I know we were not planning to update the Next.js package last year (#5226), but I don't know if the team would still like to upgrade the package

Axios version 0.27.2 has several vulnerability issues here (https://security.snyk.io/package/npm/axios/0.27.2), and updating to 0.30.2 is the bare minimum to mitigate some of the vulnerabilities on the list. 

# How Has This Been Tested?

- [x] Manually run the Docker image. Start the local environment. Creating pipelines.
1. Update the package.json file within mage_ai/frontend with Next.js version 14.2.33 and axios 0.30.2.
2. Initialize the Docker image since the packages have been changed.
3. Start the local environment of the front end and back end of the application. ( I used the Docker images to start my local environment.
4. Creating a new pipeline, running the pipeline once, and deleting the pipeline run.
- [x] Clicking through different routes of the application
1. Clicking on different pages of the application.
2. Trying to change the settings with the application.

# Checklist

cc:

Hi @johnson-mage , I would like to know if we can update the Next.js and Axios packages to a more secure version. I don't want to update Next.js to version 13 because I would have to change the codebase to match the new routing pattern, which might break other parts of the application. The safest version is 14.2.33, and I didn't want to update to version 15 since I haven't had a chance to build with that version yet. Axios also has some vulnerabilities, and the current package indication, ^0.27.2, might still download either version 0.27 or 0.28 (which might still have some vulnerabilities). By having version ^0.30.2, we would at least be on a safer version while not breaking anything else in the current codebase.

### DIFF
--- a/mage_ai/frontend/components/BlockLayout/BlockLayoutItem/index.tsx
+++ b/mage_ai/frontend/components/BlockLayout/BlockLayoutItem/index.tsx
@@ -77,6 +77,8 @@ function BlockLayoutItem({
   width,
 }: BlockLayoutItemProps) {
   const refMenu = useRef(null);
+  const outerRef = useRef<HTMLDivElement | null>(null); // drop target node
+  const innerRef = useRef<HTMLDivElement | null>(null); // drag handle node
   const [menuVisible, setMenuVisible] = useState<boolean>(false);
   const [editing, setEditing] = useState<boolean>(false);
 
@@ -257,6 +259,15 @@ function BlockLayoutItem({
     }),
     [onDrop],
   );
+  const setDropRef = useCallback((node: HTMLDivElement | null) => {
+  outerRef.current = node;
+  if (node) drop(node);
+  }, [drop]);
+
+  const setDragRef = useCallback((node: HTMLDivElement | null) => {
+  innerRef.current = node;
+  if (!disableDrag && node) drag(node);
+  }, [drag, disableDrag]);
 
   if (detail) {
     return <BlockLayoutItemDetail buildChart={buildChart} height={height} width={width} />;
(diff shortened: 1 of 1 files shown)


johnson-mage:
@dacdao1 Upgrading NextJS (and Axios) is certainly welcome. Thank you for creating this PR. Since this is a significant upgrade for NextJS (that may cause some breaking changes), it might make sense to upgrade to the latest stable version of NextJS available, which would be v16.0.1, though that is very recent and just released last month. Otherwise, we may have to do another major upgrade for NextJS pretty soon, which could save some effort/time.

Do you think you'd be able to upgrade NextJS in Mage to v16? Or does it make sense to just upgrade to v14?

Regardless of whether it's v14 or v16, some things to keep in mind:
1. The github "Build and test / test_backend (3.10)" check should be successful. It is currently failing. Refer to error in `#2` below.
2. In addition to testing Mage with the updated version of Next in your development environment (e.g. with `./scripts/dev.sh`), you should also be testing a fresh build locally. Delete your local `node_modules` and `.next` folders, run `yarn` and then `yarn build`. We'll need to resolve these errors during the build. I was able to recreate the error from `#1`:
image

3. A big change in NextJS v14 is ["The next export command has been removed in favor of output: 'export' config.".](https://nextjs.org/docs/pages/guides/upgrading/version-14#v14-summary). An important part of our build process is creating a static export with the `yarn export_prod_base_path` command (see the package.json file scripts section). Since `next export` is no longer supported, we'll need to make the appropriate change. Refer to NextJS [docs](https://nextjs.org/docs/pages/guides/static-exports) on static exports.

5. When you are running Mage in your dev env, are you coming across this "NextRouter was not mounted" error when trying to open a file from the file browser on the Pipeline Editor page?
nextrouter not mounted

6. Also check for any minimum version requirements for other dependencies like react, react-dom, and typescript.

Issue #5985, opened 2025-12-30
[BUG] Google Cloud Run Jobs are retried by the client after 900s even when the user timeout is longer


sm-data:
### Mage version

v0.9.78

### Describe the bug

When running a pipeline block using the GCP Cloud Run executor, the execution fails with a `google.api_core.exceptions.RetryError` if the job takes longer than 900 seconds (15 minutes). 
This happens even if `timeout_seconds` is configured to a higher value (e.g., 3600) in `metadata.yaml`

### To reproduce

1. Configure a Mage project to use the GCP Cloud Run executor.
2. In `metadata.yaml`, set the timeout to be longer than 15 minutes:
```
gcp_cloud_run_config:
  timeout_seconds: 3600
```
4. Create a pipeline block that sleeps for or takes longer than 900 seconds to execute.
```
import time
import datetime as dt
if 'custom' not in globals():
    from mage_ai.data_preparation.decorators import custom

@custom
def transform_custom(*args, **kwargs):
    start_time = dt.datetime.now()
    print(f"{start_time}: Starting")
    for i in range(1,10):
        time.sleep(100)
        print(f"{dt.datetime.now()}: Finished process #{i} in what should be {i*100} seconds. Observed as {dt.datetime.now() - start_time}")

    return {}
```
6. Trigger the pipeline.
7. Observe that after exactly 900 seconds, the Mage process raises a RetryError and marks the block as failed (or attempts a retry), even though the Cloud Run job is still running successfully on GCP.

### Expected behavior

The Mage client should respect the `timeout_seconds` configuration when waiting for the Cloud Run job to complete. If I set the timeout to 3600 seconds, the client should wait up to 3600 seconds for the `operation.result()` to return.

### Additional context

The issue appears to be in `mage_ai/services/gcp/cloud_run/cloud_run.py` on line 91

The `operation.result()` call is made without arguments, which causes it to use the default timeout.

[This documentation](https://googleapis.dev/python/google-api-core/latest/operation.html#google.api_core.operation.Operation) from google says that:
> If the timeout argument is specified in the result() method, it will override the polling.timeout property

Hence I have tested changing this line to be:
`response = operation.result(timeout=cloud_run_config.timeout_seconds)`
Which works as expected.

wangxiaoyou1993:
Resolved in this PR: https://github.com/mage-ai/mage-ai/issues/5985

PR #5986, opened 2025-12-30
using the timeout from the project-wide config in the .result() rather than falling back to default

sm-data:


sm-data:
#5985

sm-data:
@wangxiaoyou1993 please could you have a look? I've tested this on my own deployment, and have put a fair amount of detail in the issue  #5985  - do let me know if you want any more.
It's a fairly simple issue - I don't think that the sklearn errors in the checks are caused by this change...!

Issue #6062, opened 2026-03-20
Filter pipeline runs by datetime range


wangxiaoyou1993:
**Is your feature request related to a problem? Please describe.**

Currently, the pipeline runs list page (/pipeline-runs) does not provide a way to filter runs by a specific date/time range. This makes it difficult for users to quickly locate runs within a particular time window (e.g., debugging recent failures, auditing historical runs, or analyzing activity over a given period). As the number of pipeline runs grows, this lack of filtering significantly impacts usability and efficiency.

**Describe the solution you'd like**

Introduce a datetime range filter for the pipeline runs list page.

This would involve:

Backend
* Extend the pipeline runs API to support filtering by a datetime range (e.g., start_time and end_time query parameters).
* Maintain backward compatibility for existing API consumers.

Frontend
* Add a date/time range picker to the /pipeline-runs filter UI.
* Allow users to specify both start and end datetime values.
* Preserve filter state in the URL (if applicable) for shareability and refresh persistence.

ckabiroan:
<img width="949" height="475" alt="Image" src="https://github.com/user-attachments/assets/06abf14a-99f7-477f-9db0-9411f5a7db06" />

Here's the initial quick design.

This is based on the current Logs.

<img width="958" height="165" alt="Image" src="https://github.com/user-attachments/assets/1b11a149-7b0d-481f-8fbb-1ffe3646cfda" />

Pros:
- Fastest development.
- Minimal changes.
- More consistent interface for filtering by datetime range.
- Highly visible / discoverable
- Easy to remove or refactor later

Cons:
- There will be two filter buttons
- Takes up horizontal space

Recommendation:
1. Use this as the v0.1 design for table filtering to quickly let users filter by datetime range.
2. Then work with design to iterate on a more elegant filtering UI/UX.

ckabiroan:
Here's a demo video of the prototyped pipeline runs date time filtering UI/UX based off the Logs UI/UX.

https://www.youtube.com/watch?v=BuBMlwMHgXg

PR #6071, opened 2026-03-26
Extend filter to allow date range filtering, add execution date filter to pipeline runs (#6062)

olga-007:
## Summary                                                                                                                                                                                                                        
                                                                                                                                                                                                                                    
  Seamlessly integrates date range filtering into the existing filter menu for pipeline runs (#6062). Execution date was chosen as the filter field because it best fits the task requirements and is already visible as a column in the table, making the filter immediately verifiable by users.
                                                                                                                                                                                                                                    
  ## Approach     

  Instead of adding a separate date picker button, the date range filter lives inside the existing filter menu. Users get one filter entry point — no extra buttons, no new interaction patterns to learn. An alternative design with a separate date picker button was considered but rejected to avoid a fragmented filtering experience with multiple entry points.
                                                                                                                                                                                                                                    
  The implementation is generic: any page using the shared ToggleMenu/Toolbar can add date range filtering by passing a config object. Multiple date range filters can coexist alongside toggle filters in the same menu.

  ## Test plan

  - [ ] Filter by start date, end date, and both — verify results match                                                                                                                                                             
  - [ ] Clear dates with X button, change time values — verify URL updates
  - [ ] Confirm existing toggle filters still work alongside date filters                                                                                                                                                           
  - [ ] Verify applied filter count badge is correct (includes date range filters)
  - [ ] Verify "Defaults" button resets date range filters
  - [ ] Verify on both `/pipeline-runs` and `/manage/pipeline-runs`                                                                                                                                                                 
   
  ## Notes                                                                                                                                                                                                                          
                  
  - Start/end date validation (start < end) can be added as a polish pass.
  - ToggleMenu could be renamed to FilterMenu to reflect its expanded scope.
  - Calendar stays open after date selection for consistency with calendar behavior elsewhere in the app.

### DIFF
--- a/mage_ai/api/policies/PipelineRunPolicy.py
+++ b/mage_ai/api/policies/PipelineRunPolicy.py
@@ -101,6 +101,8 @@ PipelineRunPolicy.allow_query([
     'backfill_id',
     'disable_retries_grouping',
     'end_timestamp',
+    'execution_date_end',
+    'execution_date_start',
     'global_data_product_uuid',
     'include_all_pipeline_schedules',
     'include_pipeline_tags',
--- a/mage_ai/api/resources/PipelineRunResource.py
+++ b/mage_ai/api/resources/PipelineRunResource.py
@@ -42,6 +42,12 @@ class PipelineRunResource(DatabaseResource):
             pipeline_schedule_id = parent_model.id
 
         start_timestamp, end_timestamp = get_query_timestamps(query_arg)
+        execution_date_start, execution_date_end = get_query_timestamps(
+            query_arg,
+            start_key='execution_date_start',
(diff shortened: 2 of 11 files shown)


wangxiaoyou1993:
Can you upload a screenshot or screen recording showing how you test the feature?

olga-007:
I'm currently traveling internationally and don't have access to my laptop,
so I'm unable to record a demo right now. The original instructions didn't
mention a screenshot or recording as a requirement — just implementing and
submitting a PR, which I've completed. You should be able to run the code
and verify it directly. That said, I'm happy to jump on a call to walk
through the implementation once I'm back. Let me know what works best for
you!


Olga

PR #6082, opened 2026-04-04
feat: add truncate_before_full_table_sync option to full table sync

### LINKED ISSUE #4708: Data Integration - Add truncate table to full table sync
I though that a Full table integration would truncate the table and then place the full table into the destination. However, the result is that the data is not truncated before.

I think that having an extra option to either truncate the table or not should be enough. I understand that in some cases we just want to copy the whole table into a new partition and some other, just have one partition.

Running a block to truncate the table before running the integration pipeline.

Zoya072:
Fixes #4708

Added a new config option `truncate_before_full_table_sync` to data integration 
SQL destination blocks (PostgreSQL, MySQL, and other SQL destinations).

When set to `true`, the destination table is truncated once before the first 
batch is written during a FULL_TABLE sync. This ensures stale or duplicate 
records are removed before a full reload.

Previously, users had to manually run a separate block to truncate the table 
before running the integration pipeline. This change makes that a first-class 
config option.

## Usage
Add to your destination config YAML:
    truncate_before_full_table_sync: true
    
- Manually tested by running a FULL_TABLE sync with truncate_before_full_table_sync: true
  and verifying the table was truncated before new records were inserted.

- Verified that existing pipelines without the flag are unaffected (defaults to false).

- Verified that truncation only happens once per stream per run, even when 
  multiple batches are processed.

### DIFF
--- a/mage_integrations/mage_integrations/destinations/base.py
+++ b/mage_integrations/mage_integrations/destinations/base.py
@@ -189,8 +189,25 @@ class Destination(ABC):
     def after_process(self) -> None:    # noqa: B027
         pass
 
-    def export_batch_data(self, record_data: List[Dict], stream: str, tags: Dict = None) -> None:
-        raise NotImplementedError('Subclasses must implement the export_batch_data method.')
+    def export_batch_data(self, stream, schema, records, replication_method=None, **kwargs):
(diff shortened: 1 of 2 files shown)


wangxiaoyou1993:
Can you share screenshots of your testing results with Mage data integration pipelines?

PR #6084, opened 2026-04-07
Fix: truncate destination table for FULL_TABLE replication

### LINKED ISSUE #4708: Data Integration - Add truncate table to full table sync

Zoya072:
# Description
This change fixes stale data accumulation for `FULL_TABLE` replication in SQL destinations.

Previously, when the destination table already existed, Mage only ran `ALTER TABLE` (to add new columns if needed) and then performed inserts. Because of that, rows deleted from the source were never removed from the destination, so stale data could remain indefinitely.

For `FULL_TABLE` replication, the destination should reflect the source as a full replacement. To support that behavior, this change adds a `TRUNCATE TABLE` step before inserts when:
- the table already exists, and
- the replication method is `FULL_TABLE`

Changes made:
- Added `build_truncate_table_commands()` in `mage_integrations/mage_integrations/destinations/sql/base.py`
- Updated `build_query_strings()` to truncate the destination table for `FULL_TABLE` replication when the table already exists
- Preserved existing behavior for `INCREMENTAL` and `LOG_BASED` replication methods

This ensures `FULL_TABLE` replication behaves as a full refresh and prevents stale data accumulation.

Fixes #4708

# How Has This Been Tested?
I validated the change by reviewing the SQL destination flow and confirming the new logic path for `FULL_TABLE` replication when the destination table already exists.

Reproduction flow:
1. Use a SQL destination with replication method set to `FULL_TABLE`
2. Ensure the destination table already exists
3. Run a sync where source data has fewer rows than a previous sync
4. Confirm that the destination table is truncated before inserts
5. Verify that stale rows are not retained in the destination

- [x] Test A: Verified `FULL_TABLE` path now generates `TRUNCATE TABLE` before inserts for existing tables
- [x] Test B: Verified `INCREMENTAL` and `LOG_BASED` paths remain unchanged

# Checklist

### DIFF
--- a/mage_integrations/mage_integrations/destinations/sql/base.py
+++ b/mage_integrations/mage_integrations/destinations/sql/base.py
@@ -46,9 +46,11 @@ class Destination(BaseDestination):
         return self.config.get("skip_schema_creation") is True
 
     def clean_column_name(self, col):
-        return clean_column_name(col,
-                                 lower_case=self.use_lowercase,
-                                 allow_reserved_words=self.allow_reserved_words)
+        return clean_column_name(
+            col,
+            lower_case=self.use_lowercase,
+            allow_reserved_words=self.allow_reserved_words,
+        )
 
     def test_connection(self) -> None:
         sql_connection = self.build_connection()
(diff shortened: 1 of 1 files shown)


wangxiaoyou1993:
Please share a short video or screenshots how you tested it in Mage UI.

joshpachner:
+1  on my side. this would be nice to have. ive found i cant reliably use the Integration pipeline type with google sheets, even with using the __google_sheet_row as an update key, the rows can just be too volatile. and often the data in a  google sheet doesnt have a key(s) for it.

